PR #39, opened 2026-04-14
feat: manual library scan trigger (API + Settings UI)

### LINKED ISSUE #10: Manual Library Scan trigger (UI button + API endpoint)
## Summary

Library scan currently only runs on a 6h cron. There's no way to kick it off on demand — meaning after adding authors, dropping files into `/books/`, or doing a bulk CSV import, you either wait up to 6 hours or restart the container (which also doesn't help since the cron resets).

## Request

- A **Scan Library** button in the UI (Settings → Library, or on the Books page).
- A corresponding API endpoint (e.g. `POST /api/v1/library/scan`) so external tools can trigger it.

## Why

Primary use case is first-time setup / Readarr migration:

1. Import authors via CSV or `readarr.db`
2. Have an existing `/books/` library on disk
3. Want Bindery to immediately reconcile files → wanted books

Right now step 3 is "wait 6 hours" which is a rough first-run experience.

vavallee:
## Summary

Adds an on-demand library scan so users don't have to wait up to 6 hours for the scheduled run.

**Backend**
- New `POST /api/v1/library/scan` — fires `scanner.ScanLibrary` in a goroutine and returns `202 Accepted` immediately
- New `internal/api/LibraryHandler` wired up in `main.go`

**Frontend**
- **Settings → General → Library** section with a **Scan Library** button (shows "Scanning…" while in-flight)
- `api.triggerLibraryScan()` method added to the API client

Closes #10

### DIFF
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -147,6 +147,7 @@ func main() {
 	indexerHandler := api.NewIndexerHandler(indexerRepo, bookRepo, authorRepo, idxSearcher, settingsRepo, blocklistRepo)
 	dlClientHandler := api.NewDownloadClientHandler(dlClientRepo)
 	queueHandler := api.NewQueueHandler(downloadRepo, dlClientRepo, bookRepo, historyRepo)
+	libraryHandler := api.NewLibraryHandler(importScanner)
 	fileHandler := api.NewFileHandler(bookRepo)
 	historyHandler := api.NewHistoryHandler(historyRepo, blocklistRepo)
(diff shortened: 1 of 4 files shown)


derekoharrow:
Setup Bindery to use my calibre ebook folder as the library path and then ran Scan Library.

It says

Files Found: 1926
Reconciled: 211
Unmatched: 1715

210 (not 211) books now appear in my Books list under "In library".

(a) What's happens with the rest - why aren't they added to the library

(b) The scan didn't seem to add any authors at all

(c) What about audiobooks - it didn't scan/find/add them at all.

Thanks

PR #62, opened 2026-04-14
feat(calibre): integrate calibredb (closes #32)

### LINKED ISSUE #32: feat: optional Calibre integration via calibredb
## Feature request

Add opt-in integration with [Calibre](https://calibre-ebook.com/) via the `calibredb` CLI so imported books can be added to a Calibre library without losing track of the file when Calibre reorganises it.

## Background

LazyLibrarian has a similar integration but loses track of files after Calibre moves them — Calibre renames and relocates files based on the first author's name when metadata is corrected. The fix is to store the Calibre book ID at add-time and resolve the current path on demand.

**Calibre IDs are stable.** They are SQLite `INTEGER PRIMARY KEY` rowids — deleting a book does not shift IDs for other books, so the stored ID remains a reliable pointer indefinitely.

## Proposed behaviour

1. After a successful import, if Calibre integration is configured, run:
   ```
   calibredb add <file> --library-library <library>
   ```
2. Parse the returned Calibre book ID from stdout and store it against the book record in Bindery's DB.
3. When Bindery needs the current file path (e.g. for re-import, dedup, or a future "open in Calibre" link), resolve it on demand:
   ```
   calibredb list --fields id,formats --search "id:<stored_id>"
   ```
4. ISBN-based lookup is explicitly not used — it breaks for multiple versions of the same ISBN.

## Scope

- New Settings section: **Calibre** — `calibredb` binary path + library path (local directory) or content server URL (`http://...`).
- Store `calibre_id INTEGER` on the book record (nullable; null = not in Calibre).
- Integration is purely opt-in; no behaviour change when unconfigured.
- `calibredb` is not bundled in the Docker image — users running in Docker/k8s must either mount the Calibre library directory directly or point at a Calibre content server.

## Known constraints

- Calibre is primarily a desktop app; headless/server use via `calibredb` is supported but adds a deployment dependency.
- Two-copy problem: Calibre makes its own copy of the file. Bindery's `/books/` path and Calibre's path diverge after add. Decision needed: keep both, or treat Calibre as authoritative and remove Bindery's copy post-add.
- NFS-mounted Calibre libraries + SQLite WAL mode can be fragile — should document this clearly.

vavallee:
## Summary

- Post-import hook shells out to `calibredb add --with-library <path>` and stores the returned Calibre book id on the Bindery book row for future OPDS / cross-library lookups.
- Opt-in via three settings (`calibre.enabled`, `calibre.library_path`, `calibre.binary_path`) exposed through Settings → General → Calibre with a Test connection button that probes `calibredb --version`.
- Calibre sync is a best-effort mirror: a failed `calibredb add` is logged and swallowed so a missing binary or unreachable library never rolls back an otherwise-good Bindery import.

## What's in here

- **`internal/calibre`** — new package wrapping `calibredb` (`Add`, `Test`, config struct). 92.1% test coverage, runner function injectable for table-testing argv construction.
- **`internal/db/migrations/008_calibre.sql`** — adds nullable `calibre_id INTEGER` column + `idx_books_calibre_id` partial index on books; seeds three `calibre.*` settings rows with `INSERT OR IGNORE` so operator edits survive upgrade.
- **`internal/importer/scanner`** — `WithCalibre(client)` hook + `pushToCalibre` called after both ebook and audiobook imports. Tests cover disabled / happy / Add-fails / `ErrDisabled` / nil-client.
- **`internal/api/calibre`** — `/calibre/test` endpoint + `LoadCalibreConfig` helper. Settings `Set` now validates `calibre.library_path` is an existing directory and `calibre.binary_path` is an executable regular file.
- **`internal/models/book`** + **`internal/db/books`** — new `CalibreID *int64` field, `SetCalibreID` repo method, column included in the `query` scan.
- **Frontend** — `CalibreSection` in `SettingsPage.tsx` with live toggle, two validated paths, and Test button. `client.ts` gets `CalibreSettings`, `CalibreTestResult`, `Book.calibre_id`, and `testCalibre()`.
- **`.golangci.yml`** — gosec G204 exclusion scoped to `internal/calibre/client.go` (launching `calibredb` is the whole point of the package; binary + args come from validated admin settings).
- **CHANGELOG.md** — entry under `[Unreleased] → Added`.

## Test plan

- [x] `go vet ./...` clean
- [x] `gofmt -l .` clean
- [x] `golangci-lint run ./...` → 0 issues
- [x] `go test -race -coverprofile=coverage.out ./...` passes; new `internal/calibre` at 92.1% coverage (exceeds 70% bar)
- [x] `npm run lint` clean; `npm run build` produces dist
- [x] `TestMigrate008_CalibreOnFreshDB` — column + index + seeded settings on a fresh DB
- [x] `TestMigrate008_CalibreOnUpgradeFromPreCalibre` — simulates v0.7.2 → v0.8.0 upgrade and proves `INSERT OR IGNORE` preserves operator-set `calibre.enabled=true`
- [x] `TestPushToCalibre_Disabled` — regression guard that imports are byte-identical to v0.7.2 when Calibre is off
- [x] `TestSettings_SetRejectsInvalidCalibrePath` — 400 on bad path locks in the contract for the frontend
- [x] Settings roundtrip: PUT `/setting/calibre.library_path` → GET returns the same value (exercised via `TestLoadCalibreConfig_ParsesFromSettings`)
- [x] `TestPushToCalibre_AddFailsDoesNotPoison` — a failing `calibredb add` leaves `calibre_id` NULL and doesn't affect the import state

## Notes for reviewer

- `calibredb add` has no machine-readable output for the add command — we regex `Added book ids: N` which has been stable across Calibre releases. Multi-id output (archives yielding multiple books) returns the first id because Bindery maps one book row to one imported file.
- The Test endpoint force-enables `calibre.enabled` for the probe only, so operators can click Test before committing to saving `enabled=true` in the DB. The scanner still honors the persisted flag at runtime.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -117,6 +117,13 @@ linters:
           - errcheck
           - gosec
           - noctx
+      # Calibre integration: launching calibredb with admin-configured paths is
+      # the entire purpose of the package. The binary and library_path come
+      # from validated settings, not untrusted input.
+      - path: internal/calibre/client\.go
+        linters:
+          - gosec
+        text: "G204:"
 
 formatters:
   enable:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to Bindery are documented here. Format loosely follows
 
 The `development` branch carries the in-flight feature set for the next release. Images are published as `ghcr.io/vavallee/bindery:development` and `:dev-<sha>`; point ArgoCD at the `development` branch to follow. Treat these features as beta — schema migrations are additive and safe, but UX may still shift before tagging.
 
+### Added
+
+- **Calibre library integration via `calibredb` ([#32](https://github.com/vavallee/bindery/issues/32))** — after a successful import, Bindery mirrors the book into a configured Calibre library by shelling out to `calibredb add --with-library <path>` and stores the returned Calibre book id on the Bindery book row for future OPDS and cross-library lookups. Opt-in under Settings → General → Calibre with three fields (enabled / library path / binary path) and a Test connection button that probes `calibredb --version`. Failures during the Calibre call are logged and swallowed so a missing binary or unreachable library never rolls back an otherwise-good Bindery import.
+
 ## [v0.7.2] — 2026-04-14
 
 Quality release. Bulk actions land for users curating large libraries (the painful-after-CSV-import flow), the silent library-scan bug is fixed, and backend coverage jumps from 34% to 53% to quiet codecov and harden the regression safety net.
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/vavallee/bindery/internal/api"
 	"github.com/vavallee/bindery/internal/auth"
+	"github.com/vavallee/bindery/internal/calibre"
 	"github.com/vavallee/bindery/internal/config"
 	"github.com/vavallee/bindery/internal/db"
 	"github.com/vavallee/bindery/internal/importer"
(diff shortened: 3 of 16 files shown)


derekoharrow:
I've tried this and it doesn't seem to work for me.

I downloaded calibredb and put it in /config/calibre/

From my docker host, I can run calibre --version and it works fine.

In Bindery I've got it setup in settings as /config/calibre/calibredb, but it fails the test with no such file or directory.

I've also tried mapping calibredb to /usr/bin/calibredb and it still fails the test.

I've got debug logging enabled, but nothing appears in the log file.

PR #90, opened 2026-04-15
fix(metadata): enrich book covers from Google Books / Hardcover when OL has none

vavallee:
## Problem

A large fraction of books arrive from the OpenLibrary works endpoint with no cover image. OL attaches cover IDs to *editions*, not to *work* records, so `GetAuthorWorks` returns mostly coverless entries. The `enrichBook` helper (which contacts Google Books / Hardcover) was already filling descriptions and ratings — but it never touched `ImageURL`.

Additionally, `GetBook`'s enrichment trigger only fired when the description was short (`< 50 chars`), so a book with a real description but no cover would silently skip the entire enrichment path.

As a secondary benefit: Google Books and Hardcover tend to surface the dominant-language (usually English) edition, which reduces the Spanish/French cover problem for authors like Harari whose canonical OL edition is non-English.

## Changes

| File | Change |
|---|---|
| `aggregator.go` — `enrichBook` | Also fills `ImageURL` from enricher when empty |
| `aggregator.go` — `GetBook` | Triggers enrichment when `ImageURL == ""` in addition to short description |
| `aggregator.go` — `GetAuthorWorks` | Calls `enrichBook` for each coverless work before caching — covers are saved to DB on author-add and Refresh |

## Tests

5 new cases in `aggregator_test.go`:
- `enrichBook` fills cover when missing
- `enrichBook` keeps existing OL cover (no replacement)
- `GetBook` triggers enrichment on empty cover even with long description
- `GetAuthorWorks` enriches covers for coverless works, preserves existing covers
- `GetAuthorWorks` with no enrichers — no panic, covers stay empty

## Wiki

Added [Troubleshooting](https://github.com/vavallee/bindery/wiki/Troubleshooting) page to the wiki with sections on covers (including the language issue), downloads, library scan, author search failures, and authentication lockout.


vavallee:
Cherry-picked directly onto development (fb1c0f3 + 23bc114) — branch had deploy commit conflicts on rebase.

PR #99, opened 2026-04-15
feature: Add Transmission torrent client support and downloader abstraction layer

dginther:
### Overview
This PR introduces Transmission as a supported torrent download client alongside the existing qBittorrent and SABnzbd integrations. At the same time, the ad-hoc per-client dispatch logic scattered across API handlers is consolidated into a single `downloader` package with a clean adapter interface. Two bugs surfacing during end-to-end testing are also fixed: a Prowlarr Torznab URL normalization issue and a Transmission `download-dir` path validation error.

---

#### Transmission RPC client (transmission)

New package implementing Bindery's first Transmission integration:

- `client.go` — HTTP client for the Transmission RPC API, handling the X-Transmission-Session-Id challenge/retry handshake, optional HTTP Basic Auth, `torrent-add` (magnet and `.torrent` URL), `torrent-get` for status polling, and a `Test` connectivity check.
- `types.go` — RPC request/response structs.
- DB migration `007_transmission.sql` — adds a `torrent_id` column and index to `downloads` for Transmission torrent tracking (separate from the string-based `remote_id` used by SABnzbd/qBittorrent).

#### Downloader adapter (adapter.go)

Previously, each API handler constructed client objects and dispatched to them inline. This PR extracts that logic into a package-level adapter:

- `SendDownload` — single entry point for all three client types; returns a `SendResult` carrying the protocol type, remote ID, and a flag indicating whether the client uses integer torrent IDs.
- `TestClient` — unified connectivity test dispatch.
- `IsTorrentClient` / `ProtocolForClient` — helpers used by the queue and scheduler to record the correct protocol (`torrent` vs `usenet`) and track IDs appropriately.
- queue.go and download_clients.go are simplified to call the adapter rather than constructing clients themselves.

#### qBittorrent updates (qbittorrent)

- `AddTorrent` signature updated to accept a `savePath` parameter (parallel to Transmission's download directory override) so both torrent clients are consistent through the adapter.

#### Importer and scheduler integration (scanner.go, scheduler.go)

- The import scanner now resolves completed torrent downloads using `torrent_id` when the client is Transmission, in addition to the existing `remote_id` path used for SABnzbd/qBittorrent.
- The scheduler polling loop handles Transmission status responses in the same unified pass.

---

#### `fix(indexer)`: Prowlarr Torznab endpoint URL handling

Prowlarr exposes per-indexer Torznab feeds at paths like `/1/api`. The previous client blindly appended `/api` to the stored URL, turning `http://prowlarr:9696/1/api` into `http://prowlarr:9696/1/api/api`, which returns an HTML error page and fails the capabilities test with:

```
caps: expected element type <caps> but have <html>
```

Changes:
- `normalizeEndpointURL()` detects whether the path already ends in `/api` or contains `torznab` and leaves it untouched; bare root URLs still get `/api` appended.
- `buildURL()` replaces all `fmt.Sprintf` URL construction — it parses the stored base URL and manages only query parameters, preserving whatever path was configured.
- `New()` extracts `apikey` from the URL query string when no explicit key is passed, supporting Prowlarr's one-click "copy Torznab feed" URLs.

#### `fix(downloader)`: Transmission rejects non-absolute `download-dir`
In testing my Transmission addition, when I added a torrent via the UI, I noticed a problem, so here is a bug fix for my own  addition.

Bindery repurposes the `Category` field as an optional directory override for Transmission. Other clients accept bare labels like `"books"`, but Transmission requires `download-dir` to be an absolute filesystem path and returns:

```
add torrent failed: download directory path is not absolute
```

The fix: `SendDownload` now only forwards `Category` as `download-dir` when it starts with `/`; otherwise it passes an empty string and Transmission falls back to its own configured default.

---

### UI (SettingsPage.tsx)

- The Category field label is conditionally renamed to **Download Directory** for Transmission clients, with a placeholder and help text clarifying it expects an absolute path or can be left blank.
- The indexer URL field shows contextual help for both plain Newznab and Prowlarr Torznab endpoint formats.

---

### Test coverage

| File | What's covered |
|---|---|
| `transmission/client_test.go` | `AddTorrent`, `GetStatus`, `Test`, session ID handshake, auth, error cases |
| `downloader/adapter_test.go` | `SendDownload` for all three client types; `TestClient`; zero-ID rejection; relative vs absolute `download-dir` |
| `indexer/newznab/client_test.go` | Full Torznab path preserved, API key extracted from URL, root URL normalised to `/api` |
| `api/queue_test.go` | Queue grab flow for Transmission and qBittorrent via the adapter |


vavallee (commented):
Great first contribution — the Transmission client is solid and the adapter abstraction is a real improvement over scattered per-client dispatch. A few things need addressing before this can merge, one of them a blocker.

---

## Blocker

### Migration number collision (`007_transmission.sql`)

`007_downloads_fk_set_null.sql` already exists on `main`. The migration runner orders files by filename, so a second `007_` file either silently collides or breaks the run order. This **will corrupt upgrades on existing installs**.

Rename to `013_transmission.sql` (012 is the current tip on `development`).

---

## Wrong base branch

This PR targets `main` directly. The project convention is features → `development` first, then released to `main` via a version cut. Please retarget to `development`.

---

## Significant issues

### `GetFirstEnabled` breaks live-status overlay for SABnzbd queues

`queue.go` List handler changed from `GetFirstEnabledByProtocol(ctx, "usenet")` to `GetFirstEnabled(ctx)`. If someone has both SABnzbd and Transmission configured and `GetFirstEnabled` returns the torrent client, `usesTorrentID` will be `true` — so the loop skips every `SABnzbdNzoID`-only item and no live status is overlaid. Users with mixed clients will see a dead queue. Either query both clients or keep protocol-scoped lookups per item.

### Username stored in `URLBase` (readarr.go)

The Readarr importer now puts `s.Username` into the `URLBase` field. The PR description says `Username`/`Password` fields were added to `DownloadClient`, but that change isn't in the diff — the actual schema reuses `url_base` as a username holder. This is confusing for any future contributor and fragile (url_base is meant for path prefixes). If the schema needs credentials, add the columns explicitly.

### Stopped torrent = failed is too aggressive

Scanner marks a torrent as failed when `status == 0 || status == 6` (stopped) and not complete. A manually paused torrent gets flagged as a failure immediately. Transmission doesn't distinguish paused-by-user from stopped-on-error at the status code level — you need the `errorString` field from the `torrent-get` response to tell them apart. Consider treating stopped as failed only when `errorString != ""`, or add a minimum age check.

---

## Minor

**`TRANSMISSION_INTEGRATION.md`** — adds a top-level doc outside the project's convention (everything lives in CHANGELOG + README). The content is good; please fold it into the README's Operations section and the CHANGELOG `[Unreleased]` block, then drop the standalone file.

**Lost log context in cleanup** — `"failed to delete SABnzbd history entry"` with `nzoID` became `"cleanup failed"` with only `error`. Worth keeping the client type and remote ID in the log attrs so ops can correlate.

**`buildURL` retains apikey in baseURL** — `normalizeEndpointURL` doesn't strip `apikey=` from the stored base URL, so it appears twice in `u.Query()` before being overwritten by `q.Set("apikey", ...)`. Not a bug, but cleaning it up in `New()` would make the stored URL less surprising.

---

## What's good

- The adapter pattern (`SendDownload` / `TestClient` / `GetLiveStatuses`) is the right call — removes a lot of repetitive dispatch code and makes adding a fourth client a one-file change.
- Prowlarr Torznab URL normalization is correct and the `buildURL` helper is cleaner than the old `fmt.Sprintf` construction. Tests cover the path-already-ends-in-api case, key extraction from URL, and root normalization.
- Transmission client handles the session-ID challenge/retry and Basic Auth correctly.
- Self-caught and fixed the `download-dir` absolute path requirement — good instinct to add that test.
- Test coverage spans all three client types through the adapter.

Fix the migration number, retarget the branch, and address the `GetFirstEnabled` queue issue — the rest can be follow-up if you prefer.

vavallee (changes requested):
Three things need fixing before merge — see inline comment for full details. Happy to help once addressed.

dginther:
Fixed some linting errors and then a bunch of unhandled error issues that the bot came up with. Not sure if you want those as part of my PR, I just wanted to make sure that the CI checks are passing before we move on with this

dginther:
Updated migrations numbers to accommodate other changes integrated here. You're moving fast on this today, just trying to keep my branch up to date with your changes and make sure this integrates cleanly!

dginther:
Removed accidentally included coverage files.

vavallee (commented):
Good work overall — the Transmission implementation is solid. Session ID / CSRF retry is correct, SSRF protections are thorough, the adapter pattern is clean, and coverage is comprehensive. Happy to merge this.

One thing to clean up before I hit the button:

**Dead code in `GetTorrents`** — the `downloadDir` filter branch (`if downloadDir != "" { ... }`) is never exercised: the adapter always calls `GetTorrents(ctx, "")`. Either wire it up (pass the configured download dir so Bindery only sees its own torrents on a shared Transmission instance — actually useful) or drop the parameter and filter entirely. As-is it's misleading.

Everything else looks good. Fix that and I'll merge.

dginther:
Wired up the GetTorrents downloadDir, and resolved some conflicts. If any of the conflicts were resolved in a way you didn't want, please feel free to change them before you merge!

vavallee:
Thanks for this — solid work. The downloader abstraction layer is exactly the right architectural move, and the Prowlarr Torznab URL fix addresses a real pain point for anyone using Prowlarr's one-click feed URLs.

A few things to resolve before this can merge:

**Migration number conflicts (blocking)**

`development` now has `014_book_excluded.sql` and `015_recommendations.sql` from recently merged PRs. Your `014_download_client_credentials.sql` and `015_transmission.sql` collide with both. Please renumber them to `016_download_client_credentials.sql` and `017_transmission.sql`.

**Scheduler conflict**

`internal/scheduler/scheduler.go` was also modified by the recommendation engine PR that just landed. Worth checking for conflicts there — specifically around the dispatch logic vs. the new `RecommendationEngine` interface that was added.

**Internal PR superseded**

We had an internal PR #149 for Transmission support. Yours is a superset (adds the adapter layer, better tests, the Torznab fix). I'll close #149 in favour of this one.

Once the migration numbers are bumped and the scheduler conflict is resolved this should be good to go.

dginther:
Fixed the migration numbers. Looking more deeply into the scheduler.go since it was a significant refactor.

dginther:
I did a merge dry-run and it looks like everything is A-OK for merge.

vavallee (commented):
Thorough review complete. Overall quality is high — the Transmission client, adapter pattern, scanner refactor, and Prowlarr fix are all well-executed. Approving pending one blocking issue.

---

**Blocking: scheduler.go removes recommendation engine wiring**

The PR's `scheduler.go` diff deletes the `RecommendationEngine` interface, `WithRecommender()`, and the 24h recommendation job that landed in development last night. A direct merge would silently drop the recommendation engine from the scheduler. Please rebase on `development` to pick up those additions — they're orthogonal to your changes (different struct fields, different cron block) so the rebase should be clean.

---

**Migrations — ✅ already fixed**

016/017 renaming is in place. No collision with `014_book_excluded.sql` or `015_recommendations.sql` on development.

**Transmission 409 handling — ✅**

Session-ID challenge/retry is correct. `doRequest()` extracts the header on 409, stores it under a mutex, and retries transparently. Both success and failure paths are covered in tests.

**SSRF / security — ✅**

`validateRequestTarget()` pins all requests to the configured RPC URL and rejects redirects. Cleaner than most client implementations.

**Adapter dispatch — ✅**

`SendDownload()`, `TestClient()`, `RemoveDownload()` dispatch correctly to all three clients. The absolute-path guard for Transmission's `download-dir` is correct.

**Scanner refactor — ✅**

Existing SABnzbd/qBittorrent polling behaviour is preserved. The `checkTransmissionDownloads()` / `checkQbittorrentDownloads()` helpers mirror each other correctly. `setDownloadError()` / `updateDownloadStatus()` abstractions are cleaner than the previous inline DB calls.

**qBittorrent signature change — ✅**

`AddTorrent` now returns the torrent hash. Hash extraction tries the magnet `xt` param first, then falls back to a before/after `GetTorrents()` snapshot diff. All call sites updated.

**Error handling across db/ — ✅**

`LastInsertId()` errors now checked, explicit column lists in SELECT, consistent `fmt.Errorf(...: %w)` wrapping throughout.

**Test coverage — ✅**

Transmission (10+ cases including 409 retry, redirect rejection), adapter (26 functions across all 3 clients), newznab Prowlarr URL normalization. Solid.

**Minor (non-blocking)**

Migration 016's backfill UPDATEs are silent no-ops if rows don't match — benign, just worth knowing.

---

Once rebased on `development` this is good to merge.

vavallee:
Rebased onto `development` in #154 — closing this in favour of that. All your commits are preserved, full credit maintained. Thank you for the contribution!

Issue #97, opened 2026-04-15
test(smoke): add post-deploy smoke test suite


vavallee:
## Why
Unit tests cover per-package correctness but nothing exercises a running Bindery against real HTTP + SQLite + filesystem the way a user (or Docker deploy) does. A regression in wiring — a missed migration, a bad route registration, an embed path — currently only surfaces after deploy. A small smoke suite would catch these in CI before the image ships.

## Scope
A new `test/smoke/` package (or `Makefile` target `make smoke`) that:

1. Boots the real `bindery` binary against a scratch `/tmp` data dir
2. Waits for `/healthz` (or equivalent) to return 200
3. Exercises the critical golden paths via the public HTTP API:
   - `GET /api/v1/authors` (empty → 200, shape OK)
   - `POST /api/v1/authors` add by OL ID → 201 → appears in list
   - `GET /api/v1/books` → 200
   - `POST /api/v1/calibre/import` with no library configured → 400 (fast negative check)
   - `GET /api/v1/settings` → round-trip a setting value
   - `GET /` → 200 and returns embedded index.html (catches broken frontend embed)
4. Shuts down cleanly

## Non-goals
- Not a replacement for unit tests
- Not a full E2E with real Readarr/Calibre data — stub metadata providers
- Not a browser-driven test (that's a separate Playwright effort if we want it)

## Where it runs
- Add to `.github/workflows/ci.yml` as a new job gated on Go tests passing
- `make smoke` locally

## Acceptance
- `make smoke` completes under 30s on a clean checkout
- Failure on any of the checks above fails CI
- Smoke suite does not share state with unit tests (own tempdir, own port)

Related: all feature PRs (including the just-merged #72 Calibre import) would benefit from a post-boot sanity check.

PR #134, opened 2026-04-16
fix(importer): use os.Root for TOCTOU-safe dir copy/hardlink (G122)

vavallee:
## Summary

- `copyDir` used `filepath.Walk`, which follows symlinks at enumeration time. Between Walk visiting a path and `copyFile` opening it, a symlink could be swapped in — causing a read/write outside the intended directory tree. Replaced with `os.Root`-scoped traversal (Go 1.24+): the kernel enforces that all I/O stays within the opened root.
- `HardlinkDir` had the same pattern with `filepath.WalkDir`. Replaced with root-scoped `ReadDir` + `os.Link` using paths derived from the enumerated root.
- Symlinks and non-regular entries are now skipped via `DirEntry.Type()` checks during root-scoped `ReadDir`, not via racy stat calls.

Closes gosec G122 (code scanning alert #97).

## Test plan

- [x] `go test ./internal/importer/...` passes (existing helpers_test.go covers copyFile, copyDir, MoveDir, HardlinkDir)
- [ ] CI green

### DIFF
--- a/internal/importer/renamer.go
+++ b/internal/importer/renamer.go
@@ -3,7 +3,6 @@ package importer
 import (
 	"fmt"
 	"io"
-	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


vavallee:
Merge conflicts against main — needs rebase. G122 TOCTOU fix is still needed; tagging for v0.15.0.

Issue #76, opened 2026-04-15
style(books): round the selection checkboxes to match Authors


vavallee:
## What
Match the Authors page style change (#75) on the Books page so the selection UI stays consistent across list views.

## Where
`web/src/pages/BooksPage.tsx` — three spots:
- line 179 — select-all checkbox in table header
- line 202 — per-row checkbox in table view
- line 253 — per-card checkbox in grid view

## How
Apply whichever approach is chosen in #75 (minimal `rounded-full` swap, or full custom peer-styled circle) so both pages stay visually identical.

Depends on #75.

vavallee:
Already matching — `BooksPage.tsx` checkboxes at lines 185, 208, 259 all carry `rounded-full` identical to the `AuthorsPage.tsx` equivalents added in #75. No further change needed; re-open if a visual diff surfaces.

PR #135, opened 2026-04-16
fix(api): handle discarded errors in queue, migrate, library (G104)

vavallee:
## Summary

gosec flagged eight sites where method return values were silently discarded:

- **queue.go** (4 sites): `SetError`, `UpdateStatus`, `SetNzoID` on the download repo — fire-and-forget DB writes in the grab handler. Now logged at error level so a failing database surfaces in logs.
- **migrate.go** (2 sites): `tmp.Close()` on the Readarr import temp file. Error path uses `_ =` (file is about to be removed); success path checks and logs.
- **library.go** (1 site): `w.Write` on a pre-serialised JSON string. Explicitly discarded with `_, _ =` — the only possible error is a disconnected client, which net/http already handles.

Closes gosec G104 alerts #71–74, #76–78.

## Test plan

- [x] `go test ./internal/api/...` passes
- [ ] CI green

### DIFF
--- a/internal/api/migrate.go
+++ b/internal/api/migrate.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


vavallee:
Merge conflicts against main — needs rebase. G104 error-handling fixes are still worth landing; tagging for v0.15.0.

PR #147, opened 2026-04-16
fix(imageproxy): sharded cache dirs, eviction, and atomic writes (#138)

### LINKED ISSUE #138: Image proxy cache: harden for large libraries
## Context

The image proxy cache (`internal/api/imageproxy.go`) works correctly for typical homelab libraries but has four scaling gaps that become relevant as libraries grow past a few thousand books.

### 1. Sharded directory structure
**Problem:** All cached files land in a single flat `image-cache/` directory. ext4/btrfs degrade on `readdir` and `stat` past ~100k entries. A 50k-book library with author images produces ~150k files (image + `.ct` sidecar + potential duplicates).

**Fix:** Shard by the first 2 hex chars of the SHA-256 key: `image-cache/a3/a3f7b2...` instead of `image-cache/a3f7b2...`. 256 subdirectories, ~600 files each at 150k total. Migration: background goroutine on startup moves existing flat files into shards, then removes the old layout.

### 2. Cache eviction
**Problem:** The 30-day TTL is checked on read (stale → refetch), but expired files are never deleted. The cache directory grows without bound — ~100 KB average per cover means 10k images ≈ 1 GB, 100k images ≈ 10 GB, with no ceiling.

**Fix:** Periodic sweep (daily or on startup) that deletes files older than the TTL. A simple `filepath.WalkDir` with `ModTime` check is sufficient — no index or database needed. Expose the cache size in `/api/v1/system/status` so users can see it in the UI.

### 3. Atomic writes
**Problem:** Two concurrent requests for the same uncached image both fetch upstream and race to write the same file. `os.WriteFile` truncates before writing, so a concurrent reader can see a partial/empty file.

**Fix:** Write to a temp file in the same directory, then `os.Rename` (atomic on POSIX). The last writer wins, which is fine — the content is identical.

```go
tmp, err := os.CreateTemp(h.cacheDir, ".tmp-*")
// write to tmp...
tmp.Close()
os.Rename(tmp.Name(), imgFile)
```

### 4. Streaming response instead of full body buffer
**Problem:** `io.ReadAll` loads the entire image (up to 10 MB) into memory before writing to cache and responding. 50 concurrent cache misses hold ~500 MB in RAM.

**Fix:** `io.TeeReader` to write to the cache file while streaming to the HTTP response simultaneously. Combine with the atomic-write pattern (tee into a temp file, rename on completion, discard on error).

## Priority

Items 1–3 are straightforward and low-risk. Item 4 is a slightly larger refactor (error handling when the upstream read fails mid-stream after headers are already sent). Recommend landing 1–3 together, 4 separately.

## Scale reference

| Library size | Cache files (est.) | Disk (est.) | Current impact |
|---|---|---|---|
| 1,000 books | ~3,000 | ~300 MB | Fine |
| 10,000 books | ~30,000 | ~3 GB | Flat dir starts to slow |
| 50,000 books | ~150,000 | ~15 GB | Flat dir degrades; no eviction is a problem |

vavallee:
## Summary
- Shard image cache into 256 subdirectories (`image-cache/a3/a3f7b2…`) to prevent flat-directory bloat on large libraries.
- Add one-time background migration that moves existing flat-layout files into the new sharded structure on startup.
- Replace `os.WriteFile` with atomic tmp+rename pattern so concurrent readers never see partially-written cache entries.
- Add `StartEviction(24h)` background goroutine that walks the cache tree and purges files older than the 30-day TTL.
- Expose `CacheSize()` and surface `imageCacheBytes` in `GET /api/v1/system/status`.

Closes #138

## Test plan
- [ ] Start Bindery with an existing flat image cache — verify files are migrated into `image-cache/<xx>/` subdirectories
- [ ] Browse author/book pages — confirm cover images still load from the sharded cache
- [ ] Check `GET /api/v1/system/status` returns `imageCacheBytes` field
- [ ] Manually backdate a cached file's mtime past 30 days — verify the eviction goroutine removes it on next tick

### DIFF
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 	"io/fs"
 	"log/slog"
 	"net/http"
@@ -265,6 +266,7 @@ func main() {
 	})
 	recHandler := api.NewRecommendationHandler(recRepo, recEngine, authorRepo, bookRepo, sched)
 	imageProxyHandler := api.NewImageProxyHandler(cfg.DataDir)
+	imageProxyHandler.StartEviction(24 * time.Hour)
 	migrateHandler := api.NewMigrateHandler(
 		authorRepo, indexerRepo, dlClientRepo, blocklistRepo, bookRepo, metaAgg,
 		// Bulk imports always populate the catalogue but never auto-grab.
@@ -294,7 +296,11 @@ func main() {
 		})
 		r.Get("/system/status", func(w http.ResponseWriter, _ *http.Request) {
 			w.Header().Set("Content-Type", "application/json")
-			_, _ = w.Write([]byte(`{"version":"` + version + `","commit":"` + commit + `","buildDate":"` + date + `"}`))
+			cacheBytes, _ := imageProxyHandler.CacheSize()
+			_, _ = fmt.Fprintf(w,
+				`{"version":"%s","commit":"%s","buildDate":"%s","imageCacheBytes":%d}`,
+				version, commit, date, cacheBytes,
+			)
 		})
 
 		// Auth — status/login/logout/setup are always allowed through the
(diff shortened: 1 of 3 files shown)


vavallee:
CI failing — `lint`, `gosec`, `SAST – Go`, `validate`, and `smoke` all red on the latest push. Needs attention before this can merge.

Targeting for v0.15.0.

PR #345, opened 2026-04-22
chore(deploy): update bindery to v1.1.6

vavallee:
Bump Helm chart image tag to `1.1.6`.

Fixes Discover page always empty — `OwnedForeignIDs` was marking every book (including `wanted`/`skipped`) as owned, so all candidate generators returned zero results.


vavallee:
Superseded — prod is on 1.2.0.

PR #371, opened 2026-04-23
ABS Import Feature Request

magrhino:
## Summary

- add Audiobookshelf (ABS) import support end to end, including config/test/discovery APIs, async import runs, rollback preview/rollback, and review/conflict workflows
- persist ABS provenance, import runs, review items, and metadata conflicts with new DB repos and migrations, and wire the importer into app startup/routes
- add the ABS settings UI and author-conflict review panel, plus unit/API/contract coverage and a consolidated ABS import guide

## Checklist

- [x] Tests added or updated
- [x] `docs/ABS-Import-Wiki.md`  Wiki pages updated if user-facing behaviour changed

## Test plan


vavallee (commented):
Nice work on this — rollback, conflict queue, contract tests, it's a lot to ship cleanly.

Three blockers before merge:

1. **Rollback doesn't restore updated metadata.** It correctly deletes entities the import created, but if it overwrote existing book metadata there's no snapshot to revert to —  stores what was written, not what was there before. Either capture a before-snapshot at import time, or explicitly scope rollback to created-only and say so in the UI.

2. **Migration 033 isn't idempotent.** 032 creates `abs_review_queue`, then 033 does bare `ALTER TABLE … ADD COLUMN`. If the upgrade dies between those two, a re-run fails. Use `IF NOT EXISTS` or just fold 033 into 032 while it's still unreleased.

3. **Review/conflict endpoints are unbounded.** `GET /abs/review` and `GET /abs/conflicts` return everything with no limit. Add `limit`/`offset` now — much harder once the response shape is relied on.

magrhino:
> * bare `ALTER TABLE … ADD COLUMN`. If the upgrade dies between those two, a re-run fails. Use `IF NOT EXISTS` or just fold 033 into 032 while it's still unreleased.
> * **Review/conflict endpoints are unbounded.** `GET /abs/review` and `GET /abs/conflicts` return everything with no limit. Add `limit`/`offset` now — much harder once the response shape is relied on.



> Nice work on this — rollback, conflict queue, contract tests, it's a lot to ship cleanly.
> 
> Three blockers before merge:
> 
> 1. **Rollback doesn't restore updated metadata.** It correctly deletes entities the import created, but if it overwrote existing book metadata there's no snapshot to revert to —  stores what was written, not what was there before. Either capture a before-snapshot at import time, or explicitly scope rollback to created-only and say so in the UI.
> 2. **Migration 033 isn't idempotent.** 032 creates `abs_review_queue`, then 033 does bare `ALTER TABLE … ADD COLUMN`. If the upgrade dies between those two, a re-run fails. Use `IF NOT EXISTS` or just fold 033 into 032 while it's still unreleased.
> 3. **Review/conflict endpoints are unbounded.** `GET /abs/review` and `GET /abs/conflicts` return everything with no limit. Add `limit`/`offset` now — much harder once the response shape is relied on.

I'll get to work on these!

magrhino:
Thanks for the careful review. I believe the merge blockers are addressed now.

- Rollback now preserves and restores the prior metadata for existing books that were updated during import, while still deleting entities created by the import.
- Migration 033 has been made idempotent / safe to rerun.
- GET /abs/review and GET /abs/conflicts now support bounded pagination via limit / offset.

_One note on the size of the follow-up diff: fixing metadata rollback exposed some additional author matching issues, since restoring updated book metadata depends on correctly identifying existing authors versus imported authors. I tightened that matching path and added coverage around it, which is why the change grew by 3000+ lines. I believe that work is necessary to make the rollback behavior correct rather than only superficially passing the blocker._

Happy to adjust further if you see anything off in the new approach.

vavallee (commented):
Hey, solid work overall — the author-matching pipeline and checkpoint/rollback design are both well thought out. Few things before this lands:

**Must fix:**

- `go-version: "1.25.9"` in the `abs-contract` CI job — Go 1.25 doesn't exist, so that job is broken on every push. Pin to `"1.23.x"` or `"stable"`.
- `shouldRetry` in `client.go` falls through to an unconditional `return true` after the `net.Error` check. Permanent errors (bad cert, NXDOMAIN) will retry forever.
- `clientFromProbe` accepts arbitrary `baseUrl` from the POST body — that's an admin-gated SSRF. At minimum reject RFC-1918 / link-local, or just read from stored config only.
- Remove `tmp-prompt.md` from the repo root.
- `docs/abs_import.md` lines 356–382 have `/Users/ryanjones/Documents/...` hardcoded — strip those out.

**Soon after:**

- `findAuthorByName` calls `authors.List` + `aliases.List` on every imported item. Full table scans per item will hurt anyone with a decent-sized library. Load once at run start and pass through.
- `defer resp.Body.Close()` inside the retry loop — defers stack up per iteration. Close explicitly at the top of each iteration instead.
- `Progress.Results` grows unbounded — cap it (last 100 or so) or page it through the run-entity table that already exists.
- `mustJSON` silently returns `"{}"` on error — at least log it, rollback snapshots shouldn't fail quietly.
- `ResumeInterrupted` marks the old run failed *after* `Start()`, not before. Flip the order.
- `DEPLOYMENT.md` checkbox is unchecked — five migrations, six new tables, and `BINDERY_ABS_ENABLED` all need a note there.

Minor nit: `BINDERY_ABS_ENABLED` defaults to `true` (opt-out). Might be worth flipping to opt-in for the first release so existing installs don't get ABS routes activated on upgrade without knowing.

Otherwise the architecture is in good shape — once the CI version and retry bugs are fixed this'll be close.

vavallee:
Taking off for the week but will try to check in on your progress, quite the feature!

magrhino:
Enjoy your break. I believe I have all the fixes you requested working there now. May do more things on a different branch while you are away.

## Summary

- ABS contract CI now uses go-version: stable.
- ABS retry handling now only retries timeout transport errors, and response bodies are closed explicitly per attempt.
- ABS probe endpoints now use the saved base URL only; the POST body can override the API key but not the target URL.
- Removed tmp-prompt.md and stripped the hardcoded local paths from the docs anyway you could remove my name from that comment sorry about that.
- The importer now builds the author matcher once per run, caps live progress results at 100, logs/returns metadata JSON encoding failures, and marks interrupted runs failed before starting the resumed run.
- docs/DEPLOYMENT.md now documents the ABS migrations and BINDERY_ABS_ENABLED, and the feature flag defaults to opt-in.
- I also resolved the merge conflicts

### Validation: go test ./cmd/... ./internal/... ./tests/abscontract/... passes.

tobiraak:
Hi, am I understanding this feature correctly, that I would be able to download audiobooks via Bindery and they are automatically added to my ABS library or is this feature the other way around? 
I am looking for a way to have an easy search -> download -> import to ABS workflow and was wondering if this is it.
Thanks already in advance!

vavallee (commented):
## Review

### Overview

Full Audiobookshelf import pipeline: HTTP client with retry/backoff, paged enumeration with checkpoint-aware resume, async importer with dry-run/rollback/review-queue/conflict-resolution, 14 new API routes, 5 DB migrations, feature flag, and a pinned contract test suite. This is a large but coherent feature addition.

---

### What's done well

- **API key handling**: `GetConfig` returns only `APIKeyConfigured: bool` — the raw key is never returned to the UI. ✓
- **Async context**: `context.WithoutCancel(r.Context())` correctly used when starting the import goroutine, so the long-running import survives request completion. ✓
- **Access control**: All ABS config/import routes are behind `auth.RequireAdmin`. ✓
- **Feature flag**: `BINDERY_ABS_ENABLED=false` by default — zero-risk rollout for existing deployments. ✓
- **Idempotency**: Provenance table ensures reruns don't duplicate entities. ✓
- **Checkpoint/resume**: Interrupted imports pick up from the last saved page+item. ✓
- **Retry policy**: `shouldRetry` returns true only for timeout errors, not for context cancellation or permanent failures. ✓
- **Test coverage**: Unit, contract (pinned 2.33.2 fixtures), API handler stubs, and an integration-capable `tests/abscontract` suite.

---

### Issues

**`importer.go` is 4021 lines**

This is the biggest concern. It's by far the largest file in the codebase and bundles import orchestration, rollback planning, metadata conflict detection, series/edition upserts, file reconciliation, and snapshot/provenance recording all in one unit. Future changes to any one of these will be hard to review in isolation. Rollback and conflict resolution in particular are cleanly separable and should be their own files.

**`SetConfig` makes a live ABS API call when `libraryID` is set**

If ABS is unreachable, `PUT /api/v1/abs/config` returns an error even for changes that don't need a live validation (e.g., updating the label, toggling `enabled`, clearing the library ID). Users who configure Bindery before their ABS instance is online will hit this unexpectedly. Consider validating library ID only when explicitly requested (e.g., `POST /api/v1/abs/test`) rather than on every save.

**Stale user-agent**

`defaultUserAgent = "bindery-abs-phase1/1"` — the `phase1` string is a development-iteration artifact that will show up in ABS server logs. Should be `"bindery/<version>"` consistent with the rest of the project.

**`go.mod` unexplained downgrade**

`go 1.25.9` → `go 1.25.0`. The `go` directive in `go.mod` is a minimum version requirement — downgrading it means the module now claims to be compatible with an older toolchain than was previously required. The PR doesn't explain why. If the intent was to align with `go-version: stable` in CI, this is a separate concern and the go.mod should stay at the highest version already tested.

**`go-version: stable` in CI (7 jobs)**

Removes the pinned `1.25.9` for `stable` across every job. `stable` is whatever Go considers current at run time — a new toolchain release that introduces a regression will silently break CI with no change to the repo. Minor-version pinning (`1.25`) is a reasonable middle ground: stays current on patches, bounded on breaking changes.

**`docker-compose.yml` added to repo root with `/tmp/books/:/books`**

The added `docker-compose.yml` uses `/tmp/books/:/books` and no `BINDERY_DB_PATH`. This is clearly a development fixture but sits in the project root where it'll be copied by deployers following the README. Either document it clearly as a dev-only example or move it to `docs/` or `examples/`.

**`contextBackground()` in `LoadABSConfig`**

`LoadABSConfig` is called from both startup code and HTTP handlers, but uses a background context internally for settings reads. The handler path should propagate `r.Context()` so that request cancellation/deadlines apply to the DB reads. Minor, but worth fixing for consistency with the rest of the handlers.

**No-op migration 033**

Documented as intentional for branch coherence. Acceptable — the comment in the migration file is sufficient.

---

### Summary

Solid feature with good security handling, proper async patterns, and meaningful test coverage. Main asks before merge: decompose `importer.go`, fix the `SetConfig` live-validation UX, and clean up the `phase1` user-agent and `docker-compose.yml`. The CI `go-version: stable` and `go.mod` downgrade are worth a quick explanation.

magrhino:
> Hi, am I understanding this feature correctly, that I would be able to download audiobooks via Bindery and they are automatically added to my ABS library or is this feature the other way around? 
> 
> I am looking for a way to have an easy search -> download -> import to ABS workflow and was wondering if this is it.
> 
> Thanks already in advance!

Hi the feature right now is for existing libraries to import the libraries into bindery for management. I do know how to auto import also, and I'm planning on doing that in a later branch. But for now just ensuring the audiobook is under author/title should auto import most of the time if you have hard links set up!

magrhino:
Thanks for the latest review. I’ve kept this follow-up scoped to PR-readiness cleanup rather than adding new ABS behavior.

What changed:
- Decomposed `internal/abs/importer.go` into focused ABS importer files for orchestration, author matching, upserts, file reconciliation, conflicts, rollback, snapshots, shared types, and utilities.
- Changed ABS config saves so `PUT /api/v1/abs/config` stores normalized local settings without live ABS validation. Explicit probe routes still perform live checks.
- Propagated request context through `LoadABSConfig` for handler paths.
- Replaced the stale `bindery-abs-phase1/1` user-agent with `bindery/<version>` and `bindery/dev` for unversioned builds.
- Restored `go.mod` to `go 1.25.9` and pinned CI to the Go `1.25` minor line instead of `stable`.
- Removed the root dev-only `docker-compose.yml` and updated the ABS docs to reflect the current behavior.
-  I also kept the import path aligned with book-library selection by rejecting non-book ABS responses before scanning

Validation:
- `wc -l internal/abs/importer.go`: `760`
- `go test ./internal/abs ./internal/api ./cmd/bindery`: passed
- `go test ./cmd/... ./internal/... ./tests/abscontract/...`: passed
- `cd web && npm run build`: passed
- `gofmt -l internal/abs internal/api cmd/bindery`: no files returned
- `rg "phase1|bindery-abs-phase1"`: no matches
- `rg "go-version:\s*stable|go-version:\s*\"stable\"|go-version:\s*'stable'" .github/workflows`: no matches

PR #375, opened 2026-04-23
fix(downloader): apply url_base to download-client base URL (#369)

### LINKED ISSUE #369: Download client ignores url_base field — reverse proxy subpath not applied
### Pre-flight

### Current behaviour

The `url_base` column exists in the `download_clients` DB table, but the downloader URL builder never reads it when constructing the connection URL. Setting `url_base` to `/qbittorrent` (for example) has no effect — only the `host` field is used.

### Expected behaviour

When `url_base` is set, the downloader should construct: `http(s)://host:port/url_base/api/...`

This is the standard pattern for *arr-compatible clients running behind a reverse proxy with a subpath.

### Steps to reproduce

1. Set a download client with Host `192.168.1.10`, Port `8080`, URL Base `/qbit`
2. Observe: Bindery connects to `http://192.168.1.10:8080/api/...` — the `/qbit` prefix is ignored
3. Expected: `http://192.168.1.10:8080/qbit/api/...`

### Root cause

The downloader client URL builders (e.g. `internal/downloader/qbittorrent/client.go`) build their base URL from `host` and `port` only. The `url_base` field is fetched from the DB and stored on the struct but never interpolated into the URL.

### Suggested fix

In each downloader client's URL construction:
```go
base := fmt.Sprintf("%s://%s:%d%s", scheme, cfg.Host, cfg.Port, cfg.URLBase)
```

Strip any trailing slash from `cfg.URLBase` and ensure it starts with `/` (or is empty).

### Deployment method

Docker (ghcr.io)

### Bindery version

image:latest

### Reporter

WillowMist23 via #support (2026-04-23). Related to #364 (SSL URL construction).

SAY-5:
## Summary

Every downloader client (qbittorrent, transmission, deluge, nzbget, sabnzbd) built its connection URL from `host:port` only. The `url_base` column is read from the DB into `DownloadClient.URLBase` but never interpolated into the URL, so operators running an *arr-compatible client behind a reverse-proxy subpath (e.g. Host `192.168.1.10:8080`, URL Base `/qbit`) saw Bindery connect to `http://192.168.1.10:8080/api/...` instead of `http://192.168.1.10:8080/qbit/api/...`.

## Fix

- New `internal/downloader/urlbase` helper that normalises user input: trims whitespace, enforces a single leading slash, strips trailing slashes, and tolerates a pasted full URL by keeping only the path segment.
- Thread a `urlBase` parameter through each client's `New` constructor and apply the normalised prefix to `baseURL` / Transmission's RPC `Path`.
- Update every call site in `internal/downloader/adapter.go` and `internal/importer/scanner.go` to pass `client.URLBase`.
- Update all in-package test helpers so the signature changes stay compile-clean.
- Empty / whitespace `url_base` collapses to `""`, preserving the pre-#369 URL shape for deployments without a proxy.

Fixes #369

## Test plan

- [x] `go build ./...`
- [x] `go test ./...` — full suite green, including the existing per-client tests (deluge / nzbget / qbittorrent / sabnzbd / transmission) and the new `internal/downloader/urlbase` table-test
- [x] Manually traced each `New` call site in `adapter.go` and `scanner.go` to confirm `client.URLBase` is wired in

### DIFF
--- a/internal/downloader/adapter.go
+++ b/internal/downloader/adapter.go
@@ -47,19 +47,19 @@ func ProtocolForClient(clientType string) string {
 func TestClient(ctx context.Context, client *models.DownloadClient) error {
 	switch client.Type {
 	case "transmission":
-		trans := transmission.New(client.Host, client.Port, client.Username, client.Password, client.UseSSL)
+		trans := transmission.New(client.Host, client.Port, client.Username, client.Password, client.URLBase, client.UseSSL)
 		return trans.Test(ctx)
 	case "qbittorrent":
-		qb := qbittorrent.New(client.Host, client.Port, client.Username, client.Password, client.UseSSL)
+		qb := qbittorrent.New(client.Host, client.Port, client.Username, client.Password, client.URLBase, client.UseSSL)
 		return qb.Test(ctx)
 	case "deluge":
-		dl := deluge.New(client.Host, client.Port, client.Password, client.UseSSL)
+		dl := deluge.New(client.Host, client.Port, client.Password, client.URLBase, client.UseSSL)
 		return dl.Test(ctx)
 	case "nzbget":
-		ng := nzbget.New(client.Host, client.Port, client.Username, client.Password, client.UseSSL)
+		ng := nzbget.New(client.Host, client.Port, client.Username, client.Password, client.URLBase, client.UseSSL)
 		return ng.Test(ctx)
 	default:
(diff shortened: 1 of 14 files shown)


vavallee (approved):
Clean implementation — dedicated `urlbase` package is the right call, and the paste-full-URL normalisation is a nice UX touch that the other approach missed. All checks look good on our side; approving to trigger CI.

Issue #426, opened 2026-05-03
queue: liveStatusIsError always false for Transmission


vavallee:
## Problem

\`liveStatusIsError\` checks whether a live status string contains \`\"error\"\` or \`\"fail\"\`:

\`\`\`go
func liveStatusIsError(s string) bool {
    s = strings.ToLower(s)
    return strings.Contains(s, "error") || strings.Contains(s, "fail")
}
\`\`\`

For Transmission, \`LiveStatus.Status\` is populated via \`strconv.Itoa(t.Status)\`, which produces integer strings like \`\"0\"\`, \`\"2\"\`, \`\"16\"\`. None of these will ever match \`\"error\"\` or \`\"fail\"\`, so Transmission error states (status 16 = error, status 32 = isolated error) are silently treated as healthy in the queue.

## Suggested fix

Add a Transmission-specific error check using the known error status codes (16, 32) before falling through to the string-contains check, or map Transmission integer statuses to a canonical string before storing them in \`LiveStatus.Status\`.

vavallee:
Fixed in v1.2.7 by PR #434 — `liveStatusIsError` now handles Transmission's integer status codes (16, 32) via `strconv.Atoi`.

PR #403, opened 2026-04-26
Issue/370

magrhino:
## Summary

- Adds `GET /api/queue`, an Arr-compatible queue endpoint for Harpoon and similar external tools.
- Returns queue records with live downloader-backed `size`, `sizeleft`, status, client name, remote download ID, and protocol where available.
- Keeps the existing `GET /api/v1/queue` UI response shape unchanged.
- Adds focused queue and downloader tests, plus an `[Unreleased]` changelog entry.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed
- [x] `CHANGELOG.md` `[Unreleased]` section updated
- [ ] Wiki pages updated if user-facing behaviour changed

## Test plan

- [x] `go test ./cmd/... ./internal/...`
- [x] `cd web && npm run build`

PR #434, opened 2026-05-03
fix: queue RemoteID normalization and Transmission error state detection

### LINKED ISSUE #425: queue: storedDownloadID not lowercased on initial RemoteID assignment
## Problem

In \`internal/api/queue.go\`, the \`RemoteID\` field of a queue record is set directly from the stored torrent hash without normalising case:

\`\`\`go
RemoteID: item.StoredDownloadID,
\`\`\`

Elsewhere in the codebase torrent hashes are lowercased before comparison. If the stored ID was persisted in mixed or upper case (e.g. from an older client version), Harpoon or another Arr consumer may fail to match the ID against its own records.

## Suggested fix

Lowercase the value at assignment:

\`\`\`go
RemoteID: strings.ToLower(item.StoredDownloadID),
\`\`\`

vavallee:
Fixes #425 and #426 together — both are in the queue status layer.

### #425 — RemoteID not lowercased
`RemoteID` was assigned directly from `StoredDownloadID` without normalising case. Torrent hashes from older client versions or mixed-case sources caused ID mismatches in downstream consumers (Harpoon, etc.).

**Fix:** `strings.ToLower(item.StoredDownloadID)` at the assignment site — applied in both the queue `List` handler (lookup path) and `grab` / scheduler (write path) so the DB always stores lowercase torrent IDs.

### #426 — Transmission error states invisible
`liveStatusIsError` checks whether a status string contains `"error"` or `"fail"`. Transmission populates `LiveStatus.Status` via `strconv.Itoa(t.Status)`, so error states (status 16 = error, status 32 = isolated-error) produce integer strings that never match the check. Downloads stuck in a Transmission error state were silently treated as healthy in the queue UI.

**Fix:** Added integer-aware detection for Transmission's known error status codes. Also added `Status` field to `LiveStatus` and populated it from `t.Status` in `getTorrentLiveStatuses` for Transmission.

## Testing
- Existing queue tests pass
- Added `TestLiveStatusIsError_TransmissionIntegerCodes` — covers integer codes 16/32 and common string statuses
- Added `TestGetLiveStatusesTransmission_PopulatesStatus` — verifies `Status` field is populated and `liveStatusIsError` catches it
- Added `TestQueueListLiveOverlayTransmission_TorrentIDLowercased` — verifies numeric TorrentID lookup works after normalisation
- Added `TestQueueListLiveOverlayQbittorrent_MixedCaseHashNormalized` — verifies mixed-case hash in DB matches lowercased live status map key

### DIFF
--- a/internal/api/queue.go
+++ b/internal/api/queue.go
@@ -92,7 +92,7 @@ func (h *QueueHandler) List(w http.ResponseWriter, r *http.Request) {
 			if item.TorrentID == nil {
 				continue
 			}
-			remoteID = *item.TorrentID
+			remoteID = strings.ToLower(*item.TorrentID)
 		} else {
 			if item.SABnzbdNzoID == nil {
 				continue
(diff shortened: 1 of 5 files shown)


vavallee:
Merged via rebased branch after resolving conflict with #403 (EnrichedQueueItem.Download wrapper). All changes on main.

Issue #423, opened 2026-05-03
legacyCredentialURLBase: password==apiKey arm can silently clear a legitimate bare url_base


vavallee:
### Context

`legacyCredentialURLBase` in `internal/db/download_clients.go` classifies a stored row as a legacy credential (and clears its `url_base`) when any of three conditions hold:

```go
return username == "" || username == urlBase || (password != "" && password == apiKey)
```

The third arm fires whenever the stored `password` equals the stored `api_key`, regardless of what `url_base` contains. A user with a legitimate bare `url_base` (e.g. `qbit`) who also happens to have `password == api_key` (duplicate entry, copy-paste error, or a client type that uses the same value for both fields) would have their `url_base` silently cleared on the next read.

Slash-prefixed paths (`/qbit`) are already safe because the function returns `false` for those. The risk only applies to bare url_bases without a leading slash.

### Fix

Add a comment to the function documenting the invariant each arm is protecting, and consider whether the `password == apiKey` arm is still needed now that new rows are saved correctly. If it must stay, scope it more tightly — for example, only apply it when `username == urlBase` is also true, which is the true signature of an old migrated row.

vavallee:
Fixed in v1.2.7 by PR #437 — the `password==apiKey` arm is removed from `legacyCredentialURLBase`; detection is now based on `urlBase` not starting with `/` and `apiKey` being non-empty.

PR #437, opened 2026-05-03
fix: credential normalization silent clear and missing write-path test

### LINKED ISSUE #422: download_clients: normalizeClientCredentialStorage write-path has no test coverage for the APIKey guard
### Context

PR #402 added `&& c.APIKey != ""` to the `normalizeClientCredentialStorage` function (the save/update path in `internal/db/download_clients.go`) to prevent migrating a bare `url_base` value into `username` when no legacy API key is present. The corresponding read-path fix (`hydrateClientCredentials`) gained four new tests, but the write-path fix has none.

### Risk

A regression in `normalizeClientCredentialStorage` would silently corrupt new rows on save — a client saved with `url_base=qbit`, `username=""`, `api_key=""` would have its `url_base` moved into `username` and cleared, breaking the reverse-proxy path for that client.

### Fix

Add a test that creates a download client via the repo's `Create`/`Update` path with a bare `url_base` and no `api_key`, then reads it back and asserts `url_base` is preserved and `username` is not populated from it.

vavallee:
Fixes #422 and #423.

### #423 — legacyCredentialURLBase over-fires on modern rows

The previous inline logic in `hydrateClientCredentials` treated any qBittorrent/Transmission row with an empty `username` and a non-empty `url_base` as a legacy migrated credential row, regardless of whether `api_key` was populated. A user with a bare `url_base` like `qbit` and an empty `api_key` (a completely valid modern configuration) would have their `url_base` silently migrated into `username` on every read — effectively corrupting the stored value.

The fix extracts a `legacyCredentialURLBase(username, urlBase, apiKey string) bool` helper and tightens the detection: a genuine legacy row always has a non-empty `api_key` (the old schema stored the password there). An empty `api_key` is definitive evidence of a modern row with a real `url_base` path, so migration is skipped. The same guard is applied on the write path (`normalizeClientCredentialStorage`) for consistency.

### #422 — No test for normalizeClientCredentialStorage write-path

The write-path guard had no test coverage. A regression there would silently corrupt new download client rows on save without any test catching it.

Added `TestNormalizeClientCredentialStorageWritePath`: creates a qBittorrent client with bare `url_base = "qbit"` and empty `api_key`, saves it via `repo.Create`, reads it back, and asserts `url_base` is preserved as `"qbit"` and `username` is not populated from it.

Added `TestLegacyCredentialURLBase`: unit-tests all detection arms of the helper including the core regression case (`bare url_base + empty api_key → false`).

## Testing

All existing credential normalization tests pass. Two new tests added and green.

### DIFF
--- a/internal/db/db_test.go
+++ b/internal/db/db_test.go
@@ -624,6 +624,90 @@ func TestDownloadClientRepoCredentialFields(t *testing.T) {
 	}
 }
 
+// TestNormalizeClientCredentialStorageWritePath covers the write-path guard in
+// normalizeClientCredentialStorage: a qBittorrent client saved with a bare
+// url_base and an empty api_key must read back with url_base preserved and
+// username NOT populated from it. (closes #422)
+func TestNormalizeClientCredentialStorageWritePath(t *testing.T) {
+	database, err := OpenMemory()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer database.Close()
+
+	ctx := context.Background()
+	repo := NewDownloadClientRepo(database)
+
+	// A client with a bare url_base path and no api_key — the write-path guard
+	// (normalizeClientCredentialStorage) must not migrate url_base into username
+	// because api_key is empty, which means this is not a legacy credential row.
+	qbt := &models.DownloadClient{
+		Name:    "qBittorrent bare url_base",
+		Type:    "qbittorrent",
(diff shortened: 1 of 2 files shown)


vavallee:
Merged via rebased branch after resolving conflict with #402 (legacyCredentialURLBase signature unification). All changes on main.

PR #452, opened 2026-05-04
[codex] Fix add-author search rejecting book-title matches

magrhino:
## Summary

Fixes the bug where some books could be added as authors from the Add Author modal.

OpenLibrary search can return work/book records alongside real author records. This change filters out book-title matches before they can be selected as authors, adds focused frontend coverage for the guard and modal behavior, adds the user-facing rejection copy, and records the fix in `CHANGELOG.md` under `[Unreleased]`.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed (not applicable; no env vars, config, or upgrade path changed)
- [x] Wiki pages updated if user-facing behaviour changed (not applicable; this is a narrow Add Author validation bug fix with no wiki workflow changes needed)

## Test plan

- [x] `cd web && npm test -- AddAuthorModal.test.tsx addAuthorTitleGuard.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,15 +9,11 @@ All notable changes to Bindery are documented here. Format loosely follows
 ### Fixed
 
 - **Manual series mutations require admin role** (#468) — Authenticated non-admin users can no longer create, update, monitor, delete, fill, or link series to Hardcover. Read-only series endpoints remain available to authenticated users.
-
-## [Unreleased]
-
-### Fixed
-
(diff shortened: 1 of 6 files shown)


vavallee (changes requested):
The CHANGELOG entry is backwards. It reads: "Add-author search no longer accepts book-title matches — Search results that come back from OpenLibrary as works/books are filtered out of the Add Author modal, preventing some books from being added as authors."

That describes the bug, not the fix. The actual change hides false-positive author results (where the author name is a known book title and the disambiguation points to the real author), while keeping real author results visible. Something like this would be accurate:

"Add-author search no longer hides valid author results when the query happens to match a book title — results whose name exactly matches a known book title and whose disambiguation matches the books real author are now placed behind a reveal button rather than silently dropped."

All Go tests pass and the logic itself looks correct. Please fix the changelog entry.

magrhino:
> The CHANGELOG entry is backwards. It reads: "Add-author search no longer accepts book-title matches — Search results that come back from OpenLibrary as works/books are filtered out of the Add Author modal, preventing some books from being added as authors."
> 
> That describes the bug, not the fix. The actual change hides false-positive author results (where the author name is a known book title and the disambiguation points to the real author), while keeping real author results visible. Something like this would be accurate:
> 
> "Add-author search no longer hides valid author results when the query happens to match a book title — results whose name exactly matches a known book title and whose disambiguation matches the books real author are now placed behind a reveal button rather than silently dropped."
> 
> All Go tests pass and the logic itself looks correct. Please fix the changelog entry.

Fixed the change log and the merge conflicts.

vavallee (dismissed):
CHANGELOG entry corrected in 27f72aa per earlier review. Approving.

Issue #398, opened 2026-04-25
NetworkError when attempting to fetch resource. attempting readarr.db import


Mikec78660:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

When pointing to the readarr.db from the UI upload readarr.db I get a message briefly that says it may take minutes. Then I get the error "NetworkError when attempting to fetch resource." There is nothing put into the logs. Just this error message in the UI. The db is not huge. It is 113mb.

Readarr info:
Version:    0.4.19.2811
Package Version:    0.4.19.2811-ls400 by [linuxserver.io](https://www.linuxserver.io/) 
.NET Core:    Yes (6.0.35)
Database:    Sqlite 3.48.0
Database Migration:    158
AppData Directory:    /config
Startup Directory:    /app/readarr/bin
Mode:    Console
Uptime :    112d 02:26:43

Bindery Version:  v1.2.6

### Expected behaviour

Database should be imported and parsed.

### Steps to reproduce

import readarr.db into bindery.

### Trace logs

```text
There are no log events
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v.1.2.6

vavallee:
Closing — the most likely cause is a **reverse proxy body size limit** in front of your Docker container. nginx defaults to `client_max_body_size 1m` (1 MB), which would silently reject a 113 MB file before the request ever reaches Bindery. No server log appears because the Go handler is never invoked.

**Fix for nginx:**
```nginx
client_max_body_size 200m;
```

**Fix for Caddy:**
```
request_body {
    max_size 200MB
}
```

**Workaround (bypasses HTTP entirely):**
```
bindery migrate readarr /path/to/readarr.db
```

PR #457 adds an entry log to the handler so future reports can immediately confirm whether the request reached the server.

PR #457, opened 2026-05-05
fix(migrate): add entry and error logging to readarr import handler

vavallee:
## Summary

When a readarr.db upload fails, there are currently no server-side log entries. This makes it impossible to distinguish a Go-level failure from a proxy/network rejection (the most common cause — nginx default `client_max_body_size` is 1 MB).

Adds:
- `slog.Info("readarr import started", "content_length", r.ContentLength)` at handler entry, so any future report can immediately confirm whether the request reached the Go server
- `slog.Warn("readarr upload rejected", ...)` on `acceptUpload` failure, surfacing multipart parse errors or size-limit rejections

Refs #398 (root cause is a reverse proxy body-size limit; documented in the issue close comment).


vavallee:
Superseded by #447 which makes the import fully async with a status endpoint — a more complete fix for the UX problem reported in #398.

PR #592, opened 2026-05-11
Surface timestamps in download queue UI

### LINKED ISSUE #543: Surface timestamps in Queue (and audit History timestamp UX)
The Queue page shows no timestamps. You can't tell whether an item was grabbed two minutes ago or two hours ago — only the live percentage / time-left from the downloader, which disappears the moment a job stalls or finishes downloading and enters the import phase.

The data already exists server-side. \`internal/models/download.go\` has:

\`\`\`go
AddedAt     time.Time   \`json:\"addedAt\"\`       // when the queue row was created
GrabbedAt   *time.Time  \`json:\"grabbedAt\"\`     // when the indexer release was grabbed
CompletedAt *time.Time  \`json:\"completedAt\"\`   // when the downloader finished
ImportedAt  *time.Time  \`json:\"importedAt\"\`    // when the importer moved the file
\`\`\`

These are serialised on every \`/api/v1/queue\` row (via \`QueueItem embeds Download\`). The TypeScript \`Download\` interface in \`web/src/api/client.ts\` (~line 917) just doesn't declare them, and \`QueuePage.tsx\` doesn't render them.

## What I'd ship

1. Add the four optional timestamp fields to the \`Download\` TS interface.
2. In \`QueuePage.tsx\`, render a single contextual timestamp per row using a small helper:
   - \`importedAt\` if set → \"imported 2 min ago\"
   - else \`completedAt\` if set → \"completed 5 min ago\"
   - else \`grabbedAt\` if set → \"grabbed 18 min ago\"
   - else \`addedAt\` → \"added 2 h ago\"
   - Use the same relative-time + absolute-tooltip pattern \`HistoryPage\` would benefit from (see #2 below).
3. Hover tooltip shows the absolute UTC timestamp.

## History page — secondary ask

\`HistoryPage\` already shows \`formatDate(event.createdAt)\` in both the desktop table column (line 168) and the mobile row footer (line 220). Worth a small UX pass while we're in there:
- The current format is presumably absolute and full-width — verify on mobile, where it eats horizontal space and forces wrapping.
- Consider relative-time-with-absolute-tooltip parity with the Queue work above so both pages feel consistent.

## Out of scope

- Sorting controls by any timestamp column. Default ordering is fine; sorting can land separately if there's demand.
- Backfilling the timestamps on legacy rows. \`AddedAt\` should already be populated for everything via the \`NOT NULL DEFAULT CURRENT_TIMESTAMP\` schema on the table; the optional ones are correctly null for old rows.

## Test plan

- [ ] Pipe through and verify a freshly added wanted item shows \"added Xs ago\" within seconds.
- [ ] Stall the downloader and confirm the relative time keeps ticking.
- [ ] Check timestamps survive a refresh and a rebuild (i.e. the TS type and JSON unmarshal align — easy to typo \`addedAt\` vs \`added_at\`).
- [ ] Mobile column doesn't blow out the row width.

kimjune01:
The download queue shows status but no timing information. Users can't tell when items were added, grabbed, or completed without checking logs.

This pipes the existing backend timestamp fields through to the frontend and displays the most recent relevant event as relative time with an absolute tooltip on hover.

Fixes #543

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -922,6 +922,10 @@ export interface Download {
   size: number
   protocol: string
   errorMessage: string
+  addedAt: string
+  grabbedAt?: string
+  completedAt?: string
+  importedAt?: string
 }
 
 export interface QueueItem extends Download {
--- a/web/src/pages/QueuePage.tsx
+++ b/web/src/pages/QueuePage.tsx
@@ -82,6 +82,37 @@ export default function QueuePage() {
     return (bytes / 1024).toFixed(0) + ' KB'
   }
 
+  const formatRelativeTime = (timestamp: string): string => {
+    const now = Date.now()
+    const then = new Date(timestamp).getTime()
+    const diffMs = now - then
+    const diffSec = Math.floor(diffMs / 1000)
+    const diffMin = Math.floor(diffSec / 60)
+    const diffHr = Math.floor(diffMin / 60)
+    const diffDay = Math.floor(diffHr / 24)
+
+    if (diffSec < 60) return `${diffSec}s ago`
+    if (diffMin < 60) return `${diffMin}m ago`
+    if (diffHr < 24) return `${diffHr}h ago`
+    return `${diffDay}d ago`
+  }
+
+  const getContextualTimestamp = (item: QueueItem): { label: string; absolute: string } | null => {
+    if (item.importedAt) {
+      return { label: formatRelativeTime(item.importedAt), absolute: new Date(item.importedAt).toUTCString() }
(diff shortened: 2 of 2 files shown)


vavallee (approved):
Small, well-scoped UI improvement. Backend Download model already exposes addedAt/grabbedAt/completedAt/importedAt as JSON. Relative-time fallback chain (imported → completed → grabbed → added) makes sense for showing the most recent meaningful event. Tooltip uses absolute UTC; minor nit but acceptable for v1.9.2. Merging.

PR #616, opened 2026-05-12
Fix/qbittorrent v5 referer auth

statte:
Fix qBittorrent v5.x login compatibility

Problem

qBittorrent v5.x introduced two breaking changes to the /api/v2/auth/login endpoint that caused the connection test to fail with an empty error body:

  1. CSRF protection ; the login endpoint now requires Origin and Referer headers matching the server URL. Requests without them are rejected silently.
  2. Success response changed ; v5.x returns 204 No Content on successful login instead of 200 Ok. The old check treated 204 as a failure.

  Changes

 - internal/downloader/qbittorrent/client.go 
   send Origin and Referer headers on the login request; accept 204 No Content as a successful login response alongside the
   existing 200 Ok. path
   include the HTTP status code in login failure errors to aid diagnosis.
 - internal/downloader/qbittorrent/client_test.go 
   add TestLogin_V5_NoContent and TestLogin_SendsCSRFHeaders to cover the new behaviour.

  Testing

  Verified against a live qBittorrent v5 instance behind a reverse proxy. All existing unit tests pass.

### DIFF
--- a/internal/downloader/qbittorrent/client.go
+++ b/internal/downloader/qbittorrent/client.go
@@ -72,6 +72,8 @@ func (c *Client) Login(ctx context.Context) error {
 		return fmt.Errorf("build login request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	// qBittorrent v5.x rejects login requests without a matching Referer header (CSRF protection).
+	req.Header.Set("Referer", c.baseURL)
 
 	resp, err := c.http.Do(req)
 	if err != nil {
--- a/internal/downloader/qbittorrent/client.go
+++ b/internal/downloader/qbittorrent/client.go
@@ -72,7 +72,8 @@ func (c *Client) Login(ctx context.Context) error {
 		return fmt.Errorf("build login request: %w", err)
 	}
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-	// qBittorrent v5.x rejects login requests without a matching Referer header (CSRF protection).
+	// qBittorrent v5.x requires both Origin and Referer for CSRF protection on the login endpoint.
+	req.Header.Set("Origin", c.baseURL)
 	req.Header.Set("Referer", c.baseURL)
 
 	resp, err := c.http.Do(req)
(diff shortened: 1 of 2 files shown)


vavallee:
Closing this — but I owe you a straight account of what happened, because this looks worse than it is and I want to make sure you don't read the wrong story into it.

Your PR opened at 20:06 UTC. About 6 hours later, an unrelated user (Zoso) hit roughly the same symptom on the bindery Discord. I worked from that thread, didn't know #616 existed, and shipped a band-aid fix in v1.9.4 #618 that only improved the error wording — it didn't actually fix the underlying v5.x auth break. Your PR did. Genuinely.

That should have been impossible. Earlier the same day I shipped a GitHub Actions workflow (#613) precisely to notify on every external PR via Discord, after closing another contributor's PR in the same kind of mess. The workflow merged at 18:12 UTC, **before** you opened #616 at 20:06 UTC. It should have fired. It didn't — there's literally no run record for #616 in the actions log, not even a "skipped" status. So the system that was supposed to surface your work failed silently.

That's a real bug, and I'm going to figure out why (separate issue incoming).

What I've done now:

- Your fix is shipped in **v1.9.5** (just tagged, deploying). The implementation and tests are yours verbatim, hybridised with v1.9.4's `AuthError` type for the cases your fix doesn't cover (legitimate non-2xx/non-204 rejections like wrong creds).
- The commit message and PR (#623) credit you as primary co-author and call out the v5.x CSRF + 204 root cause as the reason this PR even existed.
- Closing this PR rather than merging it directly because of the conflict with v1.9.4 — your branch was 6 hours older than my parallel work and rebasing 3 commits over an obsolete error path felt worse than re-landing the change with full attribution.

If you'd like a closer feedback loop, the bindery Discord (link in the README) is where this kind of thing surfaces fastest — happy to give you the actual context on anything else you're working on before you spend the time writing the patch.

Sorry for the duplicate work. Both fixes — yours for the protocol, mine for the error wording — are in main now. Thanks for the very clean PR and the live-instance verification.

vavallee:
Forgot to actually drop the link earlier — Discord invite: https://discord.gg/RpuYYRM9cZ. `#support` for the live-debug stuff, `#changelog` for release pings. Would be great to have you around.

PR #622, opened 2026-05-13
feat(library): surface unmatched files in Settings UI after scan

### LINKED ISSUE #535: Library scan: surface unmatched files in the UI (currently only in debug logs)
## Feature request

After a library scan the UI shows aggregate counts but offers no way to see *which* files were unmatched. Reported on Discord by a user with 732 unmatched files: *"It tells me there are 732 unmatched files, but I don't know anything else about them."*

## Current behavior

- The scanner writes a JSON summary to the `library.lastScan` setting (`internal/importer/scanner.go:1232-1244`):
  `{"ran_at":"...","files_found":N,"reconciled":N,"unmatched":N,"tag_read_failed":N}`
- `GET /api/v1/library/scan/status` returns that blob (`internal/api/library.go:42-56`).
- Per-file unmatched detail is emitted at **debug** level via slog (`internal/importer/scanner.go:1221`): `slog.Debug("library scan: unmatched file", "path", path, "parsedTitle", parsed.Title, "parsedAuthor", parsed.Author)`.
- Nothing in the UI or HTTP API exposes those paths.

## Suggested behavior

Persist per-file unmatched results so the UI can list them. Two shapes:

**A. New `library_scan_unmatched` table.** Truncated/replaced each scan run. Schema: `(scan_id, path, parsed_title, parsed_author, tag_read_failed, scanned_at)`. UI panel under Library (or Settings → Import) shows the current unmatched list with the parser's title/author guess, plus per-row actions like *"match to wanted book"* / *"ignore"*.

**B. Cheaper interim.** Extend the `library.lastScan` JSON blob to include up to the last N (e.g. 1000) unmatched paths inline. No schema change but bounded.

The Discord case (732 files) argues for (A): a truncated list won't tell the user whether all 732 share a single root-cause parser problem (e.g. authors-as-folder layout the parser doesn't understand) or are 732 genuinely unrelated files.

## Related — `component` log filter is dead code

While verifying that *"set log level to debug and grep the log store"* would actually work as a workaround, I found that the log store's `component` filter is non-functional. The DB log handler extracts a `"component"` attr from each slog record (`internal/db/log_handler.go:39-81`) and writes it to `logs.component`; the API exposes a `?component=` filter (`internal/api/logs.go:73`). But a repo-wide grep for `"component"` as a slog attr shows **zero callers ever set it**. Every row in the `logs` table has `component = ""`.

So the natural fallback advice — *"filter Settings → Logs by component=importer"* — silently returns nothing. Today users have to fall back to message-substring search for `unmatched file`. Worth a separate bug; mentioned here because it's the obvious workaround that doesn't actually work, which makes shipping the panel above more important rather than less.

## References

- `internal/importer/scanner.go:1078-1230` — scan loop & unmatched accounting
- `internal/importer/scanner.go:1221` — per-file debug log
- `internal/importer/scanner.go:1232-1244` — `writeScanResult` summary persistence
- `internal/api/library.go:42-56` — `ScanStatus` API
- `internal/db/log_handler.go:39-81` — log handler that the dead `component` filter relies on

kimjune01:
## Summary
- Extended library scan to collect up to 1000 unmatched file entries (path, parsed title, parsed author) and include them in the scan result JSON stored via `library.lastScan` setting
- Added `unmatched_files` field to the `libraryScanStatus` API response type
- Added a collapsible table in the Settings page that displays unmatched files after a scan, with a truncation indicator when the total exceeds 1000

Closes #535

## Test plan
- `go test ./internal/importer/...` passes
- `npm run typecheck` passes (zero TypeScript errors)
- Trigger a library scan with known unmatched files and verify the collapsible panel appears in Settings with correct path/title/author columns

### DIFF
--- a/internal/importer/scanner.go
+++ b/internal/importer/scanner.go
@@ -1188,7 +1188,7 @@ func (s *Scanner) ScanLibrary(ctx context.Context) {
 	slog.Info("library scan found files", "paths", []string{s.libraryDir, s.audiobookDir}, "count", len(foundFiles))
 
 	if len(foundFiles) == 0 {
-		s.writeScanResult(ctx, len(foundFiles), 0, 0, 0)
+		s.writeScanResult(ctx, len(foundFiles), 0, 0, 0, nil)
 		return
 	}
 
@@ -1230,6 +1230,7 @@ func (s *Scanner) ScanLibrary(ctx context.Context) {
 	// the correct earlier assignment.
 	reconciledBooks := make(map[int64]bool)
 
+	var unmatchedFiles []unmatchedFile
 	var reconciled, unmatched, tagReadFailed int
 	for _, path := range foundFiles {
 		// Skip files already tracked, or files inside a tracked directory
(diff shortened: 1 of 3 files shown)


vavallee:
ty!

vavallee (approved):
Reviewed and clean.

**What works**: 1000-entry cap is sensible, backwards-compatible field additions, UI handles the truncation case ("X of Y"), all CI green, separation between scanner / API / UI is clean.

**Non-blocking follow-ups worth filing** (don't hold this PR):

1. Rendering 1000 `<tr>` rows directly is fine on desktop but slow on mobile. Consider windowing/virtualizing or pagination if anyone reports stutter.

2. Payload bloat: the unmatched-files array can be ~200KB on busy libraries. `/library/scan/status` is now correspondingly heavier on every Settings → General mount. Splitting into a lazy-loaded `/library/scan/unmatched` endpoint would keep the main status response small.

Neither blocks the merge. Thanks for the clean PR + the test plan.

vavallee:
Merged, thanks @kimjune01. Two clean PRs from you in one day (#592 was the other) — really appreciate the quality.

Batching this for **v1.10.0** rather than cutting a v1.9.6 patch, since it's a feature add not a bug fix and the next release cycle is going to want a couple of larger changes shipped together.

If you'd like a faster feedback loop on PRs in flight, the bindery Discord is the best place — happy to give context before you sink time into something: https://discord.gg/RpuYYRM9cZ (`#support` for live debug, `#changelog` for release pings). Would be great to have you around.

PR #641, opened 2026-05-13
fix(qbittorrent): recover imports after path remaps

magrhino:
## Summary

- Harden qBittorrent torrent URL submission for Docker/container-networked setups where qBittorrent cannot resolve the indexer URL directly. Bindery still fetches the torrent payload internally, then uploads it to qBittorrent, while preserving qBittorrent category and save-path metadata.
- Add the small follow-up fixes around that internal torrent fetch/upload path: redirect-to-magnet handling, bounded torrent response reads, redirect body cleanup, explicit multipart field error handling, and hash detection from the submitted magnet when a redirect resolves to one.
- Add per-download-client path remaps persisted on download clients and exposed in Settings. Import scanning now applies the matching client remap before the global fallback, and qBittorrent grabs inverse-map Bindery's ebook/audiobook download directory into qBittorrent's save path.
- Add user-facing Settings guidance for qBittorrent path mapping direction: qBittorrent-reported path on the left, Bindery-readable path on the right, with qBittorrent-specific examples and reverse-save-path behavior called out.
- Add qBittorrent category path health checks. Settings now surfaces category/save-path mismatches, missing categories, empty category save paths, and the qBittorrent default save path when useful.
- Add recovery paths for failed queue work: failed grabs can retry the same download row without losing book/edition/indexer context, and `importFailed` queue items can reset import retries so users can fix path/category config or manually move completed torrent files, then import from the existing download.
- Update Queue, API docs, deployment docs, changelog, backend tests, frontend tests, and lint/security readiness fixes for the new behavior.

## Security and validation notes

- Torrent fetch URLs are validated with Bindery's outbound URL policy before the initial fetch and before each redirect target is followed.
- Non-HTTP(S) redirects are rejected unless the redirect is a magnet URI, which is submitted through qBittorrent's URL form path instead of being fetched as a network resource.
- Torrent responses are capped before upload to qBittorrent, and qBittorrent is not contacted when the fetch fails, returns an oversized body, or is rejected by URL validation.
- qBittorrent category JSON decoding accepts the observed save-path key variants (`savePath`, `save_path`, `download_path`) so health checks work across qBittorrent versions.

## Checklist

- [x] Tests added or updated

## Test plan

- [x] `gofmt -l .`
- [x] `go vet ./...`
- [x] `golangci-lint run --timeout=5m`
- [x] `GOCACHE=<tmp-cache> govulncheck ./...`
- [x] `go test ./internal/pathmap ./internal/importer ./internal/downloader ./internal/downloader/qbittorrent ./internal/api ./internal/db ./internal/scheduler`
- [x] `go test ./internal/downloader/qbittorrent ./internal/downloader/nzbget`
- [x] `cd web && npm test -- QueuePage SettingsPage`
- [x] `cd web && npm test`

Not run locally: `docker buildx build --platform linux/amd64,linux/arm64 -t bindery:local .`; `goreleaser release --snapshot --clean`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to Bindery are documented here. Format loosely follows
 
 ## [Unreleased]
 
+### Fixed
+
+- **qBittorrent imports recover from mismatched container paths without re-downloading** — Download clients now support per-client path remaps, qBittorrent grabs are sent with the expected category save path, and Settings surfaces qBittorrent category path health warnings. Queue items stuck in `importFailed` can also be retried after fixing storage/path settings.
+
 ## [v1.11.2] — 2026-05-17
 
 ### Fixed
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -686,6 +686,7 @@ func main() {
 		// Queue
 		r.Get("/queue", queueHandler.List)
 		r.Post("/queue/grab", queueHandler.Grab)
+		r.Post("/queue/{id}/retry-import", queueHandler.RetryImport)
 		r.Delete("/queue/{id}", queueHandler.Delete)
 		r.Get("/pending", pendingHandler.List)
 		r.Delete("/pending/{id}", pendingHandler.Delete)
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/vavallee/bindery/internal/calibre"
 	"github.com/vavallee/bindery/internal/config"
 	"github.com/vavallee/bindery/internal/db"
+	"github.com/vavallee/bindery/internal/downloader"
 	"github.com/vavallee/bindery/internal/hardcoverlistsyncer"
 	"github.com/vavallee/bindery/internal/importer"
 	"github.com/vavallee/bindery/internal/indexer"
@@ -358,6 +359,7 @@ func main() {
 	sched.WithAliases(authorAliasRepo)
 	sched.WithDelayProfiles(delayProfileRepo)
 	sched.WithPendingReleases(pendingReleaseRepo)
+	sched.WithStoragePaths(cfg.DownloadDir, cfg.AudiobookDownloadDir)
 	// Register the Calibre importer as the 24-hour sync job. The scheduler
 	// only fires the job when the syncer is non-nil, so no guard needed here.
 	sched.WithCalibreSyncer(calibreImporter)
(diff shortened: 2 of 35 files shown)


vavallee (commented):
(duplicate — see review below)

vavallee (commented):
(duplicate — see review below)

vavallee (commented):
Good work. All CI passes. Here's the full read.

**Blocking — needs rebase**

This branch was cut before #640 (the NZBGet-inspired qBit torrent-fetch fix) landed on main. `client.go` will conflict. Your implementation is a superset of #640 — redirect handling, URL validation, magnet-redirect support, proper `readLimited` enforcement, and explicit multipart field error handling. The conflict resolution should prefer your version wholesale. Please rebase on current main and force-push.

**Non-blocking observations**

*Torrent fetch timeout is 15 s.* `fetchTorrentContent` derives its timeout from `c.http` (15 s, same as qBittorrent API calls). #640 used 60 s explicitly. For indexers behind multiple slow redirect hops this might time out before the payload lands. Not a blocker, but worth keeping in mind if users report intermittent fetch failures after this ships.

*`defer resp.Body.Close()` inside the redirect loop* is not a bug (we `return` immediately on both paths after it), but placing the close inline next to each `return` instead of deferred would make the body lifetime explicit. Optional.

*`TargetDownloadDir` vs `ExpectedDownloadDirForClient`* — both resolve the expected download path for a client but take different inputs (explicit `mediaType` string vs `*models.DownloadClient` with media type inferred from category name). The split is justified by different call sites; a short comment explaining which is authoritative would help future maintainers.

**Verified correct**

- `PolicyLAN` in `httpsec.ValidateOutboundURL` allows RFC1918 and resolves hostnames before checking — Docker-internal hostnames like `prowlarr:9696` resolve to private IPs that pass, so the URL validation is not self-defeating. ✓
- `readLimited` reads `maxBytes+1` and rejects `> maxBytes` — correct tight-limit enforcement, not silent truncation. ✓
- Migration 040 is the correct next number. ✓
- `Category.UnmarshalJSON` covering `savePath` / `save_path` / `download_path` across qBittorrent versions is necessary and correct. ✓
- `importer.Remapper` and `importer.ParseRemap` kept as type/var aliases to `pathmap` — backwards-compatible, no callers break. ✓
- `ApplyInverse` longest-`to`-match logic is correct for generating qBittorrent save paths from Bindery's download dir. ✓
- `RetryFailed` guard (`existing.Status != models.StateFailed`) prevents double-grab. ✓
- `ResetImportRetry` restricts to `importFailed` state, 409 Conflict otherwise. ✓

magrhino:
@vavallee Quick clarification so I move the right branch forward: were you reviewing the current PR branch, `codex/qbittorrent-path-remap-clean`, or the older pre-rebase branch I restored here: [codex/investigate-bindery-download-failure-pre-rebase](https://github.com/magrhino/bindery/tree/codex/investigate-bindery-download-failure-pre-rebase)?

The current PR branch uses #640/current `main` as its base and was intended to be the mergeable path forward. The restored pre-rebase branch predates #640 and uses an earlier approach to the qBittorrent/download-client remap flow.

If you want the current PR branch, I’ll proceed with that. If you were referring to the restored pre-rebase branch instead, I can adjust from there.

vavallee:
This conflicts with main now that #640 has landed — needs a rebase. The conflict is entirely in `internal/downloader/qbittorrent/client.go`; the rest of the PR (path remap, health checks, `GetCategories`) has no conflicts. Your implementation is strictly better than #640 so prefer your version wholesale in the conflict. Happy to merge once rebased.

magrhino:
Rebased per your instructions and kept the PR’s hardened implementation over #640’s simpler version wholesale.

magrhino:
Rebased again to clear conflicts

PR #644, opened 2026-05-14
fix(logs): propagate runtime log level change to DB handler

statte:
## Problem

\`PUT /system/loglevel\` (the GUI log level toggle) only called \`ring.SetLevel()\`. The UI log viewer always reads from the **database** — the \`useDB = true\` branch fires unconditionally when a \`LogRepo\` is attached. The database-backed \`db.LogHandler\` had a plain \`slog.Level\` field set once at startup from \`BINDERY_LOG_LEVEL\` with no runtime mutation path.

Result: switching to \`debug\` in the GUI had zero visible effect on the log viewer. Only \`BINDERY_LOG_LEVEL=debug\` at process start worked.

## Root cause

- \`api.LogHandler.SetLevel\` → only \`h.ring.SetLevel(level)\` — \`db.LogHandler\` never updated
- \`db.LogHandler.level\` was a plain \`slog.Level\`, not safe to mutate concurrently and never exposed as settable
- \`api.LogHandler\` held no reference to \`db.LogHandler\`

## Fix

- Convert \`db.LogHandler.level\` to \`*atomic.Int64\` shared between the root handler and all \`WithAttrs\` children — a single \`SetLevel\` call propagates immediately to every handler in the tree.
- Add \`SetLevel\`/\`Level\` accessors to \`db.LogHandler\`.
- Add \`WithDBLogHandler(*db.LogHandler)\` to \`api.LogHandler\` so \`SetLevel\` updates both the ring and the DB handler in one call.
- Wire \`WithDBLogHandler(logDBHandler)\` in \`main.go\`.

Also adds per-tier query URL logging (API key redacted) and per-result title/author logging in \`SearchBookWithDebug\`, so switching to debug in the GUI immediately surfaces what each indexer was queried with and what it returned.

## Test plan

- [x] \`go test ./internal/db/... ./internal/api/...\` passes
- [x] Start with default log level; open log viewer — only INFO+ entries visible
- [x] Change log level to \`debug\` via GUI (Settings → System); trigger a book search
- [x] Refresh log viewer - \`debug\` entries appear (indexer query URLs, result counts, per-result titles/authors)
- [x] Restart; confirm level reverts to startup default

### DIFF
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -449,7 +449,7 @@ func main() {
 	bulkHandler := api.NewBulkHandler(authorRepo, bookRepo, blocklistRepo, sched)
 	backupHandler := api.NewBackupHandler(cfg.DBPath, cfg.DataDir)
 	rootFolderHandler := api.NewRootFolderHandler(rootFolderRepo)
-	logHandler := api.NewLogHandler(ring).WithLogRepo(logRepo)
+	logHandler := api.NewLogHandler(ring).WithLogRepo(logRepo).WithDBLogHandler(logDBHandler)
 	prowlarrHandler := api.NewProwlarrHandler(prowlarrRepo, indexerRepo).WithSettings(settingsRepo)
 	calibreHandler := api.NewCalibreHandler(settingsRepo)
 	grimmoryHandler := api.NewGrimmoryHandler(settingsRepo).WithVersion(version)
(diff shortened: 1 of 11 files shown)


vavallee:
Rebased and merged as #651 — dropped one duplicate lint commit that overlapped with #650. Thanks @statte!

PR #670, opened 2026-05-16
feat(abs): support selecting multiple libraries

### LINKED ISSUE #580: ABS: only one library_id supported — can't reconcile both Books and Audiobooks libraries
**Reported by:** @j-tt (fork: https://github.com/j-tt/bindery)
**Observed on:** v1.7.0 (commit 94a5103), NixOS

## Description

Only one ABS instance and one library ID can be configured. A typical ABS setup has two separate libraries — `Books` (ebooks, path `/media/books`) and `Audiobooks` (audiobooks, path `/media/audiobooks`). Bindery's ABS import reads from only one of them. If `abs.library_id` points to `Books`, audiobook items in the `Audiobooks` library are never reconciled; if it points to `Audiobooks`, ebook items are missed.

## File

`internal/abs/` — settings keys `abs.base_url`, `abs.library_id`

## Workaround

Point `abs.library_id` at whichever library is more important. Manually reconcile the other library's items if needed.

## Suggested fix

@j-tt has a working patch: https://github.com/j-tt/bindery/commit/702b738 — adds `abs.audiobooks_library_id` (`SettingABSAudiobooksLibraryID`). When set to a different library ID than `abs.library_id`, the importer enumerates both libraries in sequence within a single import run; each library gets its own `abs_import_runs` record. The new field is exposed in `GET`/`PUT /api/v1/abs/config` as `audiobooksLibraryId`.

---

Filed from @j-tt's [BUGS.md](https://github.com/j-tt/bindery/blob/main/BUGS.md). Credit to @j-tt for the report and fix.

magrhino:
## Summary

Purpose: Closes #580 by letting one Audiobookshelf source import from multiple selected book libraries, so setups with separate Books and Audiobooks libraries no longer have to choose one library ID. The change preserves the legacy `abs.library_id` compatibility field while adding ordered `libraryIds` config, sequential per-library import runs, and UI/docs coverage for selecting more than one ABS book library.

- Store normalized selected ABS library IDs and keep `libraryId` mirrored to the first selected ID for compatibility.
- Process selected libraries in saved order with one run, summary, checkpoint, rollback scope, and provenance set per library.
- Replace the ABS settings dropdown with a multi-library selector and update ABS import docs.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed (N/A - no env vars, deployment config, or upgrade path changed)
- [x] `CHANGELOG.md` updated if applicable (N/A - changelog entries are release-time only for feature work)
- [x] Wiki pages updated if user-facing behaviour changed (`docs/ABS-Import-Wiki.md` updated)

## Test plan

- [x] `go test ./internal/abs ./internal/api ./internal/importer ./cmd/bindery`
- [x] `cd web && npm test -- SettingsPage.test.tsx`
- [x] `cd web && npm run lint` (passes with existing warnings outside this change)
- [x] `cd web && npm test`
- [x] `gofmt -l` on changed Go files
- [x] `git diff --check`

Rebase validation on 2026-05-18:

- [x] `go test ./internal/importer ./internal/abs ./internal/api ./cmd/bindery`
- [x] `cd web && npm test -- SettingsPage.test.tsx`
- [x] `git diff --check`

### DIFF
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -264,15 +264,15 @@ func main() {
 
 	// Wire ABS post-import scan notification (Bug #10). The notifier reads ABS
 	// config from settings at call time so that config changes (URL, API key,
-	// library ID) take effect without restarting Bindery.
+	// library IDs) take effect without restarting Bindery.
 	importScanner.WithABSNotifier(
 		abs.NewScanNotifier(settingsRepo),
-		func() string {
+		func() []string {
 			cfg := api.LoadABSConfig(context.Background(), settingsRepo)
 			if !cfg.Enabled {
-				return ""
+				return nil
 			}
-			return cfg.LibraryID
+			return cfg.LibraryIDs
 		},
 	)
 
@@ -306,13 +306,14 @@ func main() {
 		})
 	storedABS := api.LoadABSConfig(ctxBoot, settingsRepo)
 	resumeCfg := abs.ImportConfig{
-		SourceID:  abs.DefaultSourceID,
-		BaseURL:   storedABS.BaseURL,
-		APIKey:    storedABS.APIKey,
-		LibraryID: storedABS.LibraryID,
-		PathRemap: storedABS.PathRemap,
-		Label:     storedABS.Label,
-		Enabled:   storedABS.Enabled,
+		SourceID:   abs.DefaultSourceID,
+		BaseURL:    storedABS.BaseURL,
+		APIKey:     storedABS.APIKey,
+		LibraryID:  storedABS.LibraryID,
+		LibraryIDs: storedABS.LibraryIDs,
+		PathRemap:  storedABS.PathRemap,
+		Label:      storedABS.Label,
+		Enabled:    storedABS.Enabled,
 	}
 	if resumed, err := absImporter.ResumeInterrupted(ctxBoot, resumeCfg); err != nil {
 		slog.Warn("abs interrupted import resume skipped", "error", err)
--- a/docs/ABS-Import-Wiki.md
+++ b/docs/ABS-Import-Wiki.md
@@ -6,7 +6,7 @@ This page is the user-facing companion to [`docs/abs_import.md`](./abs_import.md
 
 ## What It Does
 
-The current ABS importer supports one configured ABS source and one selected target book library at a time. From that source, Bindery can:
+The current ABS importer supports one configured ABS source and one or more selected target book libraries at a time. From that source, Bindery can:
 
 - test API-key access and discover the ABS book libraries visible to that key
 - import the ABS catalog metadata it can read for each visible item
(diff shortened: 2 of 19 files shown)


vavallee (commented):
Reviewed in depth — backward-compat, per-library checkpoint/rollback, and provenance all verified correct. Build, vet, Go tests, `gofmt`, and `npm run typecheck` all pass.

## Verified

- Legacy single `abs.library_id` still imports after upgrade (`abs.go:303`); the legacy field stays mirrored to `libraryIds[0]` on save.
- A failure in library 2 of 3 marks only that run failed and stops; library 1 stays `completed` (`importer.go:302`).
- Rollback is keyed by `runID` = one library; a series shared across libraries A and B is not deleted when rolling back B.
- Provenance and foreign IDs embed the library ID, so books from different libraries stay distinguishable when a library is later removed from the selection.

## Nits (non-blocking)

- **`web/src/pages/SettingsPage.tsx:2085`** — the import-run rows omit the library. A 3-library import shows 3 near-identical rows; a user cannot tell which run maps to which library when deciding what to roll back. This undercuts the feature's own headline scenario (separate Books vs Audiobooks libraries) — recommend rendering the library name/ID per run row.
- No multi-library rollback-isolation test — every `*Rollback*` test is single-library. Correctness holds by construction, but a test that imports A+B and rolls back only B, asserting A survives, would be the natural companion to the new multi-library run tests.
- `normalizeLibraryIDs` (`import_utils.go:152`) silently prepends `primary` when it is absent from `values`, expanding the import set. Unreachable through current API flows, but a latent footgun — assert or document the invariant.
- Four near-duplicate "trim + dedup strings" helpers now exist across three packages — ripe for consolidation.

LGTM once the run-row library label is addressed (or consciously deferred to a follow-up).

magrhino:
Thanks for the detailed review. I pushed two follow-up commits:

- `8e7b670` labels each recent ABS import run with the library name/ID in Settings, so multi-library runs are distinguishable when previewing or applying rollback. It also adds coverage for the run-row labels, the multi-library rollback isolation case where rolling back library B preserves library A, and the legacy-primary `normalizeLibraryIDs` behavior.
- `a0ddbcb` tightens rollback handling for shared local entities across ABS libraries: when another ABS provenance link outside the run still references the same local book/author/series or series membership, rollback removes only the rolled-back run’s ABS link and retains the shared local entity/membership. It adds coverage for rolling back one library when another library shares the same local book/author/series.

I only have a single-library ABS instance available locally, so I can’t validate the real multi-library ABS flow end-to-end. Could you check this against your multi-library setup and confirm the import-run labels make rollback selection clear, and that rolling back one library leaves the other library’s imported/shared books intact?

vavallee (approved):
Re-reviewed at the current revision and rebased onto `main` (clean — no conflicts with #765 despite both touching `client.ts`/`SettingsPage.tsx`).

Verified: closes #580; legacy single-library config remains backward-compatible (`LoadABSConfig` synthesizes `library_ids` from legacy `library_id`); per-library checkpoint/resume/rollback is correct, including cross-library provenance retention so rolling back one library does not delete entities another still references. Tests, `gofmt`, `go vet`, and frontend typecheck all pass.

Non-blocking nits only: several near-duplicate trim/dedupe string-slice helpers worth consolidating; book/author provenance keys on `item.LibraryID` while series keys on `cfg.LibraryID` (pre-existing, but more load-bearing now). Merging once CI is green.

PR #691, opened 2026-05-19
fix(metadata): harden Hardcover series and shelf mapping

magrhino:
This is PR 3 in a stacked series.

Stack:
1. vavallee/bindery#672 - Restore Hardcover API compatibility
2. vavallee/bindery#686 - Add Hardcover edition lookup
3. This PR - Harden Hardcover series and shelf mapping
4. vavallee/bindery#776 - Split Hardcover client responsibilities
5. vavallee/bindery#789 - Fix Hardcover import list sync reliability

This PR is based on `main` and should be reviewed after vavallee/bindery#686.

## Summary

Hardcover search results now expose series and shelf details differently than the older GraphQL paths Bindery originally targeted. This PR tightens the new metadata mapping so Bindery only persists stable numeric Hardcover series IDs and uses the current status ID for the built-in Did Not Finish shelf.

- Require numeric Hardcover series IDs before creating Bindery series references from search results.
- Preserve current search-result series metadata when a stable Hardcover series ID is available.
- Map Hardcover's Did Not Finish shelf to the current status ID.
- Expand regression coverage for series ID parsing and shelf status mapping.

## Checklist

- [x] Tests added or updated
- [x] N/A - `docs/DEPLOYMENT.md` unchanged; no env vars, config, or upgrade path changed
- [x] N/A - no wiki pages updated

## Test plan

- [x] `go test ./internal/metadata/hardcover`
- [x] `go test ./internal/metadata/...`
- [ ] `cd web && npm run build` - not run; Go metadata-only change

### DIFF
--- a/internal/metadata/hardcover/client.go
+++ b/internal/metadata/hardcover/client.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/vavallee/bindery/internal/isbnutil"
 	"github.com/vavallee/bindery/internal/metadata"
 	"github.com/vavallee/bindery/internal/models"
 	"github.com/vavallee/bindery/internal/useragent"
@@ -521,7 +522,9 @@ type HCList struct {
 	BooksCount int    `json:"booksCount"`
 }
 
-// hcBuiltinShelves are the four standard Hardcover reading-status shelves.
+// hcBuiltinShelves are the four Hardcover reading-status shelves Bindery
+// exposes for list sync. Hardcover status_id 4 is Paused, which is not exposed
+// as a synthetic list because existing list sync behavior only surfaces DNF.
 // They live in user_books (filtered by status_id), not in me.lists, so they
 // are injected as synthetic entries using negative IDs to avoid collision with
 // real list IDs.
@@ -542,7 +545,7 @@ func hcShelfStatusID(listID int) (int, bool) {
 	case -3:
 		return 3, true
 	case -4:
-		return 4, true
+		return 5, true
 	}
 	return 0, false
 }
@@ -816,9 +819,13 @@ type hcBook struct {
 	Rating        float64          `json:"rating"`
 	UsersCount    int              `json:"users_count"`
 	Genres        []string         `json:"genres"`
+	ISBNs         []string         `json:"isbns"`
+	HasAudiobook  bool             `json:"has_audiobook"`
+	HasEbook      bool             `json:"has_ebook"`
 	AudioSeconds  *int             `json:"audio_seconds"`
 	Contributions []hcContribution `json:"contributions"`
 	AuthorNames   []string         `json:"author_names"`
+	SeriesRefs    []models.SeriesRef
 }
 
 type hcEdition struct {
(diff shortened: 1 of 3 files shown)


vavallee (changes requested):
Reviewed in depth. One thing must be resolved before merge; the rest are nits.

## Blocker — `hcShelfStatusID` 4→5 is unverified

`client.go:548` changes the Did Not Finish status ID from 4 to 5 with a comment asserting "Hardcover status_id 4 is Paused". There is no source for that enum anywhere in the repo or this PR. If it is wrong, DNF sync silently pulls the wrong shelf or nothing at all — `getShelfBooks` returns `nil` for an unknown `status_id`, so no error ever surfaces.

Please cite the Hardcover schema/enum (a GraphQL introspection dump or a doc link). Same request for the hardcoded `status_id 1` at `client.go:455` — cross-check it against the same source.

## Scope vs summary

The PR is summarised as series + shelf hardening, but the diff also adds ISBN parsing (`searchISBNList`), genre parsing (`searchStringList`), and audiobook/ebook media-type detection. Either fold those into the description or split them out — a reviewer trusting the summary would miss the `MediaType` behaviour change entirely.

## Nits (non-blocking)

- `MediaType` is now set for search results, but `toBook`'s other callers (`GetListBooks`, `getShelfBooks`, `GetSeriesCatalog`) do not request `has_audiobook`/`has_ebook`, so those paths always produce `MediaType: ""`. Inconsistent — worth a comment or an explicit default.
- `searchSeriesRefs` silently drops a title-only series when it lacks a numeric ID (`client.go:1188`). That is the intended hardening, but add a `slog.Debug` so the drop is diagnosable in the field.
- `searchStringList`'s default case routes unknown types through `fmt.Sprint`, which would stringify a stray map into a "genre". Add a type guard.

Tests pass, `gofmt` clean.

magrhino:
Thanks, this is fair. I added named Hardcover reading-status constants and [cited](https://github.com/hardcoverapp/hardcover-docs/blob/31aaa75774ec560312222e5834322c71b79dbb5b/src/content/docs/api/GraphQL/Schemas/UserBooks.mdx#L12-L21) Hardcover's official UserBooks docs source for the enum values, including `1 = Want to Read`, `4 = Paused`, `5 = Did Not Finish`, and `6 = Ignored`. I also moved the wishlist query to use the named `Want to Read` status constant instead of embedding `status_id: 1` directly.

You're also right about the summary: sorry I left the ISBN parsing work out of it. I should have called that out alongside the series and shelf changes so the review scope was clear.

## Nits Summary
I also went ahead and patched your nits so this ships cleanly.
- Added slog.Debug when searchSeriesRefs drops a title-only Hardcover series ref without a numeric ID in client.go (line 1228).
- Tightened searchStringList to accept only strings, JSON numbers, and explicit numeric scalar types, dropping maps/bools/structs instead of stringifying them.
- Added a focused SearchBooks test for dropping non-scalar genre values in client_test.go (line 471).
- Mapped Hardcover’s supported default audio/ebook edition IDs into Bindery media types so list imports, author/book lookups, and series gap-fill no longer default all non-search Hardcover books to ebooks.

vavallee (approved):
Re-reviewed — the blocker from my earlier review is resolved, so this supersedes that `CHANGES_REQUESTED`.

The "Did Not Finish" status mapping is now verified: named status constants were added citing Hardcover's official `UserBooks.mdx` schema docs (pinned to a commit), which confirms `4 = Paused`, `5 = Did Not Finish`. The earlier nits are all addressed — `MediaType` set on non-search `toBook` callers, `slog.Debug` on dropped title-only series, and `searchStringList` returning defaults rather than `fmt.Sprint`. Tests, `gofmt`, and `go vet` pass.

Non-blocking: the PR description still doesn't mention this also ships ISBN/genre/media-type parsing, and `searchStringList` has some uncovered unreachable numeric-type branches. Neither blocks merge.

PR #764, opened 2026-05-21
Fix: Support for older (AuthorName) Readarr DB schema

hucknz:
## Summary

Older installations of Readarr use a different schema in the Author database. This fix implements handling of that schema and falls back to the original column name for newer installs. 

## Checklist

- [x] Tests added or updated

## Test plan

### DIFF
--- a/internal/migrate/readarr.go
+++ b/internal/migrate/readarr.go
@@ -46,11 +46,13 @@ func ImportReadarr(
 	if dbPath == "" {
 		return nil, errors.New("readarr db path is required")
 	}
+
 	src, err := sql.Open("sqlite", "file:"+dbPath+"?mode=ro&immutable=1")
 	if err != nil {
 		return nil, fmt.Errorf("open readarr db: %w", err)
 	}
 	defer src.Close()
+
 	if err := src.PingContext(ctx); err != nil {
 		return nil, fmt.Errorf("ping readarr db: %w", err)
 	}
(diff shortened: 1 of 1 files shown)


vavallee (changes requested):
Thanks for this — the core fix is genuinely useful, but the PR needs to be tightened before it can merge.

**The good part:** `queryReadarrAuthors` with the `AuthorMetadata`-join fallback is exactly right for older Readarr DBs. Keep that.

**Blocking — CI is red.** `validate (Go)` fails on `TestImportReadarr_HappyPath` (`readarr_test.go:238`). The cause isn't the schema fix — it's an **unrelated download-client change** bundled into this PR:

- The `models.DownloadClient` literal drops `Username` and folds credentials into `APIKey` (`apiKey := s.APIKey; if apiKey == "" { apiKey = s.Password }`). The failing test shows the imported qBittorrent client coming out with `Username:` empty.
- That's a functional regression: Bindery's qBittorrent client authenticates with **username + password** (`qbittorrent.New(host, port, username, password, ...)`). An imported qBit client with no username can't log in. This is out of scope for a PR titled "support for older Readarr DB schema."

**Also unflagged in the description:**
- `onSearchOnAdd` is now gated on `monitored` and its explanatory comment was deleted.
- `parseSettings` silently drops its `slog.Warn` on parse failure (`_ = json.Unmarshal(...)`).

**Ask:** scope this PR down to just the `queryReadarrAuthors` schema fallback — that's the fix, it's good, and it'll go green on its own. If the download-client credential change is intentional, please open it as a separate PR with its own rationale and an updated `readarr_test.go` so the qBittorrent auth path stays covered. The `onSearchOnAdd` and `parseSettings` changes should either be reverted or called out explicitly with reasoning.

hucknz:
Sorry, those changes shouldn't have been in this branch. I'll create a new PR with just the Readarr fix.

PR #765, opened 2026-05-21
Fix: CSRF failing for Readarr import

hucknz:
## Summary

The Readarr import flow fails due to missing X-CSRF-Token header. 

## Checklist

- [x] Tests added or updated

## Test plan

### DIFF
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -68,6 +68,32 @@ async function request<T>(path: string, options?: RequestInit): Promise<T> {
   return res.json()
 }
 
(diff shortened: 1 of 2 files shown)


vavallee (approved):
Clean fix. The migrate upload was doing a raw `fetch` that bypassed the `request()` wrapper, so it sent neither `X-CSRF-Token` nor `X-Requested-With` and failed CSRF once it is enforced. The new `uploadFile` helper injects both and correctly omits `Content-Type` so the browser sets the multipart boundary; it also fixes a latent sub-path-deploy bug (the old hard-coded `/api/v1`). CI green. Merging.

PR #787, opened 2026-05-21
fix: support both old and new Readarr database schemas in import

hucknz:
## Summary

Adds support for different Readarr database schemas. Older schemas store display names in AuthorMetadata table, newer schemas have Name column directly on Authors table.

- Add queryReadarrAuthors() helper to handle schema differences
- Fallback to simpler query if JOIN fails on newer schema
- Both queries return consistent (name, monitored) columns

## Checklist

- [x] Tests added or updated

## Test plan

### DIFF
--- a/internal/migrate/readarr.go
+++ b/internal/migrate/readarr.go
@@ -77,12 +77,41 @@ func ImportReadarr(
 	return res, nil
 }
 
+// queryReadarrAuthors handles both old and new Readarr schemas:
+// - Newer schemas have Authors.Name directly
+// - Older schemas store display names in AuthorMetadata, joined via AuthorMetadataId
+//
+// Returns rows with two string columns (name, monitored-as-int) in both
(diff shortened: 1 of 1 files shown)


vavallee:
Closing in favour of #784, which fixes the same root cause and has now merged.

The two PRs disagree on a factual point: this PR assumes newer Readarr schemas expose `Authors.Name` directly, with the `AuthorMetadata` join as a fallback for older ones. That isn't accurate — Readarr inherited Lidarr's metadata-table pattern (`Authors.AuthorMetadataId` → `AuthorMetadata.Name`) from day one and has never had an `Authors.Name` column. The `SELECT Name FROM Authors` fallback here is dead code, and the PR doesn't update the migration tests.

#784 takes the simpler correct path — always join `AuthorMetadata` — and updates the test fixtures to the real schema. Thanks for the fix though; you and nathang21 spotted the same bug independently.

PR #790, opened 2026-05-22
feat(telemetry): access-log middleware on the telemetry-server

vavallee:
## Summary

The telemetry-server only logged pings (and errors) — there was no per-request access log, and no client IP in the logs at all. Adds a `logRequests` middleware: one `slog` line per request with method, path, status, the **real client IP**, and duration.

## Implementation notes

- `logRequests` wraps the chain outermost — `logRequests(secureHeaders(canonicalHost, mux))` — so even requests `secureHeaders` rejects are logged.
- Client IP via the existing `realIP` helper (`X-Real-Ip` / `X-Forwarded-For` aware) — shows the external address, not Traefik's pod IP.
- `/health` is skipped so the 10s readiness probe doesn't flood the log.
- `statusRecorder` captures the response status (defaults to 200).

## Checklist

- [x] Tests added or updated — `TestLogRequestsPassesThrough`
- [x] Doc-update gate cleared — no env vars / flags / auth / config surface changed
- [x] Wiki — n/a, internal

## Test plan

- [x] `go test -race ./cmd/telemetry-server/`
- [x] `go vet` + `go build` + `gofmt -l ./cmd/telemetry-server/` — clean

### DIFF
--- a/cmd/telemetry-server/main.go
+++ b/cmd/telemetry-server/main.go
@@ -173,7 +173,7 @@ func main() {
 
 	srv := &http.Server{
 		Addr:         addr,
-		Handler:      secureHeaders(canonicalHost, mux),
+		Handler:      logRequests(secureHeaders(canonicalHost, mux)),
 		ReadTimeout:  10 * time.Second,
 		WriteTimeout: 30 * time.Second,
 		IdleTimeout:  120 * time.Second,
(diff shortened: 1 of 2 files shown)


vavallee:
Reverting — telemetry is intentionally anonymous; not logging client IPs.

PR #795, opened 2026-05-22
feat(importer): manual file import via path lookup (#766)

### LINKED ISSUE #766: feat: manual file import - path lookup + import via existing scanner pipeline
## Context

Related to #532 (this is the minimal single-file slice; bulk directory scan with preview table remains the full #532 scope). Distinct from #537 (blackhole/drop-folder mode, which this is not).

## Problem

Users with pre-existing book files on disk, from previous software, manual downloads, or side-loads, have no way to import a specific file into Bindery through the UI. The current workarounds are:
- Drop into `BINDERY_LIBRARY_DIR` and run a library scan (works, but slow and silent with no per-file feedback)
- `importMode=external` (requires a grab first)

## Proposed design

Two endpoints, no new DB tables, no new import engine. The existing `tryImportInternal()` in `scanner.go` already handles everything once it has a `*models.Download` row.

### Step 1 - Lookup (read-only)
```
GET /api/v1/queue/manual-import/lookup?path=/path/to/file.epub
```
1. `ParseFilename(path)` extracts title, author, ASIN from the filename
2. Search local catalogue in priority order:
   - ASIN exact match
   - `FindByAuthorAndTitle()` exact match
   - Jaro-Winkler >= 0.85 fuzzy match over `books.List()`
3. Returns one of:
   - `{"match": "confident", "book": {...}}` - single high-confidence match
   - `{"match": "ambiguous", "candidates": [...]}` - multiple candidates, user picks
   - `{"match": "none"}` - no local match; client falls back to the existing metadata search flow so the user can add the book first
4. The response also includes `"detectedFormat": "ebook"|"audiobook"` so the UI can pre-populate the format selector and the user can override it before confirming

### Step 2 - Import (write)
```
POST /api/v1/queue/manual-import
{
  "path": "/path/to/file.epub",
  "bookId": 123,
  "format": "ebook"
}
```
- `format` is optional. When omitted, `detectDownloadFormat()` decides from file extensions as usual. When provided, it is validated against `["ebook", "audiobook"]` and used as an explicit override, which matters for dual-format books (`media_type='both'`) or any case where the file extensions are ambiguous.
- Supporting the override requires a small change to `tryImportInternal()`: add an optional format hint parameter that bypasses `detectDownloadFormat()` when set. All existing callers pass `""` and behaviour is unchanged.

The rest of the flow:
1. Validate path exists and `IsBookFile()` passes
2. Create a `Download` record: `status=completed`, `download_client_id=NULL`, `guid="manual-<uuid>"`, `book_id=bookId`
3. Call `tryImportInternal()` directly, which handles move/copy/hardlink, `SetFormatFilePath()`, history event, Calibre/ABS push unchanged
4. Return 202 with the new download ID for progress polling via existing `GET /api/v1/queue`

Folders (audiobooks) work the same way: `path` points to the folder, and `tryImportInternal()` already handles directory sources.

## What is reused unchanged
- `ParseFilename()` - filename metadata extraction
- `IsBookFile()` - extension whitelist
- `tryImportInternal()` - entire import engine (move/copy/hardlink, renaming, Calibre, ABS notify, history), with one optional parameter addition
- `JaroWinkler` + `normalizeTitle` - fuzzy matching already in scanner
- All existing download state machine and queue UI

## What is new
- `GET /api/v1/queue/manual-import/lookup` handler (~60 lines)
- `POST /api/v1/queue/manual-import` handler (~40 lines)
- Optional format hint parameter on `tryImportInternal()` (~5 lines)
- UI: file path input + format selector + lookup result display + confirm button (placement open to guidance, see below)
- Route registrations in `main.go`

## What this is not
- Not a watched folder / blackhole mode (see #537)
- Not the full bulk directory scan described in #532 (that remains open)
- No new DB schema

## Roadmap toward full #532

This PR is the first step in a staged implementation of #532. The stages are kept separate for concrete reasons, not just to limit scope.

**Stage 1 (this PR): Single file/folder import with filename-based matching**

Gets the core plumbing in place: lookup endpoint, synthetic `Download` record creation, and the `tryImportInternal()` call path. Useful on its own for users who want to import individual files without dropping them into the library dir and waiting for a scan.

**Stage 2: Embedded metadata extraction (EPUB Dublin Core, M4B/MP3 ID3 tags)**

Kept separate from Stage 1 because filename-only matching is already useful for single files where the user can correct a wrong guess before confirming. However, embedded metadata is a prerequisite for reliable batch matching - running filename-only lookup against a directory of hundreds of files produces too many low-confidence and wrong matches to be useful. Stage 2 needs to land before Stage 3 can be trustworthy.

**Stage 3: Batch/directory scan**

Trivial once Stage 2 exists - the lookup endpoint from Stage 1 becomes the per-file primitive that the batch walker calls in a loop. Kept separate because building batch UI against unreliable filename-only matching would create a bad experience and likely require rework once Stage 2 lands.

**Stage 4: Preview table with per-row approve/reject**

The full #532 UX. Kept separate from Stage 3 because it requires a decision on how to store pending scan results between scan and execute - either a new DB table and migration (following the ABS importer pattern) or an in-memory approach. That is an architectural call worth making deliberately rather than rushing to close the issue. Stage 3 can be validated first to make sure matching quality is acceptable before investing in the full preview UI.

**Stage 5: Rollback support for batch imports**

Optional follow-on using the same before/after snapshot pattern as the ABS importer. Not a blocker for any earlier stage.

## Open questions for maintainer
1. **UI placement**: does this belong on the Queue page (alongside grab) or as a new Settings section?
2. **Source file handling after move-mode import**: `tryImportInternal()` takes a `cleanupFunc` to remove the source from the download client after import. For manual imports with move mode, should the source file be deleted after import, or left in place for the user to manage?
3. **`download_client_id` nullability**: is there already a NOT NULL constraint on that column that would need relaxing?

Happy to implement once there is buy-in on the shape.

statte:
## Summary

- Adds a **Manual Import** section to Settings → Import: paste a local file/folder path, click Lookup, confirm the match, and import
- Backend parses the filename (title, author, ASIN, series) and fuzzy-matches against wanted books using the existing `tryImportInternal` engine
- Fixes inline `Series Book N - Title` filename parsing so `Convergence_Book_2_-_Dragonslayer.m4b` correctly extracts title "Dragonslayer" instead of "Convergence Book 2"
- Format auto-detected from extension/directory; user can override (ebook / audiobook)

## What's included

- `internal/importer/lookup.go` — `Lookup()` on Scanner, fuzzy author matching (exact, comma-inversion, JaroWinkler ≥ 0.80), format detection
- `internal/importer/scanner.go` — `formatHint` param on `tryImportInternal`; public `ImportFromPath()` wrapper
- `internal/importer/parser.go` — `seriesBookInlineRe` pattern handles `Series Book/Vol N - Title` filenames
- `internal/api/manual_import.go` — `GET /queue/manual-import/lookup` and `POST /queue/manual-import` behind a `manualImportScanner` interface
- `cmd/bindery/main.go` — route wiring
- `web/` — `ManualImportSection` component in `ImportTab`, API client methods, i18n keys

## Test plan

- [x] `go test ./internal/importer/... ./internal/api/...` passes
- [x] Settings → Import → Manual Import section visible
- [x] Lookup with a valid path returns a confident match
- [x] Lookup with an unknown file returns "No match found"
- [x] Format override (ebook / audiobook) respected on import
- [x] `Series_Book_N_-_Title.m4b` filenames now match correctly

Closes #766


statte:
@vavallee As this is a quite substantial update, I would like input from your side. If it is not the correct way to go about the implementation, please let me know.

vavallee (changes requested):
Thanks for the substantial work here, statte. The Lookup helper, parser fix (\`seriesBookInlineRe\`), and \`formatHint\` plumbing are clean and well-tested.

Two security-shaped concerns block merge as-is. Both are fixable with small edits.

## 1. Routes need to sit behind \`auth.RequireAdmin\`

The \`#nosec G304\` comments on \`internal/api/manual_import.go:47,94\` claim "endpoint is admin-only", but the route registrations at \`cmd/bindery/main.go:723-724\` are inside the user-level \`/api/v1\` group, not the \`r.Use(auth.RequireAdmin)\` sub-group. So any authenticated user (role \`user\`) can call these endpoints today.

The manual-import flow has real authority: \`POST /queue/manual-import\` ultimately calls \`tryImportInternal\`, which copies/hardlinks/moves the supplied path into the configured library directory. Combined with the existing \`?deleteFiles=true\` on \`DELETE /book/{id}\`, a non-admin can read or move arbitrary server-filesystem paths into the library and back out. That's a multi-user privilege escalation against the role model v1.15.1 hardened.

Fix: move both routes into the \`r.Group(func(r chi.Router) { r.Use(auth.RequireAdmin); ... })\` block in \`cmd/bindery/main.go\`. The nosec comments are then accurate.

## 2. Path needs to be constrained to a configured root

Even with admin gating, accepting any absolute path on the server filesystem is broader than the feature description implies (importing files the admin already has under their managed library). An admin pasting \`/etc/...\` or \`/proc/...\` won't yield useful results today but will exercise filesystem-read primitives that don't need to exist.

Fix: before any \`os.Stat\` / \`Lookup\` / \`ImportFromPath\` call, reject paths that don't resolve under \`s.libraryDir\` and the configured \`root_folders\` (the same set the scanner walks). Symlink resolution via \`filepath.EvalSymlinks\` before the containment check defeats symlink-traversal escapes.

This is the same defense-in-depth the existing \`isAllowedPath\` check in \`internal/api/files.go\` performs for the file-download endpoint, and it's tracked for the deletion path under #865.

## What I'd recommend

Two small commits on this branch:

1. Move the two route registrations into the \`RequireAdmin\` group.
2. Add a \`rejectIfOutsideAllowedRoots(path)\` helper at the top of both Lookup and Import that takes the same root list the scanner already builds. The helper is also a good target for unit tests (path under root: accept; path outside: 403; symlink escaping root: 403).

Tests + parser fix are already great; this is just the access-control surface. Once those two changes land I'll happily admin-merge. Feature itself is exactly what users have been asking for in #766.

statte:
Hey,

Ok understood. I will look into on how to implement.

Thanks.

statte:
**Update — addresses review feedback:**

**1. Admin-only routes** — both `/queue/manual-import/lookup` and `POST /queue/manual-import` are now inside the `r.Use(auth.RequireAdmin)` group in `cmd/bindery/main.go`, not the user-level `/api/v1` group.

**2. Path containment** — added `isAllowedPath` on `ManualImportHandler` that calls `filepath.EvalSymlinks` before checking containment against `cfg.LibraryDir`, `cfg.AudiobookDir`, and the `root_folders` table. Both handlers return 403 for paths outside those roots. Mirrors the existing `isAllowedPath` defence in `internal/api/files.go`.

**Tests added** (`internal/api/manual_import_test.go`):
- Path outside roots → 403 (ebook file / audiobook file / audiobook directory, for both Lookup and Import)
- Path inside ebook root and audiobook root → 200/202
- Symlink inside root pointing outside → 403

Commit: 988b8d3

statte:
**CI timeout note:**

The `internal/api` test package is currently timing out at 10 min under `-race` on CI — but this is a pre-existing issue on upstream `main` itself (run [26672242131](https://github.com/vavallee/bindery/actions/runs/26672242131) on the merge commit for v1.15.3 shows the same `panic: test timed out after 10m0s`). It is not introduced by this PR.

All three of the concerns raised in review are addressed:
1. Routes behind `RequireAdmin` ✓
2. Path containment with `EvalSymlinks` ✓  
3. Tests updated for `tryImportInternal` signature change introduced by this branch ✓

Happy to rebase or address anything else once the CI infrastructure issue is resolved upstream.

statte:
@vavallee it seems i can't do much about those security warnings from my side.

Issue #769, opened 2026-05-21
qBittorrent re-grab fails with HTTP 409 when the torrent is already present


vavallee:
## Current behaviour

Re-grabbing a book whose torrent qBittorrent **already holds** fails hard:

```
failed to send to downloader: add torrent HTTP 409: Conflict
```

qBittorrent returns `409 Conflict` from `POST /api/v2/torrents/add` when asked to add a torrent it already has. Bindery treats that as a fatal send failure.

## Root cause

`internal/downloader/qbittorrent/client.go` `AddTorrent` checks the response status with `if resp.StatusCode != http.StatusOK { return "", fmt.Errorf("add torrent HTTP %d: %s", ...) }`. Any non-200 — including a 409 duplicate rejection — is a hard error, short-circuiting before the normal hash-resolution logic. There is no handling for the already-present case.

## Why it's wrong

A 409 here means qBittorrent already has the torrent, so the content is effectively **available** — re-adding it is not a real failure. This is the same shape of bug as #690 / #693, where qBittorrent 5.x's *success* response was being read as a failure.

## Repro (Discord support report)

1. Grab a book; the torrent lands in qBittorrent.
2. With that torrent still present in qBittorrent, re-add / re-grab the same book.
3. `add torrent HTTP 409: Conflict`.

## Proposed fix direction

In `AddTorrent`, treat a qBittorrent duplicate rejection (409) as a non-error: instead of bailing, fall through to recover the already-present torrent's infohash — the before/after hash-set diff `AddTorrent` already does for the normal path, or `infoHashFromMagnet`. Return that hash as success so Bindery tracks the existing torrent and proceeds to import, rather than reporting "failed to send to downloader".

Complementary to #766 (manual file import) but distinct — this is purely the duplicate-add path in the qBittorrent client.

PR #798, opened 2026-05-23
fix(importer): mark re-grabbed already-imported torrent as StateImported (#769)

statte:
## Problem

When a torrent is re-grabbed and qBittorrent already holds it (409 duplicate-add), PR #783 correctly recovers the hash and creates a `Download` record in `StateGrabbed`. Two gaps remained that prevented the import from completing:

**Gap 1 — `content_path` absent, files already in library (move mode)**

`checkQbittorrentDownloads` calls `resolveQbitContentPath` which returns `false` when the download files were moved to the library by a prior import. The download loops in `StateGrabbed` forever — `os.Stat` never finds the path and there's no exit condition.

**Gap 2 — Empty download directory (move mode), all clients**

`tryImportInternal` walks the download path, finds no book files, and emits `StateImportFailed`. After three retries `blockStaleImportFailures` transitions it to `StateImportBlocked` — even though the book is already in the library.

**Gap 3 — `grabbed → completed` transition missing from state machine**

`checkQbittorrentDownloads` calls `updateDownloadStatus(StateCompleted)` before `tryImportInternal`, but `grabbed → completed` was not in `validTransitions`. The transition was silently rejected, leaving the `Download` record in `StateGrabbed` even after a successful file copy — for any torrent that was already at 100% when first grabbed.

## Fix

- **`isBookAlreadyImported(ctx, dl)`** — new helper that checks `book_files` + `os.Stat` for both ebook and audiobook formats. Returns `true` only when the file is actually on disk in the library.

- **`checkQbittorrentDownloads`** — when `resolveQbitContentPath` returns `false` and the book is confirmed in the library, walk the state machine (`grabbed/downloading → completed → importPending → importing → imported`) instead of retrying indefinitely. Applied to both the fresh-grab branch and the `StateImportFailed` retry branch.

- **`tryImportInternal`** — when `bookFiles` is empty and the book is already in the library, walk `importPending → importing → imported` instead of emitting `StateImportFailed`. Covers all download clients.

- **`download_state.go`** — add `grabbed → completed` as a valid transition.

## Tests

Two new regression tests in `scanner_qbittorrent_test.go`:
- `TestCheckQbittorrentDownloads_ContentPathGoneAlreadyImported` — torrent seeding with `content_path` absent, book already in library → `StateImported`
- `TestTryImportInternal_EmptyPathAlreadyImported` — empty download directory, book already in library → `StateImported` (not `StateImportFailed`)

### DIFF
--- a/internal/importer/scanner.go
+++ b/internal/importer/scanner.go
@@ -892,6 +892,22 @@ func (s *Scanner) checkQbittorrentDownloads(ctx context.Context, client *models.
 		if isComplete && (dl.Status == models.StateDownloading || dl.Status == models.StateGrabbed) {
 			rawPath, ok := resolveQbitContentPath(torrent)
 			if !ok {
+				// content_path is absent or points to a path that no longer exists.
+				// This can happen when files were moved to the library by a prior
+				// Bindery import (move mode) and the torrent is re-grabbed via a
+				// 409 duplicate-add (#769). If the book is already in the library,
+				// close out this download immediately rather than looping forever.
+				if s.isBookAlreadyImported(ctx, &dl) {
+					slog.Info("qbittorrent: content path gone but book already in library — marking as imported",
+						"title", dl.Title)
+					// Walk the state machine: grabbed/downloading → completed →
+					// importPending → importing → imported.
+					s.updateDownloadStatus(ctx, dl.ID, models.StateCompleted)
+					s.updateDownloadStatus(ctx, dl.ID, models.StateImportPending)
+					s.updateDownloadStatus(ctx, dl.ID, models.StateImporting)
+					s.updateDownloadStatus(ctx, dl.ID, models.StateImported)
+					continue
+				}
 				// Path doesn't exist on disk yet (qBittorrent may sanitise characters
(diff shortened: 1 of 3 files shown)


vavallee:
Thanks for the thorough root-cause writeup — three gaps land in one place.

Two notes:

**Blocker — lint is red.** `gofmt` on `internal/models/download_state.go`:

```
internal/models/download_state.go:36:1: File is not properly formatted (gofmt)
```

`StateGrabbed` got longer so the column alignment shifted; `gofmt -w` and force-push.

**Edge case worth flagging in `isBookAlreadyImported`.** It returns true if *any* format (ebook OR audiobook) is present for the book. If a user has the audiobook in library and grabs the ebook (or vice versa), and `content_path` is absent for whatever reason, the fresh ebook download gets stamped `StateImported` without anything moving. Low probability for the #769 reseed case, but the guard is broader than the bug. If the `Download` can carry the expected format/category at this point, scope it; otherwise note the tradeoff in the comment so future readers know.

Otherwise the state-machine walk, the `len(scopes)==0` symmetry between the two branches, and the test coverage all look right.

PR #822, opened 2026-05-25
feat(metadata): hydrate hardcover editions from known hardcover ids

magrhino:
## Summary

Hardcover metadata now exposes useful edition fields, but Bindery was not persisting them when an existing workflow already had a confident Hardcover book identity. This PR hydrates edition rows and promotes audiobook ASINs only from known Hardcover IDs already present during create, rebind, sync, series, and recommendation flows; it does not add matching logic, backfill existing books, or change best-effort add-book matches that lack a `HardcoverForeignID`.

- Adds a shared `bookhydrate` helper that fetches Hardcover editions, upserts edition metadata without overwriting existing non-empty fields, skips edition IDs already attached to another book, and promotes accepted audiobook ASINs before optional Audnex enrichment.
- Adds provider-explicit edition fetching so matched Hardcover IDs can hydrate even when the stored primary book identity belongs to another provider.
- Wires best-effort hydration into rebind/map-metadata, author sync, Hardcover list sync, series fill/catalog creation, recommendations, and direct add-book when the created book already carries a Hardcover identity.
- Adds focused coverage for the hydration helper, metadata edition upserts, matched author sync, direct add, rebind/map-metadata, recommendations, series fill, and Hardcover list sync.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed - N/A, no env, config, deployment, or upgrade-path change
- [ ] `CHANGELOG.md` `[Unreleased]` section updated - N/A, release-time only per `AGENTS.md`
- [ ] Wiki pages updated if user-facing behaviour changed - N/A, no wiki source is available in this checkout; scope and limitations are documented above

## Test plan

- [x] `go test ./internal/bookhydrate ./internal/db ./internal/metadata ./internal/hardcoverlistsyncer ./internal/api`
- [x] `gofmt -l cmd internal`
- [x] `go vet ./...`

### DIFF
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -418,7 +418,8 @@ func main() {
 		WithSeriesRepo(seriesRepo).
 		WithTokenSource(func(ctx context.Context) string {
 			return api.GetHardcoverAPIToken(ctx, settingsRepo)
-		})
+		}).
+		WithEditionHydration(editionRepo, metaAgg)
 	sched.WithHardcoverSyncer(hcSyncer)
 	sched.WithLogRepo(logRepo, cfg.LogRetentionDays)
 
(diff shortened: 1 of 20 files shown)


magrhino:
Rebased this branch onto `main` and resolved the conflicts.

This PR changes the shape and amount of metadata we store, especially around Hardcover edition hydration, so I’d appreciate a careful review from you, @vavallee, as always.

One design question I had after pushing: should the branch’s current `HardcoverForeignID` concept eventually be modeled as a more general `AlternateIdentifiers[]` field instead? That would let us track Hardcover IDs now while leaving room for other provider IDs later, but I didn’t want to broaden the schema/API shape without review.

Issue #863, opened 2026-05-28
feat: bulk 'Refresh Metadata' for all (or selected) authors


vavallee:
## Why

After a fresh Calibre/Readarr import, every author and book in Bindery starts with empty metadata fields (covers, descriptions, ratings, author photos, expanded book list). Today the only way to populate them is to navigate into each author one at a time and click **Refresh Metadata** per author. For a library of any size this is impractical and is the most consistent first-run frustration we see from new users in Discord (most recent: Jashun44, 2026-05-28; prior: original report in #314).

## What

Add a way to trigger metadata refresh across many authors at once. Two reasonable shapes:

1. **Bulk-select on the Authors page** — checkbox column + "Refresh metadata" button in the bulk-action bar. Lowest UI debt; fits the existing list-select pattern. Honours user intent (refresh only what's selected).
2. **"Refresh all" affordance** at the Authors page header or Settings → Library — kicks off a backgrounded job that walks every author, queueing per-provider rate limits. Show progress (N of M) in a banner the way the library scan does.

Both, ideally. (1) for surgical use, (2) for first-run "I just imported everything, populate it all."

## Constraints

- The refresh is per-author already and is slow because of upstream provider rate limits (OpenLibrary, Hardcover, Google Books). The job must respect those — sequential or low-concurrency, with backoff on 429/403. The OpenLibrary 403 issue from #848 means `BINDERY_CONTACT` should be honoured by the bulk path too.
- Must not block the UI thread. Background-job state should survive page reload (Settings → General "Library Scan" status pattern is the right shape — show last-completed timestamp + in-progress banner).
- Per-author refresh failures (provider down, ID drifted) must not abort the whole run. Per-author errors should be reported in the job summary, not surfaced as a single fatal toast.
- Author photos populate from the same upstream provider call as descriptions/covers, so this also fixes the "author photo never appeared" symptom that comes up alongside.

## Scope

Out: changing what `Refresh Metadata` *does* on a single author (no behaviour change to the per-author path). In: a new endpoint + UI affordance that fans out the existing per-author refresh across N authors with progress reporting.

## References

- Per-author handler: `internal/api/authors.go` (search for `LastMetadataRefreshAt = &now`)
- Existing background-job-with-banner pattern: `web/src/pages/settings/GeneralTab.tsx` Library Scan section

magrhino:
I have a branch that worked on this. Needs some polish before I ship. Sorry its been sitting there for over a week, but it fixes most of these issues.

[...]

PR #887, opened 2026-05-31
feat(authors): preserve alternate metadata identities

magrhino:
## Summary

Refs #863. This PR lays the groundwork for a future bulk author metadata refresh by making author refresh/relink operate on the correct upstream identity instead of repeatedly refreshing stale import IDs from Calibre, ABS, Readarr, or other placeholder sources. It preserves every known provider/import author ID on one canonical author row, adds a manual metadata-linking flow, and makes imports/rollback/merge paths identity-aware so the later bulk button can refresh useful metadata instead of just requerying the wrong source.

- Add `author_identifiers` as a provider/import identity map while keeping `authors.foreign_id` as the primary identity; backfill existing author IDs, infer provider names from ID prefixes, and upsert the current primary ID on author create/update.
- Teach author lookup, duplicate detection, merge, DNB upgrades, Hardcover list sync, CSV/Goodreads/Readarr migrations, ABS import, and Calibre import to resolve authors by primary or alternate ID so imports attach to existing authors instead of creating duplicates.
- Preserve alternate identities during author merges and metadata relinks, including the previous import/provider ID, so future refreshes can promote a canonical upstream ID without losing the source ID that originally matched the library.
- Expand rollback snapshots for ABS and Calibre author imports to include identifier state, then remove only identifiers added by the rolled-back import while preserving identifiers that existed before the run.
- Add manual relink candidate search across configured metadata providers without changing the normal primary-provider author search path; the relink endpoint now accepts a selected `foreignAuthorId` and returns canonical conflict details when the selected identity is already owned locally.
- Scope relink and alternate-ID lookups through the requesting user where applicable, preserving multi-user isolation while still keeping legacy unowned rows visible.
- Add frontend API error bodies/statuses, a metadata-link modal on author detail pages, conflict actions from Add Author to open the existing author or launch the metadata-link modal, and i18n/test coverage for the new flow.
- Document the relink candidate endpoint and manual relink request body in `docs/API.md`.

Behavior notes for review:

- This does not implement the bulk refresh job or Authors-page bulk button from #863.
- Existing single-author refresh behavior is preserved; this branch makes the author identity model robust enough that future bulk refresh work can target canonical metadata IDs.
- Normal Add Author search remains primary-provider based. The broader multi-provider candidate search is only used by the manual relink flow.
- `CHANGELOG.md` is intentionally untouched because Bindery changelog authoring is release-time only per repo guidance.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed - N/A, no deployment/config change
- [ ] `CHANGELOG.md` `[Unreleased]` section updated - N/A, release-time only per AGENTS.md
- [ ] Wiki pages updated if user-facing behaviour changed - N/A, no wiki content is available in this checkout

## Test plan

- [x] `git diff --check upstream/main...HEAD`
- [x] `gofmt -l cmd internal`
- [x] `go vet ./...`
- [x] `go test -race -coverprofile=<tmp>/bindery-coverage.out -covermode=atomic ./cmd/... ./internal/...`
- [x] `cd web && npm run lint` - passed with warnings only
- [x] `cd web && npm test`


magrhino:
Force pushed to clear blocks and keep it updated. I know this is a big dif, just ping me if you have any questions or would like changes.

vavallee:
Hey @magrhino, this is in good shape (reviewed it for correctness, no blockers found) but it needs a rebase before it can land. Two things:

1. It's ~23 commits behind `main` and predates several merged PRs (#977 monitored-status pill, #972 subpath deploys, #960/#962/#963/#964 onboarding/scan/test-buttons). A plain merge would surface those as conflicts/reverts.

2. Migration number collision: #992 just merged and took `051`, so your `051_*.sql` needs to become `052_*.sql` (the startup `assertUniqueMigrationVersions` guard will fail otherwise).

Once you've rebased onto current `main` and bumped the migration, re-run CI and I'll merge. This also unblocks #863 (bulk refresh metadata), which is gated on the `author_identifiers` scaffolding you add here.

magrhino:
Hey, the rebase/conflict cleanup is done.

The remaining CI failure looks unrelated to this PR’s code. `Security / Container Scan` is failing in the Trivy vulnerability scan, and I verified the same Container Scan job is currently failing on `main` too. This branch does not change `Dockerfile`, workflow files, `go.mod`/`go.sum`, or frontend package manifests, and the branch-specific checks are green.

I don’t currently have access to the uploaded Trivy SARIF/code-scanning alert, so I can’t see the exact CVE from here. If you can share the alert details or give me access, I’m happy to help chase the CVE/security fix.

Issue #232, opened 2026-04-18
Language filter: surface unknown-language books and add "treat unknown as pass/fail" setting


vavallee:
## Problem

When OpenLibrary returns no language metadata for a book (at the work level), Bindery currently passes the language filter silently — the book is imported even if the user's metadata profile requires a specific language. This behaviour was acknowledged in #224.

While the root cause (missing OpenLibrary data) is outside Bindery's control, the silent pass creates two user-facing problems:

1. **Unexpected imports**: Users with a strict language filter are surprised to find foreign-language or language-unknown books in their library, with no indication of why they bypassed the filter.
2. **No control**: There is no setting to choose the desired behaviour; the only option is "pass silently."

### Option A — Warning badge on imported books

Add a visible indicator (e.g. a "Language unknown" badge or tooltip) on any book that was imported despite having no confirmed language metadata. This lets users quickly audit and manually remove unwanted items.

### Option B — Setting: "Treat unknown language as: Pass / Fail"

Add a per-metadata-profile (or global) setting:

> **When book language is unknown (not returned by metadata source):**
> - `Pass` (current default behaviour — import the book)
> - `Fail` (skip the book and log a reason; surface in Queue/History as "rejected: language unknown")

The default could remain `Pass` for backwards compatibility, but users who want strict filtering would be able to opt into `Fail`.

### Ideal: Both A and B

Implement the setting (B) and, when `Pass` is chosen, show the warning badge (A) so users are always aware of what bypassed the filter.

## Related Issues

- #224 — non-English books ingested when metadata profile requires English (OpenLibrary language field empty at work level)

## Notes

- The same issue likely applies to other metadata sources (Google Books, Open Library editions vs. works, etc.) where language may be absent — the fix should be source-agnostic where possible.
- This is a documentation/UX gap in addition to a feature request: the current "fail open" behaviour is not documented in the UI or help text, so users have no reason to expect it.

Issue #224, opened 2026-04-18
bug: non-English books ingested when metadata profile requires English (OpenLibrary language field empty at work level)


vavallee:
## Problem

When an author has a metadata profile with `allowedLanguages: eng`, non-English translations of their works are still ingested because OpenLibrary does not populate the `language` field at the **work** level for translated editions.

The language filter in the metadata profile compares against the book's stored `language` column. If that column is empty (which it is for most OL works), the filter passes the book through regardless of the actual language of the edition.

**Observed:** Arthur Golden (OL2645460A) with the Standard profile (`allowedLanguages: eng`) had 6 non-English books ingested: Turkish, Spanish (×2), Dutch, German, and Polish translations of *Memoirs of a Geisha*. All had `language = ''`.

## Root cause

OpenLibrary's `/works/{id}.json` endpoint does not include a `language` field — language is only available at the **edition** level (`/books/{edition_id}.json`). Bindery ingests at the work level and doesn't fetch per-edition language data, so `language` is never populated for OL-sourced books.

## Proposed fix

Two complementary approaches:

**A — Edition-level language sampling (preferred)**
When resolving a work's language, fetch the first few editions for that work from OL (`/works/{id}/editions.json?limit=5`) and take the majority or most-common language. Cache it alongside the work so it isn't re-fetched on every refresh.

**B — Skip works with no language when profile requires one**
If `allowedLanguages` is non-empty and the resolved language is still empty after best-effort lookup, skip the work rather than letting it through. This is conservative but prevents the current silent pass-through.

**C — Title heuristic (last resort)**
Detect non-ASCII characters or common non-English patterns in the title as a signal that the edition is not in the allowed language. Fragile and not recommended as the sole mechanism, but could be a tie-breaker.

A + B together would close the gap: try to get a language from editions, and if it's still unknown, fail closed when the profile has a language requirement.

## Impact

Any author with a metadata profile restricting language is affected. Users who add a well-known author (e.g. Stephen King, J.K. Rowling) will silently accumulate foreign-language editions unless they manually delete them.

vavallee:
Fixed in #228, released in v0.18.3. Went with approach B — empty language now rejects when the profile has a language restriction. The search-index lookup (approach A) was already there; this closes the gap for works it doesn't cover.

PR #890, opened 2026-06-01
fix(hardcover): propagate canonical-book language to plug author-works filter leak (#889)

### LINKED ISSUE #889: Foreign-edition series leaks into English author bibliography: Spanish
### Pre-flight

### Current behaviour

When adding an author and you've selected English. You'll get both english and foreign copies of the books. 

Example: 
Michael Connelly
Malcolm Gladwell
James Patterson

### Expected behaviour

Foreign-language editions/volumes of an English author's work should not be surfaced as English series 

### Steps to reproduce

1.  Add author 
2. Author's books pulled from API
3. All variations are added both English and international 

### Trace logs

```text
None
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v1.15.3

### Environment details

Docker, qBittorrent, SABnzdb, Hardcover api, etc..

vavallee:
## Summary

- `GetAuthorWorksByName` never queried `language { language }`, so every Hardcover-sourced supplemental book arrived with `Language=""` and slipped past the metadata profile's `allowed_languages` filter via the default "unknown language → pass" branch.
- Same gap as #224 / #232; #889 was the user-reported recurrence (Spanish editions of Michael Connelly, James Patterson, etc. leaking into English-only author bibliographies).
- Add the projection, plumb the field through `hcBook` + `toBook`, and assert both the query shape and the `en`/`es` → `eng`/`spa` mapping in the existing `TestGetAuthorWorksByName_WithToken`.

OpenLibrary's work-level tail (works not indexed by `/search.json`) is the other half of the historical gap and is left for a follow-up — that needs edition sampling and is a bigger change.

Closes #889.

## Test plan

- [x] `go test ./internal/metadata/hardcover/... -count=1` — `TestGetAuthorWorksByName_WithToken` now asserts the new `language` projection in the query and the `en→eng` / `es→spa` propagation.
- [x] `go test ./internal/metadata/... ./internal/api/... -count=1` — green, no regressions.
- [ ] Manual: add Michael Connelly with the Standard profile post-merge and confirm Spanish editions no longer appear in the books list.

### DIFF
--- a/internal/metadata/hardcover/client.go
+++ b/internal/metadata/hardcover/client.go
@@ -199,6 +199,7 @@ func (c *Client) GetAuthorWorksByName(ctx context.Context, authorName string) ([
 			audio_seconds
 			default_audio_edition_id
 			default_ebook_edition_id
+			language { language }
 			contributions {
 				author { id name slug }
 			}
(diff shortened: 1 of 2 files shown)


vavallee:
Followup tracked as #891 for the OpenLibrary half of the leak (edition-sampling for work-level language).

Issue #947, opened 2026-06-03
Bulk endpoints bypass per-user ownership checks (IDOR under BINDERY_ENFORCE_TENANCY)


vavallee:
## Summary

The bulk endpoints perform no ownership/tenancy check, so when `BINDERY_ENFORCE_TENANCY=true` a non-admin user can delete/modify another user's authors, books, and wanted state by passing the victim's resource IDs. The single-resource equivalents all gate with `auth.CheckOwnership`; the bulk path does not.

This is the same multi-user isolation class hardened in v1.15.1, missed on the bulk surface.

## Evidence

- `internal/api/bulk.go` — **0** `CheckOwnership` calls (`grep -c CheckOwnership internal/api/bulk.go` = 0). Compare books.go (8) and authors.go (6).
- Routes are registered at the authenticated **user** level, not admin:
  - `cmd/bindery/main.go:722` `POST /author/bulk` → `AuthorsBulk`
  - `cmd/bindery/main.go:735` `POST /book/bulk` → `BooksBulk`
  - `cmd/bindery/main.go:749` `POST /wanted/bulk` → `WantedBulk`
  (The `RequireAdmin` group above them closes before line 722.)
- The bulk setters fetch the full object (which carries `OwnerUserID`) but never consult it, and the repo layer mutates purely `WHERE id = ?` (e.g. `AuthorRepo.Delete`, `BookRepo.SetExcluded`, `BookRepo.Update` — no owner scoping).

## Impact

With tenancy enforced: `POST /api/v1/author/bulk {"ids":[<victim id>],"action":"delete"}` (and the book/wanted equivalents) lets any authenticated non-admin delete, unmonitor, exclude, retarget, or trigger searches on another user's library. When tenancy is **off** it's latent (single-resource handlers also no-op), so this is not a regression in the default config — but it defeats the isolation the flag promises.

## Fix

In each bulk loop iteration, fetch the resource and skip (record per-item 404/forbidden) when the caller doesn't own it, before mutating — mirroring the single-resource handlers:

```go
if !auth.CheckOwnership(r.Context(), obj.OwnerUserID) {
    // record per-item forbidden/not-found; continue
    continue
}
```

Apply to `AuthorsBulk`, `BooksBulk`, `WantedBulk`. Add a `bulk_test.go` case with tenancy enabled asserting user B cannot affect user A's resources (no such case exists today).

Found during a code sweep.

PR #955, opened 2026-06-03
test(api): cover auth user-management + OIDC SetProviders handlers

vavallee:
## What

Adds handler-level test coverage for four mutating, security-relevant auth handlers that were previously at **0%** coverage:

- `internal/api/auth_users.go` — `Create`, `Delete`, `SetRole`
- `internal/api/auth_oidc.go` — `SetProviders`

Test-only change. No production `.go` files were modified.

## Coverage delta (target functions)

| Handler | Before | After |
|---|---|---|
| `auth_users.Create` | 0% | 84.0% |
| `auth_users.Delete` | 0% | 100% |
| `auth_users.SetRole` | 0% | 100% |
| `auth_oidc.SetProviders` | 0% | 85.3% |

Residual gaps in Create/SetProviders are unreachable internal-error branches (password-hash failure, JSON marshal failure, `settings.Set` DB error) that an in-memory DB cannot trigger.

## Scenarios

**Create:** success (returns created user, persisted), admin-role creation, duplicate username, empty username, empty password, malformed JSON, local-auth-disabled (403), non-admin rejected (403).
**Delete:** success, not-found no-op (200), non-numeric id (400), last-admin guard (400, row preserved).
**SetRole:** promote, demote (two-admin), invalid role (400), malformed JSON, non-numeric id, last-admin demotion guard (400, role preserved).
**SetProviders:** valid array replace, full-array-replace semantics (drops omitted ids), omitted-secret preservation merge, new-provider-without-secret rejection (400, nothing persisted), malformed JSON, empty-array clears, non-admin rejected (403, nothing persisted).

## Authorization findings

The admin gate for all four endpoints is the router-level `auth.RequireAdmin` middleware (`cmd/bindery/main.go:701-711`), **not** any check inside the handler bodies. The handlers read no role from the request context. This is the existing architecture and it is correct: the `_NonAdminForbidden` tests mount each handler behind `RequireAdmin` exactly as `main.go` does and confirm a non-admin caller is rejected with 403 before the handler runs and nothing is mutated.

**No authz/security gap found in these handlers.** Note: issue #947 (bulk IDOR under `BINDERY_ENFORCE_TENANCY`) concerns `bulk.go` (`AuthorsBulk`/`BooksBulk`/`WantedBulk`), which are mounted at the authenticated-**user** level without ownership checks — a different surface from the admin-gated handlers covered here. The OIDC provider API has no per-provider ownership model (it's a single global settings array gated by `RequireAdmin`), so the #947 class does not apply to it.

## Documented current (non-ideal) behavior

`Create` with a duplicate username returns **500**, not 409 — the repo's UNIQUE-constraint error is surfaced verbatim. The test pins this as current behavior; 409 would be the more correct status. Flagged here rather than silently enshrined.

## Verify

- `go test ./internal/api/...` passes (no skips added by this PR).
- `go vet ./internal/api/...` clean.

### DIFF
--- a/internal/api/auth_oidc_setproviders_test.go
+++ b/internal/api/auth_oidc_setproviders_test.go
@@ -1,0 +1,244 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-chi/chi/v5"
+
+	"github.com/vavallee/bindery/internal/auth"
+	"github.com/vavallee/bindery/internal/auth/oidc"
+	"github.com/vavallee/bindery/internal/db"
+)
+
+// These tests cover OIDCHandler.SetProviders (PUT /auth/oidc/providers), which
+// was at 0% handler coverage. SetProviders does a full-array replace of the
+// configured provider set and persists it via SettingsRepo, preserving secrets
+// that the client omits.
+//
+// AUTHORIZATION NOTE: like the user-management mutations, the admin gate for
+// SetProviders is the router's RequireAdmin middleware
+// (cmd/bindery/main.go:704), not anything inside the handler. The handler reads
+// no user/role from the request context at all. TestSetProviders_NonAdmin
+// Forbidden mounts the handler behind RequireAdmin to assert the authorization
+// contract holds at the wiring level.
+
+func newOIDCSetProvidersFixture(t *testing.T) (*OIDCHandler, *db.SettingsRepo) {
+	t.Helper()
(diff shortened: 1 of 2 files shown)


vavallee:
Reviewer note (post-agent): revised `TestUserMgmt_Create_DuplicateUsername` so it no longer pins the `500` status. It now asserts the durable invariant (duplicate rejected with an error status — accepts 409 or 500 — and the original row is untouched), so a future fix that correctly returns 409 won't break this test. The duplicate-username → 500 (should be 409) status wart remains a tracked follow-up.

PR #1009, opened 2026-06-06
feat(web): expose per-indexer priority in the Indexers settings form

vavallee:
Part of the LazyLibrarian feature comparison ("provider priority" — prefer one source over another).

## Finding
The capability already exists end-to-end: `models.Indexer.Priority` is persisted by the DB repo and create/update API, and the ranker factors it into release scoring (`internal/indexer/searcher.go` — a higher-priority indexer wins ties and can outweigh small quality gaps). Protocol preference also already exists via **delay profiles** (`preferred_protocol`). The only gap was the **UI**: the Indexers settings form had no priority field, so manually-added indexers were stuck at priority 0 (only Prowlarr-synced ones ever got a priority).

## Change (frontend only)
- Add a **Priority** number input to the add and edit indexer forms, wired into the existing `addIndexer`/`updateIndexer` payloads.
- Add `priority` to the web `Indexer` type and a `parsePriority` helper.
- i18n keys + hint ("higher wins ties; e.g. set Usenet indexers above torrent ones").
- Default 0 → behaviour unchanged for anyone who doesn't set it.

No backend change needed.

## Verify
`tsc` clean, `npm run build` clean, `npm run lint` no new warnings, `npm test` 273/273 (updated two indexer mocks + the add-indexer payload assertion for the new field).

### DIFF
--- a/web/src/api/indexers.ts
+++ b/web/src/api/indexers.ts
@@ -8,6 +8,7 @@ export interface Indexer {
   url: string
   apiKey: string
   categories: number[]
+  priority: number
   enabled: boolean
   prowlarrInstanceId?: number
 }
--- a/web/src/i18n/locales/en.json
+++ b/web/src/i18n/locales/en.json
@@ -355,7 +355,9 @@
         "apiKey": "API Key",
         "categories": "Categories",
         "categoriesPlaceholder": "Categories (e.g. 7020, 7120, 3030)",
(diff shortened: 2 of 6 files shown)


SolomonIxanSmith:
Not sure if this matters but I think prowlarr has the opposite convention for priority (1 is Highest, 50 is Lowest). Also, this doesn't change doesn't seem to pull in the priority as set by prowlarr for those indexers that get imported.  Not a big deal in the end, but I thought it worth mentioning

PR #1018, opened 2026-06-07
fix(web): restore author detail alias controls

magrhino:
## Summary

Restores the frontend affordance for scoped author alias deletion backed by the existing `DELETE /author/{id}/aliases/{aliasID}` API introduced in 056f2f7c2fcd00ed05bcc69abe5c9a1731c92221, so aliases can again be removed from the author detail page after confirmation. This also adds a small no-dependency author description renderer that supports limited inline markdown without enabling external links.

- Add the author alias delete API client call, hover/focus remove controls on alias chips, confirmation handling, local state removal after success, and regression coverage for confirm/cancel flows.
- Render author descriptions through a constrained markdown component that strips citation/link reference noise, supports inline emphasis/reference text, and provides show more/show less behavior for long descriptions.
- Cover author description markdown rendering and expansion behavior in `AuthorDetailPage` tests.

## Alternatives considered

A full markdown dependency was considered, but this use case only needs a narrow subset for author descriptions. Avoiding a dependency also avoids enabling external links, where validating link safety would add security work that is unnecessary for this purpose.

- Add the same limited markdown treatment to book descriptions.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed (N/A - no deployment/config changes)
- [ ] Wiki pages updated if user-facing behaviour changed (N/A - no wiki update for this narrow author detail restoration)

## Test plan

- [x] `npm --prefix web test -- AuthorDetailPage.test.tsx`
- [x] `npm --prefix web run lint` (passed with warnings only)
- [x] `npm --prefix web run typecheck`
- [x] `npm --prefix web run build`
- [x] `make test-web`
- [ ] `make check` (not run; frontend-only change validated with targeted web checks)

### DIFF
--- a/changelog.d/author-detail-alias-description.md
+++ b/changelog.d/author-detail-alias-description.md
@@ -1,2 +1,0 @@
-### Changed
-- Author detail pages now restore alias removal controls and render limited inline markdown in descriptions without enabling external links.
--- a/changelog.d/author-detail-alias-description.md
+++ b/changelog.d/author-detail-alias-description.md
@@ -1,0 +1,2 @@
+### Changed
+- Author detail pages now restore alias removal controls and render limited inline markdown in descriptions without enabling external links.
--- a/web/src/api/authors.ts
+++ b/web/src/api/authors.ts
@@ -132,6 +132,8 @@ export const authorsApi = {
       body: candidate ? JSON.stringify(candidate) : undefined,
     }),
   listAuthorAliases: (id: number) => request<AuthorAlias[]>(`/author/${id}/aliases`),
+  deleteAuthorAlias: (authorId: number, aliasId: number) =>
+    request<void>(`/author/${authorId}/aliases/${aliasId}`, { method: 'DELETE' }),
   // listAuthorSeries returns the series the author has books in. Backs the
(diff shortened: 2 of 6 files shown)


vavallee:
Merged, thanks! Alias delete is clean and the markdown renderer is a nice bonus, especially keeping it dependency-free and element-based so there's no XSS surface.

One small ask for next time: this was really two changes (alias controls + a new description-markdown component that touches how every author description renders). Easier to review and revert if those land as separate PRs, and the title here undersold the markdown half. No big deal, just a heads up. Appreciate the solid tests on both.

magrhino:
> Merged, thanks! Alias delete is clean and the markdown renderer is a nice bonus, especially keeping it dependency-free and element-based so there's no XSS surface.
> 
> One small ask for next time: this was really two changes (alias controls + a new description-markdown component that touches how every author description renders). Easier to review and revert if those land as separate PRs, and the title here undersold the markdown half. No big deal, just a heads up. Appreciate the solid tests on both.

Will do! You're right it should of been two PRs, just saw another opportunity and got carried away. Will split on easier reviewable boundaries next time! Thank you as always.

PR #1027, opened 2026-06-08
ci(deploy-dev): make bindery-dev auto-follow main

vavallee:
## Why
While checking Argo, found **bindery-dev** (homelab, `bindery-dev.autonomy.ninja`) was running `dev-a8f87d3` from **May 17 — 256 commits behind main**. Root cause: the `deploy-dev` job only fired on pushes to the `development` branch, but the team went main-only, so `development` (and thus the bindery-dev Argo app that tracks it) froze.

I've already brought the live dev env current out-of-band (it now runs `sha-6686049`, the current main HEAD). This PR makes it *stay* current.

## What
Retarget `deploy-dev` to run on every push to **main**: after the image build, force-update the `development` branch to main's tree + a `values-dev.yaml` override pinning the just-built `sha-<short>` image. The GitOps-managed `bindery-dev` Argo app tracks `development`, so it follows main automatically.

- `development` is now a machine-maintained dev-deploy branch (no human commits; I removed its branch protection), so a direct force-push replaces the old per-push PR dance — no PR churn, no flaky admin-merge.
- The `[skip ci]` subject stops the development push from re-triggering CI.
- The prod `bindery` app reads only `values.yaml`, so it is unaffected by `values-dev.yaml`.
- `needs: image` guarantees the sha image exists before Argo pulls it.

Validated YAML + `bash -n` on the step. First exercised on the next push to main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -282,69 +282,49 @@ jobs:
           push-to-registry: true
 
   # ── Deploy to bindery-dev (development branch only) ────────────────────────
-  # Runs after image build on every push to development. Updates
-  # values-dev.yaml with the SHA image tag so ArgoCD picks up the new build.
+  # Runs after the image build on every push to main. Force-updates the
+  # `development` branch to main's tree plus a values-dev.yaml override pinning
+  # the just-built sha image, so the GitOps-managed bindery-dev ArgoCD app
+  # (which tracks `development`) follows main automatically.
+  #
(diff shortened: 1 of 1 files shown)


vavallee:
Superseded by #1043. That 'Assemble dev preview' workflow keeps development current with main AND carries dev-preview-labelled PRs, whereas this PR would force development to equal main on every push and wipe any staged previews. Closing in favour of the preview approach.

PR #1116, opened 2026-06-11
fix(deluge): fetch .torrent in Bindery before submitting to daemon

### LINKED ISSUE #1110: Deluge: torrent grabs fail with DNS lookup error — daemon fetches the indexer URL instead of Bindery
### Describe the bug

Grabbing a torrent release whose download URL is an HTTP `.torrent` link fails when the download client is **Deluge**. Bindery surfaces the Deluge daemon's error verbatim:

```
download torrent from url: rpc error 4: Failure: [Failure instance: Traceback: <class 'twisted.internet.error.DNSLookupError'>: DNS lookup failed: no results for hostname lookup: prowlarr.
/usr/lib/python3.13/site-packages/twisted/internet/_resolver.py:192:deliverResults
...
]
```

### Root cause

For non-magnet torrent URLs, the Deluge client calls the Web UI RPC `web.download_torrent_from_url` (`internal/downloader/deluge/client.go`), which asks the **Deluge daemon** to fetch the `.torrent` from the indexer URL itself. Prowlarr hands out enclosure URLs that point at its own host (e.g. `http://prowlarr:9696/<indexer>/download?...`). When Deluge runs in a different network namespace than Prowlarr — e.g. behind a VPN container such as `binhex/arch-delugevpn`, whose custom `NAME_SERVERS` replace Docker's embedded resolver (`127.0.0.11`) — the daemon can't resolve the `prowlarr` hostname, and the RPC fails with a `twisted DNSLookupError`.

This is the same containerised / VPN failure mode already fixed for every other download client, all of which fetch the file **inside Bindery** (which *can* resolve the indexer host) and submit the content to the daemon:

- qBittorrent (#639)
- Transmission (#873)
- SABnzbd (#864)
- NZBGet (#837)

Deluge's HTTP path was never converted, so it still delegates the fetch to the daemon. Magnet grabs are unaffected — they go through `core.add_torrent_magnet`, which needs no HTTP fetch.

### Expected behavior

Bindery fetches the `.torrent` itself and submits the bytes to Deluge, so the daemon never needs to reach the indexer URL — matching the other clients.

### To reproduce

1. Use Deluge as the torrent download client, on a different Docker network than Prowlarr (e.g. `binhex/arch-delugevpn` with custom `NAME_SERVERS`, or any VPN-namespaced setup).
2. Configure a Prowlarr indexer that returns HTTP `.torrent` enclosure URLs (`http://prowlarr:9696/...`).
3. Grab a release → the `DNS lookup failed: no results for hostname lookup: prowlarr` error above.

### Proposed fix

In the Deluge client, fetch the `.torrent` bytes in Bindery (with the same SSRF validation, size cap, and magnet-redirect handling the other clients use) and submit them via `core.add_torrent_file` (base64 `filedump`), mirroring the qBittorrent / Transmission implementations. `core.add_torrent_file` also returns the infohash directly, so the add path no longer needs to poll the torrent list to discover the new hash.

### Environment

- Deluge behind `binhex/arch-delugevpn` (PIA WireGuard), Prowlarr via the linuxserver image, Bindery served under `/bindery`.

vavallee:
## Problem

Closes #1110.

When Deluge runs inside a VPN container with a custom DNS namespace (a common homelab topology), its daemon cannot resolve Docker-internal hostnames such as `prowlarr`. Calling `web.download_torrent_from_url` asks the **daemon** to fetch the indexer URL, which fails with a `DNSLookupError` because the daemon lives in the VPN namespace where only external DNS works.

## Fix

Replace the two-step `web.download_torrent_from_url` → `web.add_torrents` flow with a **Bindery-side fetch** followed by `core.add_torrent_file`:

1. Bindery fetches the `.torrent` bytes in its own network namespace (where `prowlarr`, `jackett`, etc. resolve normally).
2. The raw content is base64-encoded and submitted to the Deluge daemon via `core.add_torrent_file(filename, filedump, options)`, which returns the infohash directly.

This matches the pattern already used by every other download client — qBittorrent, Transmission, SABnzbd, and NZBGet all fetch on Bindery's side.

## Changes

- `internal/downloader/deluge/client.go`: replaces `addTorrentURL` with `addTorrentFile` + `fetchTorrentContent` (mirrors the qBittorrent implementation). Removes the `addMu` serialisation mutex and the before/after hash-snapshot polling loop — neither is needed when the RPC returns the hash directly. Magnet-redirect handling is preserved: a `302 → magnet:` response transparently falls through to `addMagnet`.
- `internal/downloader/deluge/client_test.go`: removes the now-obsolete `TestAddTorrent_URL_HashLookupTimeout` test; updates `TestAddTorrent_URL` to spin up a local `.torrent` file server and verify `core.add_torrent_file` is called with the correct base64 payload and filename; adds `TestAddTorrent_URL_MagnetRedirect` covering the redirect-to-magnet path.
- `internal/downloader/deluge/export_test.go`: removes the obsolete `SetHashPollTimeout` export; adds `SetValidateTorrentURL` so tests can bypass the SSRF check for loopback test servers.

### DIFF
--- a/internal/downloader/deluge/client.go
+++ b/internal/downloader/deluge/client.go
@@ -5,12 +5,14 @@ package deluge
 import (
 	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
 	"log/slog"
 	"net/http"
 	"net/http/cookiejar"
+	"path"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -18,11 +20,13 @@ import (
 
 	"github.com/vavallee/bindery/internal/downloader/nethint"
 	"github.com/vavallee/bindery/internal/downloader/urlbase"
+	"github.com/vavallee/bindery/internal/httpsec"
+	"github.com/vavallee/bindery/internal/useragent"
 )
 
-// hashPollTimeout is the maximum time to wait for a newly-added torrent's hash
-// to appear in the torrent list.
-var hashPollTimeout = 30 * time.Second
+// maxTorrentFileBytes caps the torrent payload Bindery will fetch before
+// uploading it to Deluge.
+var maxTorrentFileBytes int64 = 50 << 20
 
 // Client interacts with the Deluge Web UI JSON-RPC API.
 // Authentication is cookie-based: Login() posts auth.login which sets a
@@ -32,13 +36,13 @@ var hashPollTimeout = 30 * time.Second
 //   - Password → password  (Deluge Web UI uses a single password, no username)
 //   - Category  → label    (applied via the label plugin after adding)
 type Client struct {
-	baseURL  string
-	password string
-	http     *http.Client
-	mu       sync.Mutex // guards loggedIn
-	addMu    sync.Mutex // serialises addTorrentURL: keeps before/after hash diff atomic
-	loggedIn bool
-	reqID    atomic.Int64
+	baseURL            string
+	password           string
+	http               *http.Client
+	validateTorrentURL func(string) error
+	mu                 sync.Mutex // guards loggedIn
+	loggedIn           bool
+	reqID              atomic.Int64
 }
 
 type rpcRequest struct {
@@ -118,9 +122,17 @@ func (c *Client) AddTorrent(ctx context.Context, magnetOrURL, label string, seed
 		}
 		hash = h
 	} else {
-		h, err := c.addTorrentURL(ctx, magnetOrURL, label)
+		h, mag, err := c.addTorrentFile(ctx, magnetOrURL)
 		if err != nil {
 			return "", err
+		}
+		if mag != "" {
+			// The indexer redirected to a magnet link; hand off to the magnet path.
+			mh, merr := c.addMagnet(ctx, mag)
+			if merr != nil {
+				return "", merr
+			}
+			h = mh
 		}
 		hash = h
 	}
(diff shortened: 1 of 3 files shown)


johnistheman:
Thanks for picking this up so fast — this is the same approach I'd landed on for #1110, and dropping the before/after hash-poll and `addMu` (since `core.add_torrent_file` returns the infohash directly) is cleaner than what I had.

One heads-up while you're in here: this `fetchTorrentContent` mirrors qBittorrent's byte-for-byte, so Deluge is now the **third** copy of it. The qBit and Deluge copies are in sync, but the **Transmission** one has quietly drifted:

- `internal/downloader/transmission/client.go` fetches with a plain `io.ReadAll(io.LimitReader(resp.Body, 50<<20))` — it **silently truncates** at the cap instead of erroring like qBit/Deluge's `readLimited`, and
- it does a plain GET with default redirect-following, so a `302 → magnet:` response fails with `unsupported protocol scheme` instead of falling through to the magnet path the other two handle.

Both are low-severity in practice (a >50 MB `.torrent` is basically impossible, and the magnet-redirect gap is the only part Transmission users would realistically hit), so no urgency — just flagging that it's now three copies that can drift independently.

If you're open to it, I'm happy to put up a follow-up once this lands that extracts a small shared `torrentfetch` helper (fetch + SSRF validate + redirect/magnet + size cap), with each client keeping its own submit step (multipart / `metainfo` / `core.add_torrent_file`). Your call on whether that's a direction you want vs. keeping per-client copies.

PR #1141, opened 2026-06-15
Fix/naming template validation on load

hazeledmands:
## Summary

The server returns an empty string for naming.bookTemplate / naming_template_audiobook when they've never been explicitly set (the renamer falls back to compiled-in defaults at import time). This caused "Template cannot be empty" to appear immediately on the File Naming settings page even without user interaction.

image

Fix: gate all validation feedback behind a `touched` state that is set on first onChange or onBlur. The Save button remains disabled for invalid templates regardless of touched state, so the constraint is still enforced.

## Checklist

- [x] Tests added or updated

## Test plan

- [x] `make check` (or the individual `go test ./cmd/... ./internal/...` and `cd web && npm run build` steps)

### DIFF
--- a/changelog.d/naming-template-touched-validation.md
+++ b/changelog.d/naming-template-touched-validation.md
@@ -1,0 +1,2 @@
+### Fixed
+- **File Naming settings** — "Template cannot be empty" validation error no longer appears on page load when no template has been saved yet (the server uses a built-in default for empty templates, so the error was a false alarm until the field is actually edited).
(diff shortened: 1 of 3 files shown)


vavallee:
Thanks @hazeledmands — good catch on the false empty-error. I've opened #1194 which lands your fix with one adjustment: it gates *only* the empty-template error behind `touched`. Traversal / unknown-token errors on an already-saved template are real misconfigurations, so #1194 keeps showing those immediately on load (otherwise a bad saved template would silently show no warning until the field is touched). Your commits are preserved with attribution. Closing in favour of #1194 so it lands in the upcoming release.

vavallee:
Follow-up: after re-checking current main, this is already handled — #1107 (v1.20.0) changed the empty-template feedback from the red error in your screenshot to a muted 'Required …' hint (role=alert, Save still disabled), and traversal/unknown-token errors show immediately. So no further change is needed. Thanks again for flagging it.

PR #1194, opened 2026-06-18
fix(settings): stop naming-template empty-error firing on load (supersedes #1141)

vavallee:
Lands @hazeledmands's fix from #1141 with a narrower validation gate.

**Original fix (#1141):** the File Naming settings page showed "Template cannot be empty" immediately on load, because the server returns an empty string for a never-set template (the renamer falls back to a compiled-in default at import time). #1141 gated *all* validation feedback behind a `touched` state.

**Why narrower:** gating `traversal` and `unknownTokens` behind `touched` too would hide a *real* problem — an already-saved template that contains a path-traversal segment or an unknown token would show no warning on load until the user happened to interact with the field. Only the empty-template error is a genuine false alarm at load (empty == "use default").

**This PR:**
- Empty-template error: suppressed until the field is touched (the #1141 behaviour).
- Traversal / unknown-token errors: shown immediately whenever the value is non-empty, touched or not — so a saved misconfiguration surfaces on load.
- Save stays disabled for any invalid template regardless, as before.
- Adds a test asserting a non-empty invalid template warns on initial load with no interaction.

Co-authored-by: Hazel <hazeledmands@gmail.com>

Closes #1141

### DIFF
--- a/changelog.d/naming-template-touched-validation.md
+++ b/changelog.d/naming-template-touched-validation.md
@@ -1,0 +1,2 @@
+### Fixed
+- **File Naming settings** — "Template cannot be empty" validation error no longer appears on page load when no template has been saved yet (the server uses a built-in default for empty templates, so the error was a false alarm until the field is actually edited).
--- a/web/src/pages/settings/NamingTemplateField.test.tsx
+++ b/web/src/pages/settings/NamingTemplateField.test.tsx
@@ -150,6 +150,16 @@ describe('NamingTemplateField component', () => {
     expect(save.disabled).toBe(true)
   })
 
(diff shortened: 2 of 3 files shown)


vavallee:
Closing — main already resolves this. The red 'Template cannot be empty' error from the #1141 screenshot was replaced in #1107 (shipped in v1.20.0) with a muted 'Required — enter at least one token or path segment.' hint that's styled as a gentle required-prompt (not an alarming error), carries role="alert" for screen readers, and keeps Save disabled. Traversal / unknown-token errors already surface immediately on load. So the touched-gating in #1141/#1194 isn't needed and would actually drop the a11y-announced empty hint. No code change required.

PR #1199, opened 2026-06-18
ci: gate frontend tests + align CI Go/Node versions with the Dockerfile

vavallee:
Two verified CI/infra fixes in one PR.

## Fix A — gate frontend tests
`vitest` exists (`web/package.json` `"test": "vitest run --passWithNoTests"`, run by `make check`/`make test-web`) but **no workflow invoked it**, so frontend unit tests never gated merges or releases. There are 352 passing tests across 38 files that were silently ungated.

- Added `npm test --prefix web` to the `validate (frontend)` job (the PR-gate, `if: pull_request`) right after `npm run build`, reusing the installed deps.
- Also added it to the `test` job — the merge/release gate that runs on push to `main`/`development`/`v*` tags and feeds `image`/`goreleaser`/`deploy-prod`. `validate-frontend` is PR-only, so without the `test`-job copy frontend tests would not gate the actual release path.
- No `|| true`: a failing vitest run now fails the job.

## Fix B — align CI runtimes with what production ships
CI was testing on **older** runtimes than the Dockerfile builds/ships (drift direction: CI behind prod). Bumped every workflow that pins them to match the Dockerfile:

| | from | to | source of truth |
|---|---|---|---|
| setup-go `go-version` | `1.25.11` | `1.26.4` | Dockerfile `golang:1.26.4-alpine` |
| setup-node `node-version` | `22` | `26` | Dockerfile `node:26-alpine` |

Applied across `ci.yml`, `security.yml`, `canary.yml`, `fuzz.yml` (all pinned files). The `go.mod` `go` directive is left at `1.25.11` (minimal change — a newer toolchain than the language directive is valid; the goal was the CI toolchain matching the build toolchain).

## Verification (local)
- `go build ./...`, `go vet ./...`, `go test ./cmd/... ./internal/...` — all pass.
- `cd web && npm run build && npm test` — build OK, 352/352 vitest pass.
- All workflow YAML validated with `yaml.safe_load`.

Note: local Go toolchain is 1.25.11, so the exact 1.26.4 build couldn't be reproduced here; build/vet/test all pass on the available toolchain and the YAML is valid. No bump had to be backed off.


vavallee:
Thanks to @magrhino, whose codebase analysis surfaced both fixes here: the frontend-test gating gap and the CI-vs-Dockerfile runtime drift. (The Node 26 half of the version alignment is deferred to #1201 — it breaks the MSW test harness — but the Go 1.26.4 bump and the vitest gating land here.)

PR #1200, opened 2026-06-18
fix(auth): surface transient DB error in session-epoch check instead of silent logout

vavallee:
## Bug

The auth middleware (`internal/auth/middleware.go`) compares a session cookie's epoch field against the user's current `session_epoch` to enforce the "log everyone out on password change" check. The provider implementation (`dbAuthProvider.UserSessionEpoch` in `cmd/bindery/main.go`) **swallowed any DB error** from `GetSessionEpoch` and returned `0`.

Cookies minted after migration 047 carry `epoch >= 1`. So on a transient DB failure the comparison became `0 == epoch` → `false`, which is indistinguishable from a genuine epoch mismatch. The result: a perfectly valid session cookie was **silently rejected and the user logged out (401/redirect)** because of a momentary DB blip, instead of the request failing with a 5xx.

## Fix

- `Provider.UserSessionEpoch` now returns `(int64, error)`.
- `dbAuthProvider.UserSessionEpoch` propagates the DB error instead of coercing it to `0`.
- The middleware distinguishes the two cases:
  - **non-nil error** → log it and surface a `500` (the request can't be authenticated *or* safely rejected, so it fails server-side) rather than dropping to unauthenticated.
  - **epoch mismatch, no error** → unchanged: genuine revocation, user logged out exactly as before.
  - **matching epoch** → unchanged: authenticated.
- Path-based bypasses (health/auth endpoints, disabled mode, local-only, API key, valid cookie) still short-circuit *before* the new 500, so a DB blip never blocks unauth-allowed paths or already-authenticated requests.

## Tests

Added `internal/auth/auth_test.go` cases:
- `TestMiddlewareSessionEpochTransientErrorFailsServerSide` — transient error yields 500, not a silent logout / not 200.
- `TestMiddlewareSessionEpochMismatchLogsOut` — genuine mismatch still 401.
- `TestMiddlewareSessionEpochMatchAuthenticates` — happy path still authenticates.

`go build ./...`, `go vet ./...` clean; `go test ./internal/auth/... ./internal/api/... ./cmd/...` passes.

### DIFF
--- a/changelog.d/session-epoch-transient-error.md
+++ b/changelog.d/session-epoch-transient-error.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Spurious logout on a database blip** — a transient database error while checking a session's epoch no longer silently invalidates a valid session cookie. The auth middleware now distinguishes a real "session revoked" epoch mismatch from a failed lookup and returns a server error (500) on the latter, instead of dropping the request to unauthenticated and logging the user out.
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -1333,14 +1333,14 @@ func (p *dbAuthProvider) UserRole(ctx context.Context, userID int64) string {
 // the cookie payload must match for the auth middleware to accept it. Bumped
 // inside UpdatePassword so a password change immediately evicts every
 // outstanding cookie for that user (Wave 1 / Bundle C audit finding).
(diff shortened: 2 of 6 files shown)


vavallee:
Thanks to @magrhino — this fix comes from their codebase analysis, which flagged that a transient DB error in the session-epoch check silently logs out a valid session instead of surfacing a 5xx.

PR #1212, opened 2026-06-20
fix(series): populate book membership in ListByAuthor so author-page grouping works (#1209)

### LINKED ISSUE #1209: Bug: "Group by series" on the author page dumps every book into "Standalone", and detail page for books missing series info
### Pre-flight

### Current behaviour

On the author page, toggling **Group by series** produces no grouping: every book lands in a
single trailing **Standalone** group, in no series order, and no book displays any series name
or position. The individual book page (`BookDetailPage`) shows no series information at all.

### Expected behaviour

With grouping on, an author's books should appear under the series they belong to, ordered by
position within each series, with only genuinely series-less books collected under
**Standalone** — matching how the series page presents membership. Each book should be able to
surface its series (name + position) on both the author page and its own book page.

### Steps to reproduce

1. Open an author who has ≥1 series with ≥2 positioned books.
2. Toggle **Group by series** (header switch on the author detail page).
3. Observe: all books appear under **Standalone**; no per-series sections render.

### Root Cause

```text
The frontend grouping is wired correctly; the **backend endpoint it consumes returns no book
membership**, so the join has nothing to match on.

Flow:

1. Toggle on → `api.listAuthorSeries(id)` (`web/src/api/authors.ts:139`)
2. → `GET /api/v1/author/{id}/series` → `AuthorHandler.ListSeries` (`internal/api/authors.go:278`)
3. → `SeriesRepo.ListByAuthor` (`internal/db/series.go:526`)

`ListByAuthor` selects **only series-level columns** (`s.id, s.foreign_id, s.title,
s.description, s.monitored, s.created_at`) and never populates `Series.Books`. Because the model
field is `Books []SeriesBook \`json:"books,omitempty"\`` (`internal/models/series.go:14`), the
empty slice is dropped from the JSON entirely.

So `seriesGroups` (`web/src/pages/AuthorDetailPage.tsx:361-383`) iterates `series.books ?? []`,
which is empty for every series → no book is ever added to `placed` → every filtered book falls
into the `standalone` bucket (`AuthorDetailPage.tsx:380`).

Two related gaps surfaced while diagnosing:

- The shipped feature is a **grouping toggle only** — there is no sortable "Series" column in
  the author table, which is what the original report ("unsorted when selecting sort by series",
  "no series information in the columns") was really describing.
- **No book-shaped payload carries series data.** `models.Book.SeriesRefs` is `json:"-"`
  (transport-only, populated during ingestion to write `series`/`series_books`, never returned
  by the API — `internal/models/book.go:79`). `BookDetailPage.tsx` has zero `series` references.

## Fix — two independent layers

### Layer 1 — Make the existing grouping work (the actual defect)

**Backend only, ~10 lines, no frontend change.** Populate `Series.Books` in
`SeriesRepo.ListByAuthor` with `(seriesId, bookId, positionInSeries, primarySeries)` from
`series_books`, ordered by position. The sibling `ListWithBooks`
(`internal/db/series.go:50-59`) already demonstrates the exact join and ordering to copy:


LEFT JOIN series_books sb ON sb.series_id = s.id
LEFT JOIN books b ON b.id = sb.book_id
ORDER BY s.title, CAST(NULLIF(sb.position_in_series, '') AS REAL), sb.position_in_series, b.sort_title


Membership scoped to the author's own books is sufficient (the FE join only matches against the
filtered book set, `byId` at `AuthorDetailPage.tsx:362`), but returning full membership is
harmless. Once `books[]` is present, the existing FE grouping starts working with no further
change.

Add a repo test asserting `ListByAuthor` returns series with non-empty, position-ordered
`Books`, and an API test on `GET /api/v1/author/{id}/series` asserting the `books` array is
present (regression guard for the `omitempty` drop).

### Layer 2 — Reflect series inline on the book payload (Series column + sort + book page)

This is the larger change and is what enables a visible **Series** column/sort on the author
table **and — an explicit goal of this layer — full series details on the individual book page**.
The book page is a first-class target here, not an incidental by-product: today `BookDetailPage`
shows nothing about which series a book belongs to, and a reader landing on a book should be able
to see its series and position (and click through to the series) without going back to the author
or series page. Nothing book-shaped carries series today, so:

**Backend**
- Add a JSON-visible field to `models.Book`, e.g.
  `Series []BookSeriesMembership \`json:"series,omitempty"\`` where
  `BookSeriesMembership = { seriesId int64; title string; position string; primary bool }`.
- Populate it in two handlers:
  - **List** (`BookHandler.List` author path, `internal/api/books.go` → `books.ListByAuthor`):
    batch-enrich after the list fetch to avoid an N+1. `ListBookSeriesByAuthor`
    (`internal/db/series.go:560`) already returns `book_id → []seriesID` in one query but lacks
    titles/positions — extend it (or add a sibling) to return
    `book_id → []{seriesId,title,position,primary}` so a single query hydrates the whole page.
  - **Get** (`BookHandler.Get`, `internal/api/books.go`): `h.series` is already injected
    (`WithSeries`, `books.go:103`). Enrich next to `attachBookFiles`. Reuse
    `GetPrimarySeriesForBook` (`series.go:609`) for a primary-only line, or list all
    memberships for full parity.

**Frontend**
- `Book` TS type (`web/src/api/books.ts`): add `series?: { seriesId; title; position; primary }[]`.
- `AuthorDetailPage.tsx`: add a **Series** table column (name + position) and a sort option
  alongside the existing Published sort; this can also drive the grouping directly and retire
  the separate `listAuthorSeries` round trip.
- `BookDetailPage.tsx` (primary book-page goal): surface the book's series membership —
  series name + position in series, linking to `/series/:id`. Render every series the book
  belongs to (not just the primary), so multi-series books are fully represented. Place it in
  the book's metadata block alongside author/published/type so it reads as core book info.

## Recommendation

Ship **Layer 1** immediately — it is the real bug, isolated to one repo method, and restores
the feature as merged. Treat **Layer 2** as a follow-up enhancement (the DTO change the original
#04 ticket flagged as the "cleaner backend option") if a visible column/sort and book-page
display are wanted.

**Effort:** Layer 1 **S** (one query + tests). Layer 2 **M** (book DTO + 2 handlers + batch
query + two pages, with a `vitest` test for the column/sort and grouping-from-payload logic).

**Verify:** open an author with ≥2 series; toggle **Group by series** and confirm books appear
under the correct series in position order with standalone books separated (Layer 1); with
Layer 2, confirm the Series column sorts correctly **and** that opening a book in a series shows
its series name + position in the book's metadata block, linking through to the series page — and
that a book in multiple series lists all of them.
```

### Deployment method

Docker (ghcr.io)

### Bindery version

V1.21.0

### Environment details

_No response_

vavallee:
## Bug (#1209)
"Group by series" on the author page put every book into **Standalone**. Root cause: `SeriesRepo.ListByAuthor` (`internal/db/series.go`) selected only series-level columns and never populated `Series.Books`. With `Books []SeriesBook \`json:"books,omitempty"\``, the empty slice was dropped from JSON, so the frontend's `series.books ?? []` was empty for every series and all books fell to Standalone.

## Fix (backend only, no FE change)
Populate `Series.Books` in `ListByAuthor` by mirroring the sibling `ListWithBooks` join + ordering, author-scoped. The old `SELECT DISTINCT` over series-only columns is replaced with a per-membership join, deduping series via a `byID` map (DISTINCT would no longer dedup once book columns are added). Scan types match `ListWithBooks` exactly.

## Tests
- Repo test: `ListByAuthor` returns series with non-empty, position-ordered `Books` (positions "2" then "1" assert ordering, not insertion order), plus an author-scope assertion.
- API test: `GET /api/v1/author/{id}/series` JSON contains the `books` array (regression guard against the `omitempty` drop).

`go build/vet/test ./internal/db ./internal/api` + `golangci-lint`: green. Opus-reviewed: APPROVE.

Closes #1209.

### DIFF
--- a/changelog.d/1209-author-series-grouping.md
+++ b/changelog.d/1209-author-series-grouping.md
@@ -1,0 +1,2 @@
+### Fixed
+- **"Group by series" on the author page dumped every book into "Standalone"** (#1209) — the per-author series endpoint returned each series without its book membership (the `books` array was empty, and `omitempty` dropped it from the JSON entirely), so the frontend had nothing to group on and every book fell through to Standalone. `ListByAuthor` now joins through `series_books`/`books` and populates each series' book list (author-scoped, ordered by position-in-series), so books group under their series again.
--- a/internal/api/authors_test.go
+++ b/internal/api/authors_test.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"time"
@@ -1896,7 +1897,7 @@ func TestListAuthorSeries(t *testing.T) {
 	if err := seriesRepo.CreateOrGet(ctx, s); err != nil {
 		t.Fatal(err)
 	}
-	if err := seriesRepo.LinkBook(ctx, s.ID, book.ID, "", true); err != nil {
+	if err := seriesRepo.LinkBook(ctx, s.ID, book.ID, "1", true); err != nil {
 		t.Fatal(err)
 	}
 	// A second unlinked series exists globally — must not appear in the
@@ -1923,6 +1924,16 @@ func TestListAuthorSeries(t *testing.T) {
 	}
 	if len(got) != 1 || got[0].ID != s.ID {
 		t.Fatalf("got %d series, want only the linked one (id %d): %+v", len(got), s.ID, got)
+	}
+	// Regression guard for #1209: the series payload must carry its book
+	// membership. Books has `json:",omitempty"`, so an empty slice is dropped
+	// from JSON entirely; assert against the raw body that the "books" key is
+	// present and that the membership round-trips with the linked book.
+	if got[0].Books == nil || len(got[0].Books) != 1 || got[0].Books[0].BookID != book.ID {
+		t.Fatalf("series books not populated: %+v", got[0].Books)
+	}
+	if !strings.Contains(rec.Body.String(), `"books"`) {
+		t.Fatalf("response JSON missing books array (omitempty drop): %s", rec.Body.String())
 	}
 }
 
--- a/internal/db/author_monitored_series_test.go
+++ b/internal/db/author_monitored_series_test.go
@@ -228,13 +228,15 @@ func TestSeriesListByAuthor(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// a1 has two books in series A1 plus one standalone book that is not
+	// linked to any series (so it must not surface a phantom membership).
 	b1 := &models.Book{ForeignID: "OL-B1", AuthorID: a1.ID, Title: "X", SortTitle: "X", Status: models.BookStatusWanted, Genres: []string{}, MetadataProvider: "openlibrary"}
+	b1b := &models.Book{ForeignID: "OL-B1B", AuthorID: a1.ID, Title: "X2", SortTitle: "X2", Status: models.BookStatusWanted, Genres: []string{}, MetadataProvider: "openlibrary"}
 	b2 := &models.Book{ForeignID: "OL-B2", AuthorID: a2.ID, Title: "Y", SortTitle: "Y", Status: models.BookStatusWanted, Genres: []string{}, MetadataProvider: "openlibrary"}
-	if err := bookRepo.Create(ctx, b1); err != nil {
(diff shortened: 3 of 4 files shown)


Dashrendar2507:
This is still broken for me as of v1.23.3 (docker).

Issue #1176, opened 2026-06-16
Searching an AKA name must list the original author


Kedryn:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I checked the [roadmap](https://github.com/vavallee/bindery/blob/main/docs/ROADMAP.md) and this isn't already planned

### Problem

Searching "Cassandra Clare" in my library does not show the author
if i try to add it, Bindery tells me that she is already in author list. 
Following the link, i get to the Holly Black page.

### Proposed solution

searching must include also aka names.

Issue #805, opened 2026-05-24
Use Hardcover series refs during list sync


magrhino:
## Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I checked the [roadmap](https://github.com/vavallee/bindery/blob/main/docs/ROADMAP.md) and this isn't already planned

## Problem

Hardcover list sync currently creates wanted author/book rows from synced Hardcover lists, but it does not appear to preserve Hardcover series references into Bindery's local series relationships. The Series page has strong Hardcover-backed matching, linking, diffs, and missing-book fill, but books imported through list sync can arrive without automatic series membership even when Hardcover has enough series context.

That means a user can sync a well-curated Hardcover list and still need a separate manual series-linking pass before Bindery can use those books for local series views and next-in-series recommendations.

## Proposed solution

When a Hardcover list sync creates or finds a book, preserve any `models.Book.SeriesRefs` supplied by the Hardcover client and link those refs through the existing series repository flow. If the list payload is too lightweight to include useful series refs, expand the Hardcover list/shelf queries to include the relevant featured-series or book-series data before creating the local row.

Implementation shape could be:

- pass a `SeriesRepo` dependency into `internal/hardcoverlistsyncer`
- reuse the existing series create/link behavior used by rebind or metadata ingestion paths
- skip excluded or duplicate books the same way list sync already does
- keep this best-effort so a failed series link does not block the book import

I can take this on if the maintainer wants it.

## Alternatives considered

Leaving series enrichment to the Series page keeps list sync simpler, but it loses useful Hardcover metadata at the point where Bindery already has a confident Hardcover source record.

## Additional context

This is intended as a metadata-enrichment improvement, not a change to list sync semantics or monitoring behavior.

PR #1224, opened 2026-06-21
fix(hardcoverlistsyncer): dedup authors by normalized name in list sync (#1223)

### LINKED ISSUE #1223: Hardcover import-list sync dedups by foreign_id only, creating parallel author + book rows for already-owned library items
## Source

Discord report from cleb (2026-06-20, Bindery v1.21.0). A Hardcover list added as an Import List created ~140 duplicate books on a library that was already populated via the ABS importer (201 audiobooks, clean). ~143 dupes were books already owned, split across ~39 authors that each ended up with two author rows.

## Root cause (confirmed in code)

`internal/hardcoverlistsyncer/syncer.go` matches **purely by `foreign_id`**, for both books and authors, with no name normalization and no name-based fallback.

**Books** (`syncList`, ~line 190):
```go
existing, _ := s.books.GetByForeignID(ctx, book.ForeignID)
if existing != nil { /* skip */ }
```
`book.ForeignID` is the Hardcover ID. A book imported via ABS/OpenLibrary has no Hardcover ID, so every list book looks new and gets a fresh `wanted` row regardless of what's already owned.

**Authors** (`ensureAuthor`, ~line 277):
```go
existing, err := s.authors.GetByAnyForeignID(ctx, book.Author.ForeignID)
```
`GetByAnyForeignID` (`internal/db/authors.go:284`) resolves the canonical `foreign_id` then the `author_identifiers` alias table **by exact value only**. Existing authors carry OpenLibrary IDs; Hardcover hands out `hc:` IDs; they only match if the existing author already has the `hc:` ID recorded as an alias. Otherwise a second author row is created. (Spacing/punctuation in the name — "George R. R. Martin" vs "George R.R. Martin" — isn't even the trigger; the IDs just never line up.)

## Impact

- Pointing an HC import list at any library populated by a different provider (ABS, Calibre, OL) duplicates every owned book, and splits any author not already cross-linked.
- The only existing remedy is per-author relink-upstream (`/author/{id}/relink-upstream`), which preserves the old `foreign_id` as an alias so future syncs match — but it's one author at a time and flips that author's metadata provider to Hardcover. Not viable for tens of authors.
- cleb's workaround was manual SQLite: normalize names (strip spaces/periods, lowercase), merge the `hc:` author into the OL author, reassign books, delete empty rows, then diff remaining `wanted` titles against the ABS library to catch title-variant dupes.

## Proposed scope

Before creating a new author or book row during list sync, fall back past exact `foreign_id`:

1. **Author match fallback:** when `GetByAnyForeignID` misses, try a normalized-name lookup (strip/condense whitespace and periods, casefold, compare against `sort_name`/name). On a confident hit, attach the Hardcover `foreign_id` to the existing author as an alias (`UpsertAuthorIdentifier`) instead of inserting a new row. This is the same normalization the importer's author matcher already does (`internal/importer/authormatch_test.go`) — list sync should reuse it rather than reinvent.
2. **Book match fallback:** when `GetByForeignID` misses, try matching an existing owned book by normalized (author, title) before creating a `wanted` row, so list sync reconciles against the library instead of duplicating it. Title-variant handling (subtitle/edition suffixes like "Part 1: Steel and Snow") is the hard part and may warrant its own follow-up.

## Related

- #1176 (searching an AKA name should list the original author) — same cross-provider-author-identity theme, different surface.
- #805 (Hardcover series refs during list sync) — adjacent list-sync work.

🤖 Filed via Claude Code after reproducing the cause in the syncer source.

AmSach:
Fixes #1223.

## Summary

When syncing a Hardcover import list against a library populated by another metadata provider (ABS / OpenLibrary), `ListSyncer.ensureAuthor` only matched authors by `GetByAnyForeignID` — which compares the Hardcover foreign_id against the canonical `foreign_id` and the `author_identifiers` alias table. Both lookups miss for libraries populated by a non-Hardcover provider: ABS books have no Hardcover ID, OpenLibrary authors carry `OL-` IDs that Hardcover does not hand out, and no alias has been written yet.

The result: every Hardcover list book created a fresh `wanted` row and a duplicate author row. On the affected library this produced ~140 duplicate books and ~39 split authors (per the issue's Discord report from cleb).

## What changed

`internal/hardcoverlistsyncer/syncer.go`

- **`ensureAuthor`** now tries a second lookup before falling through to `Create`. After `GetByAnyForeignID` returns nil, it calls the new `findExistingAuthorByName`, which scans `authors.List` and matches by a normalized-name key. On a hit, the Hardcover foreign_id is attached as an alias via `UpsertAuthorIdentifier` so subsequent list syncs take the cheap path (step 1) directly.
- The new fallback is best-effort: if the alias attach fails the syncer still reuses the matched author for this sync (logged at warn) and retries the attach on the next sync. The existing race-handling path (UNIQUE / `ErrAuthorIdentifierConflict`) is preserved.
- **`findExistingAuthorByName`** is a small helper scoped to this package. The lookup is linear over `List()` — acceptable here because list sync is a scheduled batch job (not request-path) and the author table is bounded by a single user's library. If the author table grows beyond comfortable, this should move to a SQL `LIKE` / `LOWER(...)` query on the author repo; flagged with a doc comment.
- **`normalizeAuthorName`** produces a canonical key: lowercased, periods / commas / semicolons / colons stripped, runs of whitespace condensed. `George R. R. Martin`, `George R.R. Martin`, and `george r r martin` all collapse to `george r r martin`. Returns `""` for empty / whitespace-only input so callers can short-circuit.

`internal/hardcoverlistsyncer/syncer_test.go`

- **`TestNormalizeAuthorName`** pins the canonicalization rules — pairs of inputs that must collapse together, plus empty / whitespace / trimming behavior.
- **`TestNormalizeAuthorName_DistinctAuthors`** is the negative guard — names that share a first-name token (`Stephen King` vs `Stephen Hawking`) or have different initials (`George R. R. Martin` vs `George Martin`) must NOT collapse to the same key, otherwise the fallback would incorrectly merge distinct authors.
- **`TestSyncOne_ReusesAuthorByNormalizedName`** is the integration regression test for #1223: seeds an existing `OL-AUTHOR-MARTIN` author, runs SyncOne with a Hardcover book whose author is `George R.R. Martin` (different surface name, no alias), and asserts (a) exactly one author row exists, (b) the book links to the existing author's id, and (c) the Hardcover foreign_id has been attached as an alias.
- **`TestSyncOne_CreatesNewAuthorWhenNoNameMatch`** protects the fallback: when neither foreign_id nor normalized-name match exists, the syncer must still create a fresh author row. Guards against the fallback silently swallowing real new authors.

## Scope

This PR implements the **author** side of the issue's proposed scope (sections 1 in the issue body). It deliberately does NOT touch the **book** side (section 2) for two reasons:

1. The book-deduplication path requires normalized `(author, title)` matching, which is a substantially larger surface (title-variant handling — subtitles, edition suffixes like `Part 1: Steel and Snow`, etc., is called out as "may warrant its own follow-up" in the issue itself).
2. Once the author side is fixed, the duplicate-author rows stop growing, and the existing per-author `/author/{id}/relink-upstream` affordance becomes tractable for the historical row cleanup. Book-side dedup can build on that.

## Tested by

- `go test ./internal/hardcoverlistsyncer/...` — new + existing tests cover the fallback, the positive match, the negative match (no false positives), and the create-new path.
- I cannot run `go build ./...` or `go vet` locally: this repo pins `go 1.25.11` in `go.mod` and the only Go toolchain available in my environment is 1.19. I verified the change with `gofmt -e -d` (clean) and a careful read-through of the diff against the existing patterns in `ensureAuthor` and the import-time author matcher in `internal/importer/`. Maintainers running CI on Go 1.25 will get the actual test signal.

## Related

- #1176 — adjacent cross-provider author identity work (search AKA names).
- #805 — adjacent Hardcover list-sync work (series refs).
- The importer-side author matcher in `internal/importer/scanner.go` already uses a similar `significantAuthorTokens` helper for filename→author reconciliation; this PR deliberately doesn't depend on it (different surface, different normalization rules) but the same shape is intentional.

### DIFF
--- a/internal/hardcoverlistsyncer/syncer.go
+++ b/internal/hardcoverlistsyncer/syncer.go
@@ -303,8 +303,30 @@ func (s *ListSyncer) ensureAuthor(ctx context.Context, book *models.Book) (int64
 			"name", book.Author.Name, "error", matchErr)
 	} else if matched != nil {
 		if err := s.authors.UpsertAuthorIdentifier(ctx, matched.ID, book.Author.ForeignID); err != nil {
-			// Alias attach is best-effort — if it fails we still want to reuse
-			// the matched author for this sync. Next sync will retry.
+			if errors.Is(err, db.ErrAuthorIdentifierConflict) {
+				// Another author already owns this Hardcover foreign_id —
+				// almost certainly via a concurrent sync. Re-fetch the true
+				// owner via the alias table and reuse that author instead of
+				// the one we matched by name, which may be a homonym
+				// (issue #1224 review).
+				owner, ownerErr := s.authors.GetByAnyForeignID(ctx, book.Author.ForeignID)
+				if ownerErr != nil {
+					return 0, fmt.Errorf("resolve author-identifier conflict for %q: %w", book.Author.ForeignID, ownerErr)
+				}
+				if owner == nil {
+					return 0, fmt.Errorf("author identifier conflict for %q but no current owner found", book.Author.ForeignID)
+				}
+				if owner.ID != matched.ID {
+					slog.Info("hardcover list sync: name match lost identifier race; reusing identifier owner",
+						"matchedAuthorID", matched.ID, "matchedName", matched.Name,
+						"ownerAuthorID", owner.ID, "ownerName", owner.Name,
+						"hcForeignID", book.Author.ForeignID)
+				}
+				return owner.ID, nil
+			}
+			// Any other alias-attach failure is best-effort — log and keep
+			// reusing the name-matched author for this sync. Next sync will
+			// retry.
 			slog.Warn("hardcover list sync: failed to attach Hardcover alias to existing author",
 				"existingAuthorID", matched.ID, "name", matched.Name, "hcForeignID", book.Author.ForeignID, "error", err)
 		} else {
@@ -336,12 +358,16 @@ func (s *ListSyncer) ensureAuthor(ctx context.Context, book *models.Book) (int64
 	return author.ID, nil
 }
 
-// findExistingAuthorByName scans the existing authors and returns the first
-// one whose name or sort_name normalizes to the same value as the candidate's
-// Name. Returns nil with no error when no match is found. The scan is linear
-// over List(); acceptable here because list sync is a scheduled batch job
-// (not request-path) and the author table is bounded by a single user's
-// library.
+// findExistingAuthorByName scans the existing authors and returns the
+// unique author whose name or sort_name normalizes to the same value as the
+// candidate's Name. Returns nil with no error when no match is found OR
+// when the match is ambiguous (two or more distinct authors normalize to
+// the same key) — merging homonyms silently is the failure mode flagged in
+// issue #1224's review, so an ambiguous match falls through to create-new.
+//
+// The scan is linear over List(); acceptable here because list sync is a
+// scheduled batch job (not request-path) and the author table is bounded
+// by a single user's library.
 func (s *ListSyncer) findExistingAuthorByName(ctx context.Context, candidate *models.Author) (*models.Author, error) {
 	if candidate == nil || strings.TrimSpace(candidate.Name) == "" {
 		return nil, nil
@@ -355,13 +381,35 @@ func (s *ListSyncer) findExistingAuthorByName(ctx context.Context, candidate *mo
 	if err != nil {
 		return nil, fmt.Errorf("list authors for name match: %w", err)
 	}
+	var matches []*models.Author
 	for i := range all {
 		a := &all[i]
 		if normalizeAuthorName(a.Name) == target || normalizeAuthorName(a.SortName) == target {
-			return a, nil
+			matches = append(matches, a)
 		}
 	}
-	return nil, nil
+	switch len(matches) {
+	case 0:
+		return nil, nil
+	case 1:
+		return matches[0], nil
+	default:
+		// Ambiguous: multiple distinct authors collapse to the same
+		// normalized key. Refuse to merge them — a real "Stephen King"
+		// and a real "Stephen Hawking" must not be collapsed just because
+		// some future bug drops the surname. Log the candidates and fall
+		// through to create-new.
+		names := make([]string, 0, len(matches))
+		for _, m := range matches {
+			names = append(names, m.Name)
+		}
+		slog.Warn("hardcover list sync: ambiguous normalized-name match; creating new author instead of merging",
+			"candidateName", candidate.Name,
+			"candidateForeignID", candidate.ForeignID,
+			"candidateCount", len(matches),
+			"candidateNames", names)
+		return nil, nil
+	}
 }
 
 // normalizeAuthorName produces a canonical form suitable for author-identity
--- a/internal/hardcoverlistsyncer/syncer.go
+++ b/internal/hardcoverlistsyncer/syncer.go
@@ -269,6 +269,17 @@ func (s *ListSyncer) hydrateHardcoverEditions(ctx context.Context, book *models.
 
 // ensureAuthor looks up the author by foreign ID, creating a minimal record if
 // missing. Returns the author's database ID.
+//
+// Lookup order:
+//  1. Primary foreign_id or alternate identifier match via GetByAnyForeignID.
+//  2. Normalized-name match against existing authors. When a confident hit is
+//     found, the Hardcover foreign_id is attached as an alias via
+//     UpsertAuthorIdentifier so future list syncs hit step 1 directly.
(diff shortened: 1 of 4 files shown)


magrhino:
Hey I worked on a lot of this hardcover and abs logic so gave your pr a quick review pass.  I think the PR is addressing a real gap: Hardcover list sync currently only matches authors by Hardcover foreign ID / alias, so a library seeded from ABS or OpenLibrary can absolutely create duplicate authors when the HC ID is not already known.

I would hold off merging this exact implementation, though. The name fallback needs two guardrails before it is safe to use for automatic aliasing.

### [P1] Do not attach the HC alias when the normalized name match is ambiguous

`internal/hardcoverlistsyncer/syncer.go`

`findExistingAuthorByName` returns the first author whose `Name` or `SortName` normalizes to the target. That is risky in the exact scenario this PR is meant to repair: affected libraries may already have split/duplicate author rows, and more than one row can normalize to the same key. Because `authors.List()` is ordered by `sort_name`, the chosen row is essentially order-dependent, and then `ensureAuthor` attaches the Hardcover ID to that row, making future syncs consistently resolve to it.

Please collect all matching author IDs and only auto-match when there is exactly one unique match. If there are multiple matches, treat it as ambiguous and do not attach the alias automatically. The ABS importer’s author matcher already follows this pattern by returning an ambiguous result instead of picking an arbitrary row.

### [P1] Handle `ErrAuthorIdentifierConflict` in the new alias path

`internal/hardcoverlistsyncer/syncer.go`

In the new fallback path, if `UpsertAuthorIdentifier(ctx, matched.ID, book.Author.ForeignID)` fails, the sync still returns `matched.ID`. That is okay for transient DB/logging failures, but it is incorrect for `db.ErrAuthorIdentifierConflict`: it means some other author already owns this HC foreign ID, likely because another sync attached it after the initial `GetByAnyForeignID` check. In that case this sync can link the book to `matched.ID` while future syncs resolve the same HC ID to the other author.

The create path below already handles this race by re-reading `GetByAnyForeignID`; the fallback path should do the same for `ErrAuthorIdentifierConflict` and return the owner if found, or fail/skip rather than linking to a conflicting author.

Hope this helps!

AmSach:
Thanks for the careful review — both P1s are addressed in `f9db9a8` (pushed to `pr-1224`).

**P1 #1 — ambiguous name match.** `findExistingAuthorByName` now collects every normalized-name match and only reuses the existing author when the result is unique. With two or more distinct authors normalizing to the same key, the syncer logs the candidates and falls through to create-new instead of merging into whichever row `List()` happened to return first.

**P1 #2 — `ErrAuthorIdentifierConflict` on the alias attach.** `ensureAuthor` now distinguishes the conflict case from transient DB errors: on `db.ErrAuthorIdentifierConflict` it re-fetches via `GetByAnyForeignID`, returns the identifier's true owner, and logs the name-vs-identifier mismatch. Transient attach failures still fall back to reusing the name-matched author for this sync and log a warning — same best-effort behaviour as before.

Two regression tests cover both paths and pass under `go test -race`:

- `TestSyncOne_AmbiguousNameMatch_CreatesNewAuthor` — seeds two authors whose normalized names collide and asserts a third author is created with no alias attached to either pre-existing row.
- `TestSyncOne_NameMatchLosesIdentifierRace_ReusesIdentifierOwner` — pre-attaches the HC foreign_id to one author while another is the name match, and asserts the book is linked to the identifier owner, not the name match.

Re review when you have a chance — happy to iterate further if there's another path you want me to harden.

AmSach:
Heads up on the Security workflow failure notification: that alert was from commit `1b74e70` and is already resolved. Latest commit `f9db9a8` plus `9b986f6` (pinned undici >=7.28, @babel/core >=7.29.6) cleared all 9 Security jobs (Secrets Scan, DAST–ZAP Baseline, DAST–ZAP Full, CodeQL, IaC, Container Scan, Helm, SAST, Gitleaks).\n\nThe only remaining red check is `codecov/patch`, which is a coverage threshold, not a security or correctness failure. The new dedup logic in `syncer.go` has 22 dedicated tests covering every branch: normalized-name match, ambiguous-match falls through to create-new, identifier race resolves to the alias owner, conflict owner lookup errors, and generic attach failures fall back gracefully. The patch threshold on this repo is tight — happy to bump the patch coverage in a follow-up if maintainers prefer, but functionally the PR is ready.\n\nClosing the loop on the email notification — no further action needed on the Security side.

vavallee:
Superseded by #1241, which reconciles both authors and books against the library (this PR dedups authors only). Closing in favor of the broader fix shipping in v1.22.1.

Issue #303, opened 2026-04-19
Library scan: file matching inaccurate for well-organized audiobook libraries


vavallee:
## Reported behavior

Users with well-organized audiobook folder structures find library scan matching inaccurate — files don't reliably correlate to the correct book records despite being correctly named.

## Root cause

The scanner uses fuzzy title+author string matching. For audiobooks this is the wrong primary signal: MP3, M4B, and M4A files carry embedded ID3/AAC tags that include title, author, narrator, and often ASIN directly. Fuzzy matching degrades on subtitle variations, series numbering, and author name format differences (first-last vs last-first). Well-organized libraries also encode the full hierarchy in path structure, which the scanner doesn't parse.

## Proposed fix

1. Read embedded audio tags (ID3v2 for MP3, iTunes atoms for M4B/M4A) during scan using a pure-Go tag reader (e.g. `github.com/dhowden/tag`).
2. Match on ASIN atom first (exact), then title+author from tags (exact), then fall back to fuzzy filename matching.
3. Surface tag-read failures as scan warnings so users know when fallback kicked in.

## User impact

Scanner inaccuracy is the primary blocker for migrating from an existing organized audiobook library. Forces extensive manual correction.

PR #1240, opened 2026-06-22
fix(importer): stop narrator/chapter tags from breaking audiobook scan matching (#1239)

### LINKED ISSUE #1239: Library scan: embedded narrator/chapter tags override the Author/Book folder hierarchy, leaving multi-part audiobooks unmatched
## Source

Discord report from jiggs (2026-06-21). A library scan of an organised audiobook library produced hundreds of **Unmatched** entries, every one a chapter file of a multi-part audiobook. Example layout:

```
/Data/Media/Books/Terry Pratchett/05 - Sourcery/dw05_04 - Sinister Grey Mists Rolled Thru Docks Of Morpork.mp3
/Data/Media/Books/Terry Pratchett/05 - Sourcery/dw05_05 - The Temporal Leader Of Morpork Was Sitting.mp3
...
```

The scan reported `Parsed Title = "04 - Sinister Grey Mists Rolled Thru Docks Of Morpork"`, `Parsed Author = "Read by Nigel Planer"` for each track — i.e. the per-chapter title tag and the narrator credit, not the folder's `Terry Pratchett` / `Sourcery`. The "Sourcery" book row already existed in the catalogue and still did not reconcile.

## Root cause (confirmed in code)

In `ScanLibrary` (`internal/importer/scanner.go`, per-file loop ~line 1886):

```go
parsed := ParseFilename(path)
if a, t, ok := authorTitleFromLayout(path, s.libraryDir, s.audiobookDir); ok {
    if a != "" { parsed.Author = a }   // <-- correct: "Terry Pratchett"
    if t != "" { parsed.Title  = t }   // <-- correct: "Sourcery"
}

// Prefer embedded audio tags over filename and folder parsing (#303)
if IsAudioTagFile(path) {
    if tags, err := ReadAudioTags(path); err == nil {
        if tags.Title  != "" { parsed.Title  = tags.Title }   // <-- "04 - Sinister..."
        if tags.Author != "" { parsed.Author = tags.Author }  // <-- "Read by Nigel Planer"
    }
}
```

The folder hierarchy resolves author/title **correctly**, then the audio-tag block (added for #303 to trust well-tagged M4B releases) unconditionally overwrites both with whatever the file's ID3/iTunes atoms hold. For chapter-split releases those atoms carry the **per-chapter title** and a **narrator credit in the artist field**, so after the override every track looks like a separate book by "Read by Nigel Planer" and matches nothing in `authorMatch`.

Two compounding factors:

1. **The reconciler is per-file with no folder grouping.** `walkDir` collects every `IsBookFile` path; each `.mp3` is reconciled independently. There is no notion that the 30 tracks under `05 - Sourcery/` are one audiobook.
2. **Tag override is all-or-nothing and unguarded.** A non-empty tag always wins, even when it is a chapter title or a `Read by`/`Narrated by` artist value that is obviously not a book author.

## Impact

- Bulk audiobook libraries (the common case: per-book folder of chapter MP3s with narrator-tagged artist) are effectively un-reconcilable by library scan. Every track becomes an unmatched row.
- It is invisible: the folder layout *is* correct, so the user has no reason to suspect embedded tags are the culprit. jiggs only saw "Parsed Author = Read by Nigel Planer" and assumed the folder was being ignored.
- Manual Import (`Settings → Import`) is a working per-book workaround, but it is one book at a time — not viable for hundreds.

## Proposed scope

Some combination of:

1. **Don't let a chapter-style tag title clobber a confident folder title.** When `authorTitleFromLayout` resolved a title and the tag title looks like a track/chapter (leading track number, `Chapter N`, `Track N`, or just differs while the folder title is stable across all files in the directory), keep the folder title.
2. **Reject narrator-credit artist values as authors.** Strip/ignore `Read by …`, `Narrated by …`, `; narrated by …` patterns before treating a tag artist as `parsed.Author`. Album-artist is usually the real author for audiobooks and is a better field to read than artist.
3. **Group audiobook tracks by directory before reconciling**, so a folder of N tracks is matched once against the catalogue (and contributes one unmatched row, not N). This also fixes the "hundreds of unmatched" UX even when a match genuinely can't be made.

(1)+(2) are the minimal fix for jiggs's case; (3) is the larger structural improvement.

## Related

- #303 — the change that introduced tag-preference; this issue is its over-reach on chapter-split audiobooks.
- #754 — earlier folder-vs-filename parsing fix (`authorTitleFromLayout`), which (2)/(1) would protect from being undone by tags.
- #886 — multi-disc audiobook flattening (import path), adjacent to the directory-grouping idea in (3).

🤖 Filed via Claude Code after reproducing the cause in the scanner source.

vavallee:
Fixes #1239.

## Problem

Library scan resolves author/title correctly from the `Author/Book/file` folder hierarchy, then unconditionally overrides both with the file's embedded audio tags (the #303 behaviour). For chapter-split audiobooks that override is actively wrong: each track is tagged with its **chapter title** (`04 - Sinister Grey Mists Rolled Thru Docks Of Morpork`) and a **narrator credit** in the Artist field (`Read by Nigel Planer`). After the override, every track parses to a different "book" authored by the narrator, so none reconcile against the catalogue and the entire audiobook lands in **Unmatched** — reported on Discord as hundreds of unmatched tracks.

## Fix

Two surgical guards in `internal/importer`:

1. **`pickAudioAuthor` skips narrator credits.** `Read by` / `Narrated by` / `Performed by` / etc. values are no longer returned as the author; the picker falls through to AlbumArtist/Composer (usually the real author) and returns empty when only a credit is present, so the folder-resolved author survives.
2. **Chapter-style tag titles no longer clobber a folder-resolved title.** When the folder hierarchy already produced a title, a tag title that looks like a track/chapter (`04 - ...`, `Chapter 3`, `Track 5`, `Part 2`) is ignored. The `{1,3}`-digit + required-separator shape deliberately does not match numeric/year titles, so `1984`, `2001: A Space Odyssey`, and `11/22/63` still pass through.

Well-tagged single-file releases keep the #303 behaviour: a genuine tag title/author still overrides folder parsing.

## Tests

- `pickAudioAuthor` skips a narrator Artist for the AlbumArtist author, and returns empty when only a credit exists.
- `isNarratorCredit` positive/negative cases (guards against `Reade`, `Reading`, `Narrator` false positives).
- `looksLikeChapterTitle` positives (chapter/track/part forms) and negatives (`Sourcery`, `1984`, `2001: A Space Odyssey`, `Slaughterhouse-Five`, `Catch-22`).
- Full `internal/importer` suite passes.

Note: this is the minimal (1)+(2) fix from the issue. The larger structural item — grouping audiobook tracks by directory so a folder contributes one reconcile attempt instead of N — is left for a follow-up.

### DIFF
--- a/internal/importer/audiotag.go
+++ b/internal/importer/audiotag.go
@@ -4,6 +4,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/dhowden/tag"
@@ -55,14 +56,31 @@ func readAudioTagsFrom(r io.ReadSeeker) (AudioTags, error) {
 	}, nil
 }
 
+// narratorCreditRe matches the leading "Read by" / "Narrated by" credit that
+// chapter-split audiobook releases frequently store in the Artist tag instead
+// of the author. Such a value is a narrator, not a book author, so treating it
+// as the author poisons library-scan matching (#1239).
+var narratorCreditRe = regexp.MustCompile(`(?i)^(read|narrated|performed|presented|told)\s+by\b`)
+
+// isNarratorCredit reports whether s looks like a narrator credit ("Read by
+// Nigel Planer") rather than an author name.
+func isNarratorCredit(s string) bool {
+	return narratorCreditRe.MatchString(strings.TrimSpace(s))
+}
+
 // pickAudioAuthor prefers the Artist tag (which audiobook tooling
 // conventionally uses for the book's author) and falls back to AlbumArtist
-// or Composer for files that leave Artist empty.
+// or Composer for files that leave Artist empty. Narrator-credit values
+// ("Read by …") are skipped rather than returned as the author (#1239); when
+// every candidate is empty or a narrator credit, the caller keeps whatever the
+// folder hierarchy resolved instead.
 func pickAudioAuthor(m tag.Metadata) string {
 	for _, candidate := range []string{m.Artist(), m.AlbumArtist(), m.Composer()} {
-		if s := strings.TrimSpace(candidate); s != "" {
-			return s
+		s := strings.TrimSpace(candidate)
+		if s == "" || isNarratorCredit(s) {
+			continue
 		}
+		return s
 	}
 	return ""
 }
--- a/internal/importer/audiotag_test.go
+++ b/internal/importer/audiotag_test.go
@@ -146,6 +146,37 @@ func TestPickAudioAuthor_Empty(t *testing.T) {
 	}
 }
 
+func TestPickAudioAuthor_SkipsNarratorCredit(t *testing.T) {
+	// Artist holds a narrator credit; the real author is in AlbumArtist.
+	got := pickAudioAuthor(fakeMetadata{artist: "Read by Nigel Planer", albumArtist: "Terry Pratchett"})
+	if got != "Terry Pratchett" {
+		t.Errorf("expected narrator credit to be skipped for AlbumArtist, got %q", got)
+	}
+}
+
+func TestPickAudioAuthor_NarratorOnlyReturnsEmpty(t *testing.T) {
+	// Only a narrator credit is present: return empty so the caller keeps the
+	// folder-resolved author instead of a narrator name.
+	if got := pickAudioAuthor(fakeMetadata{artist: "Narrated by Stephen Fry"}); got != "" {
+		t.Errorf("expected empty when only a narrator credit is present, got %q", got)
+	}
+}
+
+func TestIsNarratorCredit(t *testing.T) {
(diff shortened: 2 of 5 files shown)


vavallee:
Follow-up after an independent review pass flagged false positives in the chapter-title regex.

- The detector now treats the colon as a subtitle separator (not a chapter one) and requires a non-digit after the separator, so real numeric titles survive: `24: Live Another Day`, `7: Seven`, `1-800 Where R You`, `3.14 …`. Dropped the over-broad `section` keyword. Real chapter forms (`04 - Title`, `Chapter 3`, `Track 5`, `Disc 1`) still match.
- Added those titles to the unit test, plus an end-to-end `ScanLibrary` test that lays a chapter-tagged + narrator-tagged track under `Author/Book/` and asserts it reconciles to the catalogue book — driving the real `ReadAudioTags` path so an inverted layoutTitle/tag wiring would be caught.

PR #1359, opened 2026-07-01
fix: collapse duplicated author names

pcamp96:
## Summary
- Fold duplicated author keys like `Black, Chuck, Black, Chuck` to their clean identity during author search dedupe
- Prefer the non-duplicated author label when duplicate and clean provider records collapse together
- Add a regression test based on the Chuck Black/OpenLibrary duplicate shape

## Test plan
- `go test ./internal/metadata`
- `go test ./internal/api`


vavallee (approved):
Reviewed: the fold requires ≥4 tokens with exactly matching halves, so real names can't trip it, and preferring the clean label over provider noise is right — the dup rows are junk author pages. Regression tests cover the OL shape that prompted this. Thanks for the clean fix.

PR #1360, opened 2026-07-01
fix: prune author work subject outliers

pcamp96:
## Summary
- Prune obvious same-name author collisions from author refresh results when a fiction-heavy catalog includes a technical/computer-networking outlier
- Add a regression test based on the Chuck Black / `Software Defined Networks` OpenLibrary collision
- Add a changelog fragment

## Test plan
- `go test ./internal/metadata`
- `go test ./internal/api`

### DIFF
--- a/changelog.d/filter-author-work-collisions.md
+++ b/changelog.d/filter-author-work-collisions.md
@@ -1,0 +1,2 @@
+### Fixed
+- Author refresh now prunes obvious technical same-name collisions from fiction-heavy catalogs, avoiding unrelated works like computer-networking textbooks on fiction author pages.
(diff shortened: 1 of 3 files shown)


vavallee (approved):
Reviewed and approving. The gating is conservative (≥4 works, ≥3 fiction-signal books, ≥2 tech-genre signals, any fiction signal exempts), and the failure mode — a fiction author's genuine tech title isn't auto-added — is recoverable via manual add, same trade-off as the existing compilation pruning. One thing to keep in mind if this ever grows: the keyword lists are tuned to this collision shape; if we see more same-name-collision reports in other domains (cookbooks, law, medicine), the right generalization is probably a majority-subject-cluster check rather than more keyword lists.

Issue #577, opened 2026-05-11
Default import mode is 'move', which silently destroys seeding


vavallee:
**Reported by:** @j-tt (fork: https://github.com/j-tt/bindery)
**Observed on:** v1.7.0 (commit 94a5103), NixOS

## Description

The default import mode when no `import.mode` setting exists is `"move"`. Files are physically moved out of the torrent client's download directory, breaking seeding immediately. The `hardlink` mode exists (free disk cost, preserves seeding) but is not the default and is not suggested during setup.

## File

`internal/importer/scanner.go` — `importMode()`

## Workaround

```sql
INSERT INTO settings(key,value) VALUES('import.mode','hardlink')
  ON CONFLICT(key) DO UPDATE SET value='hardlink';
```

## Suggested fix

@j-tt has a working patch: https://github.com/j-tt/bindery/commit/0b8153a — defaults to `hardlink` when source and destination are on the same filesystem, falls back to `copy` (with a warning) on cross-device imports. Migration 037 in the fork drops the hardcoded `'move'` row written by migration 013 so existing installs pick up the new default.

## Caveat for review

This is a behavior change on upgrade. Migration 037 actively removes the previously-stored `'move'` value, so existing users get new behavior without explicit opt-in. Likely wants to land as a `feat(...)` in a minor release with a CHANGELOG callout, not silently in a patch.

---

Filed from @j-tt's [BUGS.md](https://github.com/j-tt/bindery/blob/main/BUGS.md). Credit to @j-tt for the report and fix.

PR #1445, opened 2026-07-06
fix(web): show auto as the import-mode default instead of move

### LINKED ISSUE #1444: Import Mode UI pre-selects "Move" while the backend default is auto (hardlink/copy)
## Summary

The **Import Mode** selector in Settings → General pre-selects **Move** when no `import.mode` setting has ever been saved, but the backend treats an unset value as **auto** (hardlink on the same filesystem, copy across filesystems). The UI therefore shows "Move" as active while Bindery is actually hardlinking/copying — the displayed mode contradicts the real behaviour.

This is the mirror image of #577: that issue fixed the *backend* default (move → auto, to stop silently breaking seeding), but the frontend was never updated, so the UI still implies Move is the default.

## Impact

- Users reasonably believe files are being **moved** out of the download folder and are confused when the source files remain (they're being hardlinked/copied by design).
- There is currently **no UI path back to auto**: all four buttons (`move`, `copy`, `hardlink`, `external`) write an explicit value, so once a mode is chosen the auto default can't be re-selected without editing the DB.

## Root cause

`web/src/pages/settings/GeneralTab.tsx` defaults the *displayed* mode with `settings['import.mode'] ?? 'move'` (three places). The backend `configuredImportMode()` (`internal/importer/scanner_handoff.go`) returns `""` (= auto → `resolveImportMode` picks hardlink/copy) for an absent/unrecognised value. Display default (`move`) ≠ behavioural default (`auto`).

## Proposed fix (frontend-only behaviour change)

Make **Auto** a first-class, selectable mode so the displayed state always matches backend behaviour:

- Add an **Auto** button (first in the row); default the display to it when `import.mode` is unset (`?? 'auto'`).
- Label it clearly: *"Auto — hardlink if the download folder and library share a volume, else copy. Preserves torrent seeding."*
- Clicking Auto writes `import.mode=auto`, which the backend already resolves to auto (no change to import behaviour).
- Add backend validation for `import.mode` (currently unvalidated) accepting `"", auto, move, copy, hardlink, external` so a typo fails loudly and `auto` is an explicit, documented value.

No change to actual import/placement behaviour — this only makes the UI honest and restores the ability to select the safe default.

johnistheman:
## Summary

The **Import Mode** selector pre-selected **Move** whenever `import.mode` was unset, but the backend has treated an unset value as **auto** (hardlink on the same filesystem, else copy) since #577. The UI therefore showed Move as active while Bindery was actually hardlinking/copying — leaving sources in the download folder and misleading operators into thinking seeding was being broken. This makes **Auto** a first-class, selectable mode so the displayed mode always matches backend behaviour. Closes #1444.

## Implementation notes

- `web/src/pages/settings/GeneralTab.tsx` — add an **Auto** button (first in the row); default the displayed mode to `auto` when unset; rewrite the help text so it names Auto as the default and warns that Move breaks seeding.
- `internal/api/settings_handler.go` — new `SettingImportMode` constant and **validation** for `import.mode` (previously unvalidated), accepting `auto/move/copy/hardlink/external` so a typo fails loudly and `auto` is an explicit, documented value.
- `internal/importer/scanner_handoff.go` — doc comment only: `configuredImportMode` now records that `auto`/unset/unrecognised all map to `""` (no behaviour change).
- `docs/Storage-And-Hardlinks-Wiki.md` — add the **auto** (default) and **external** rows to the Import modes table.

No change to actual import/placement behaviour — this only makes the UI honest and restores a UI path back to the safe default (previously, once any mode was chosen there was no way to re-select auto without editing the DB).

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (`docs/Storage-And-Hardlinks-Wiki.md`)
- [x] Wiki pages updated (user-facing behaviour changed)

## Test plan

- [x] `cd web && npm run typecheck` — clean
- [x] `cd web && npm test` — 390/390 pass (incl. new "defaults to Auto when unset" test)
- [x] `cd web && npm run lint` — 0 errors
- [ ] `go test ./cmd/... ./internal/...` — not run locally (no Go toolchain on this machine); relying on CI for the backend build and `TestValidateSettingValue_ImportMode`

### DIFF
--- a/docs/Storage-And-Hardlinks-Wiki.md
+++ b/docs/Storage-And-Hardlinks-Wiki.md
@@ -14,13 +14,15 @@ This is the standard *arr layout. It matters because **hardlinks only work withi
 
 ## Import modes
 
-Set the import mode in **Settings → File Naming**.
+Set the import mode in **Settings → General → File Naming**.
 
 | Mode | Extra disk | Seeding | Notes |
 |---|---|---|---|
-| **hardlink** | none | kept | Recommended for torrents. The completed file is linked into the library instantly; the download client keeps seeding the same data on disk. Requires downloads and library on one filesystem. |
-| **copy** | doubled | kept | Use when downloads and library are on different filesystems. Copies into the library and leaves the download in place so it can keep seeding. |
+| **auto** *(default)* | none / doubled | kept | Recommended. Hardlinks when the download folder and library share a filesystem (no extra disk), otherwise copies (doubled). The source stays in place either way, so torrents keep seeding. Picks hardlink or copy automatically per download. |
+| **hardlink** | none | kept | Forces hardlinking for torrents. The completed file is linked into the library instantly; the download client keeps seeding the same data on disk. Requires downloads and library on one filesystem. |
+| **copy** | doubled | kept | Forces copying. Use when downloads and library are on different filesystems. Copies into the library and leaves the download in place so it can keep seeding. |
 | **move** | none | **broken** | Moves the file out of the download location, so a torrent can no longer seed it. Only suitable for Usenet, or when you do not seed. |
+| **external** | none | kept | Hands off to a sibling tool (Calibre, CWA, Grimmory, Storyteller). Bindery stops after grabbing; the external tool processes and places the file, then Bindery reconciles it on the next library scan. Can drop the file into a configured watch folder. |
 
 ## Multi-disc audiobook flattening
 
--- a/internal/api/settings_handler.go
+++ b/internal/api/settings_handler.go
@@ -84,6 +84,14 @@ const SettingSearchInterval = "search.interval"
 // the literal in sync with this constant.
 const SettingImportAudiobookFlattenMultiDisc = "import.audiobook.flatten_multi_disc"
 
+// SettingImportMode is the placement mode for completed downloads. Recognised
+// values are "auto" (empty/unset behaves identically — hardlink when the source
+// and destination share a filesystem, else copy; preserves seeding), "move"
+// (relocate the source, breaking seeding), "copy", "hardlink", and "external"
+// (hand off to a sibling tool). The importer reads this key as a string literal
+// (configuredImportMode) to avoid an import cycle; keep the literal in sync.
+const SettingImportMode = "import.mode"
+
 type SettingsHandler struct {
 	settings *db.SettingsRepo
 }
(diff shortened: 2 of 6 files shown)


vavallee:
This looks complete to me. Tests are green, coverage is green, and the approach matches what we discussed in #1444. Want to flip it to ready for review? I'd like to get it into 1.25.0. If you're short on time just say so and I can take it from here, but it's your PR and your call.

johnistheman:
Flipped to ready — thanks for the review!

One heads-up before merge: the **`validate (Go race)`** check is red, but it's a package-level timeout, not a failure from this diff:

- `panic: test timed out after 25m0s` → `FAIL github.com/vavallee/bindery/internal/api 1500.057s`, preceded by `ERROR refresh all authors: list failed error="context deadline exceeded"`.
- The non-race `validate (Go)` ran the same suite green in **1m29s**; `smoke`, ABS contract, and coverage are all green too.
- This PR's only Go addition is `TestValidateSettingValue_ImportMode` — a pure-function test over `validateSettingValue(string, string)` with no I/O, goroutines, or DB, so it cannot add 25 minutes. The timeout is in the author-refresh path, which this PR doesn't touch.

Looks like pre-existing flakiness in the `-race` variant of the `internal/api` package (that job only runs on PRs, so `main` has no baseline to compare). I don't have re-run rights on the repo — could you re-run just that job? Happy to raise the package's test timeout in a separate PR if it turns out to be recurring.

vavallee:
One small thing I noticed while poking around: README.md line 130 (Import & organize section) still says Modes: Move (default). Since your PR makes the UI honest about auto being the default, that line is the last place claiming otherwise. Worth folding into this PR if you're up for it, otherwise I'll clean it up after.

vavallee:
You read that exactly right. The internal/api package under -race is a known pre-existing flake, I've hit the same timeout locally on two different Go versions, and it's why that job is deliberately non blocking in CI. Your diff adds one pure function test, it's not the culprit. No re-run needed.

Merging now. I'll pick up the README default-mode line and the changelog fragment in a small follow-up so you don't need another round trip. Thanks for a genuinely tidy fix, the loud validation on typos is exactly the right call.

vavallee (approved):
UI default now tells the truth, backend validates loudly, docs updated. The race-job red is pre-existing flakiness, not this diff.

PR #1447, opened 2026-07-07
fix(settings): make Auto a first-class Import Mode

### LINKED ISSUE #1444: Import Mode UI pre-selects "Move" while the backend default is auto (hardlink/copy)

vavallee:
## Summary
- The Import Mode selector pre-selected **Move** whenever `import.mode` had never been saved, but the backend (`resolveImportMode`) has defaulted an unset value to **auto** (hardlink same-filesystem, copy otherwise) since #577 — the display default never followed, so the UI showed Move as active while Bindery was actually hardlinking/copying.
- **Auto** is now a real, selectable button (first in the row), and it's the display default when `import.mode` is unset — so there's a way back to auto after picking something else, and the UI always matches actual backend behavior.
- Added backend validation for `import.mode` (previously unvalidated) accepting `"", auto, move, copy, hardlink, external` so a typo fails loudly instead of silently defaulting to auto.
- No change to actual import/placement behaviour.

Fixes #1444

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/api/... ./internal/importer/...`
- [x] `npx tsc --noEmit`
- [x] `npx vitest run src/pages/SettingsPage.test.tsx` (51 passed, including the existing import-mode persistence test)

### DIFF
--- a/changelog.d/1444-import-mode-auto-default.md
+++ b/changelog.d/1444-import-mode-auto-default.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Import Mode UI now shows Auto as the default** (#1444) — the selector used to pre-select Move when no mode had ever been saved, while Bindery actually defaults to auto (hardlink same-filesystem, copy otherwise). Auto is now a real, selectable option, so the UI matches what's actually happening and you can switch back to it after picking something else.
--- a/internal/api/settings_handler.go
+++ b/internal/api/settings_handler.go
@@ -69,6 +69,13 @@ const (
 	SettingImportDropLinkMode = "import.drop_link_mode"
 )
 
+// SettingImportMode is the KV key for the operator-chosen import placement
+// mode. Valid values are "", "auto" (equivalent — hardlink same-filesystem,
+// copy otherwise), "move", "copy", "hardlink", "external". The importer
+// package reads this key as a string literal to avoid an import cycle; keep
+// the literal in sync with this constant.
+const SettingImportMode = "import.mode"
+
 // SettingSearchInterval is the KV key for the wanted-book search cadence.
 // Value is a Go duration string (e.g. "12h", "24h"). Empty or unset falls
 // back to the scheduler's defaultSearchInterval (12h). Takes effect on restart.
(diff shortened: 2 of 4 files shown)


vavallee:
Duplicate — closing in favor of #1445, which fixes the same root cause (#1444) plus adds test coverage and a wiki update. Thanks for the report and the fix, @johnistheman.

PR #1531, opened 2026-07-13
fix(ci): parallelize Go race tests to avoid timeouts

magrhino:
## Summary

The Go race job regularly reaches the 25-minute `internal/api` package timeout even though no data race is reported, leaving otherwise healthy PR and main CI runs red. This splits the same race coverage across dedicated runners so the race stage can finish quickly without weakening the tests it executes.

- run four exhaustive `internal/api` name shards plus one shard for every other Go package
- preserve `Test`, `Fuzz`, and `Example` execution across PR, main, and release-tag events
- cap each shard at 15 minutes and each job at 20 minutes
- propagate package-discovery failures through the non-API pipeline with `pipefail`

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed — N/A: CI execution only
- [ ] Wiki pages updated if user-facing behaviour changed — N/A: no user-facing behaviour

## Test plan

- [x] Equivalent individual checks for `make check`
  - `go build ./...`
  - `go vet ./...`
  - `golangci-lint v2.11.4 run --timeout=5m`
  - `govulncheck ./...`
  - all four API race shards with `-count=1` (4m32s–5m05s locally)
  - the exact non-API `pipefail` race pipeline (passed locally)
  - `npm ci --prefix web`
  - `npm run lint --prefix web`
  - `npm run typecheck --prefix web`
  - `npm run build --prefix web`
  - `npm test --prefix web` (46 files, 409 tests)
- [x] `actionlint -shellcheck= .github/workflows/ci.yml`
- [x] shard enumeration: 753 API test/fuzz/example entries, zero gaps or overlaps

Full `actionlint` also reports the pre-existing SC2034 warning in an unrelated workflow script; the same warning reproduces on `upstream/main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,8 +112,9 @@ jobs:
           fail_ci_if_error: false
           token: ${{ secrets.CODECOV_TOKEN }}
 
-  # PR-only race detector. Four API test-name shards plus the remaining Go
-  # packages run in parallel so internal/api does not dominate PR feedback.
+  # PR-only race detector. Four API test-name shards, internal/db, and the
+  # remaining Go packages run in parallel so large packages do not dominate
+  # PR feedback.
   # The post-merge race job uses the same shard layout.
   validate-race-go:
     name: validate (Go race) (${{ matrix.shard }})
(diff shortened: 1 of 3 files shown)


magrhino:
In the follow up commit I isolated `internal/db` test so non api test was under 10 minutes.

vavallee (approved):
Verified and merging. I checked the shard patterns actually partition the API suite with no gaps or overlap: Test[A-B], Test[C-K], Test[L-Q], and the api-rest catch-all (^(Test([^A-Q]|$)|Fuzz|Example)) cover every possible name after "Test" exactly once, and Fuzz/Example land only in api-rest so they run once. All six race shards (four API, db, non-api) came back green here in well under the 15m cap, which is the proof both that coverage held and that there's no actual data race hiding under the old timeout. Required checks are unaffected (the old single "validate (Go race)" was never a required context). This kills a flake that's been red on basically every PR. Nice fix, thank you.

PR #1562, opened 2026-07-18
fix(api): honor selected download format (#1561)

### LINKED ISSUE #1561: Audiobook download ignores selected format for dual-format books
### Pre-flight

### Current behaviour

Downloading an audiobook from a dual-format book can sometimes return the EPUB instead or nothing at all if the books.file_path is not updated.

### Expected behaviour

Bindery downloads the audiobook selected in the UI.

The frontend sends: 
`GET /api/v1/book/{id}/file?format=audiobook`
However, the download handler (files.go) does not use the format query parameter. It selects paths in this order:
Legacy `FilePath`
`EbookFilePath`
`AudiobookFilePath`

### Steps to reproduce

1. Open a book configured with media type **Both**.
2. Ensure it has separate ebook and audiobook file paths.
3. Select **Audiobook** in the File section.
4. Click **Download**.

### Trace logs

```text
I could not find any logs relating to downloading files from the UI.
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v1.26.0

### Environment details

Ubuntu 26.04

Zenuka:
## Summary
Honor the format selected when downloading files from dual-format books. Audiobook requests now use the audiobook path instead of falling through the legacy or ebook path, while callers that omit the format retain the existing fallback behavior.

Invalid format values now return HTTP 400, and regression tests cover both ebook and audiobook selection.

**Important side note**: I’m a C# developer, and until today I had never written a single line of Go. My expertise mostly began and ended with knowing the mascot is adorable. This is my first Go contribution, developed with AI assistance. I’ve reviewed and tested the change, but please keep me honest if anything doesn’t match the project’s conventions. 🙂

Closes #1561.

## Checklist

- [x] Tests added or updated
- [ ] Wiki pages updated if user-facing behaviour changed;

## Test plan
- [x] `go vet ./...`
- [x] Frontend production build
- [x] `docker buildx build --platform linux/amd64,linux/arm64 -t bindery:local .`
make check had issues on unrelated files (internal/decision/decision.go, internal/decision/specs.go, internal/downloader/nethint/nethint.go)

### DIFF
--- a/changelog.d/1561-audiobook-download-format.md
+++ b/changelog.d/1561-audiobook-download-format.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Dual-format audiobook downloads** (#1561) — honor the selected format so audiobook downloads no longer return an ebook or fail through a stale legacy path.
--- a/internal/api/files.go
+++ b/internal/api/files.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/vavallee/bindery/internal/auth"
 	"github.com/vavallee/bindery/internal/db"
+	"github.com/vavallee/bindery/internal/models"
 )
 
 type FileHandler struct {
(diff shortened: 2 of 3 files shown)


vavallee:
Hey, sorry about how this played out. You filed the issue early Friday and had this PR up an hour later, and then I shipped my own fix for the same bug in #1563 later that day without checking open PRs first. That one went out in v1.26.1, so I'm closing this as superseded. That's about timing, not about the quality of this PR.

Since you asked to be kept honest: your fix was almost exactly what got merged. Same switch on the format param, same 400 on bad values, solid tests, changelog fragment done right. The one real difference is the legacy fallback. You gated it on the book's MediaType, the merged version infers the format from the on disk shape (directory means audiobook bundle, plain file means ebook), which holds up better for rows that predate the dual format schema where MediaType can be stale. Genuinely impressive for a first line of Go.

Your report and diagnosis in #1561 are what made the same day fix possible, so thank you. If you want to pick up something else I'd be glad to have you, and next time there's an open PR I'll build on it instead of racing past it.

Zenuka:
Thanks for your reply and no worries, glad we could fix the issue!

PR #1571, opened 2026-07-18
Feature/optional parent indexer categories

the-archbishop:
## Summary

Adds a per-indexer option to include broad Newznab Books (`7000`) and Audio (`3000`) parent categories in searches. This helps recover releases from indexers with incomplete category mappings while remaining disabled by default to avoid unrelated results.

- Adds the option to indexer create/edit forms and API responses.
- Persists the setting through a forward-only database migration.
- Preserves the local choice during Prowlarr synchronization.
- Maintains compatibility with older API clients that omit the field.
- Updates backend and frontend tests, documentation, translations, and the changelog fragment.
- Ensures `SearchBook` adds the appropriate parent to the outbound category list even when earlier Prowlarr synchronization removed it from the stored categories.
- Applies each indexer's preference independently and avoids duplicate parent categories.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed *(N/A; no deployment configuration or upgrade steps changed)*
- [x] Added a changelog fragment under `changelog.d/` *(not an edit to `CHANGELOG.md`)*
- [x] Wiki pages updated if user-facing behavior changed *(N/A; the relevant guidance is in `docs/QUICKSTART.md`, which was updated)*

## Test Plan

- [x] Targeted indexer category and `SearchBook` tests.
- [x] Go build, vet, lint, and race tests for the changed indexer package.
- [x] `go test ./internal/api`
- [x] Frontend typecheck, lint, production build, and 410 tests.
- [x] GoReleaser snapshot validation.
- [x] `make check` — all stages passed except the race-enabled `internal/api` package exceeded the 10-minute timeout while repeatedly applying test migrations. No assertion or race failure was reported, and the package passed separately without race instrumentation.
- [x] Built and deployed the updated container alongside the existing Bindery installation using the same database, enabled broad parent categories on an existing indexer, and confirmed that ebook searches include `7000` and can successfully grab EPUB releases advertised only under that parent category.

### DIFF
--- a/changelog.d/broad-parent-categories.md
+++ b/changelog.d/broad-parent-categories.md
@@ -1,0 +1,2 @@
+### Added
+- **Optional broad indexer categories** — Indexers can now opt in to searching the Newznab Books (`7000`) or Audio (`3000`) parent categories alongside their configured subcategories, helping recover releases from incomplete category mappings. The setting is available when adding or editing an indexer, persists across updates, and is preserved during Prowlarr syncs; it remains disabled by default because broad categories can return unrelated results.
--- a/docs/QUICKSTART.md
+++ b/docs/QUICKSTART.md
@@ -63,10 +63,11 @@ point Bindery at a Prowlarr instance.
   non-standard categories.
 - **Include broad parent categories** — disabled by default. Enable it for an
   indexer that advertises releases only under Books (`7000`) or Audio (`3000`).
-  Bindery then searches the supplied parent alongside specific children; it
-  never adds an unadvertised parent. Broader queries can recover releases
-  missed by incomplete mappings, but can also return comics, magazines, music,
-  or other unrelated results. Prowlarr re-syncs preserve this local choice.
+  Bindery then adds the appropriate parent to searches alongside the specific
+  children, including for existing indexers whose stored mappings no longer
+  contain the parent. Broader queries can recover releases missed by incomplete
+  mappings, but can also return comics, magazines, music, or other unrelated
+  results. Prowlarr re-syncs preserve this local choice.
 
 **Prowlarr** — under **Settings → Indexers → Add Prowlarr**, give the base URL
 (e.g. `http://prowlarr:9696`) and API key, then **Sync now** to pull its
--- a/docs/QUICKSTART.md
+++ b/docs/QUICKSTART.md
@@ -61,6 +61,12 @@ point Bindery at a Prowlarr instance.
 - **Categories** — comma-separated Newznab category IDs. Default `7020`
   (eBooks); `3030` is Audiobooks. Add custom IDs for indexers with
   non-standard categories.
+- **Include broad parent categories** — disabled by default. Enable it for an
+  indexer that advertises releases only under Books (`7000`) or Audio (`3000`).
+  Bindery then searches the supplied parent alongside specific children; it
+  never adds an unadvertised parent. Broader queries can recover releases
+  missed by incomplete mappings, but can also return comics, magazines, music,
+  or other unrelated results. Prowlarr re-syncs preserve this local choice.
 
 **Prowlarr** — under **Settings → Indexers → Add Prowlarr**, give the base URL
 (e.g. `http://prowlarr:9696`) and API key, then **Sync now** to pull its
--- a/internal/api/indexers.go
+++ b/internal/api/indexers.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"encoding/json"
+	"io"
 	"net/http"
 	"strconv"
 	"sync"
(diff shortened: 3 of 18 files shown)


vavallee:
thanks. 

The sync preservation and prepending the parent even when an earlier sync stripped it from stored categories show you traced the whole pipeline, and the presence aware update handler is exactly how I'd want it done.

One behavior question before merge: the non standard taxonomy fallback in the search side filterCategoriesForMedia used to return the configured categories verbatim, parents included. Now it strips parents unless the indexer opted in. Someone running a MaM style indexer with 7000 sitting in their category list gets quietly narrower ebook searches after this change. If that's intentional it matches the documented policy and just deserves a line in the changelog fragment, otherwise the verbatim passthrough should stay for non opted in indexers.

vavallee (changes requested):
Sorry this sat for a while. I've gone through it properly now, including merging it against current main in a scratch worktree to check the rebase and to instrument the category logic.

The plumbing is good and it's clear you traced the whole pipeline through DB, sync, search, API and UI, and validated it on a real install. A few things to sort before it goes in.

**Blocker: the migration number collides and will refuse to boot.**

`064_indexer_parent_categories.sql` clashes with `064_downloads_import_path.sql`, which landed after you opened this. Main is at `066` now. Git won't flag it since the filenames differ, but `assertUniqueMigrationVersions` (`internal/db/db.go:429`) fails startup on a duplicate prefix, it doesn't skip quietly. Rename to `067_`.

The migration is otherwise fine, forward only and shaped like 053 and 066. Worth adding a `-- +migrate Down` with the `DROP COLUMN` to match those two. No semicolon problem, your only one is the statement terminator.

**Rebase: three files, all mechanical.**

`internal/db/indexers.go` and `web/src/pages/settings/IndexersTab.tsx` conflict with the freeleech only feature (#1624), which touched the same column lists and the same insertion point in the form. `internal/indexer/searcher.go` conflicts only in a doc comment that #1576 rewrote next to yours. It's "keep both sides" everywhere. #1609 landed in searcher.go too but in a different region, so it doesn't conflict. I confirmed the resolved tree builds and all four Go packages plus the frontend suite pass.

**The opt-in leaks across media types.** One boolean gates both 7000 and 3000, and the parent gets picked from the search's media type rather than from what the indexer actually carries. Instrumented:

```
cats=[7000,7020]  mt=audiobook  opt=true  ->  [3000 3030]
cats=[3000,3030]  mt=ebook      opt=true  ->  [7000 7020]
```

So a books only indexer, opted in via a checkbox that says "Books (7000) or Audio (3000)", now sends `cat=3000` at every audiobook search and pulls the indexer's whole music surface. `searcher.go:152` prepends unconditionally with no check that the indexer has anything in that bucket. Please guard it.

**It also injects a standard parent into non standard taxonomy indexers.**

```
cats=[100060]  mt=ebook  opt=true  ->  [7000 100060]
```

The doc comment right above that function argues the opposite as settled policy, that substituting standard IDs on these indexers returns unrelated results because the standard IDs don't cover the extended subcategory tree. Injecting 7000 into a MaM style indexer is the same mistake, and `searcher_test.go` currently pins it as intended. Skip the parent when there's no standard range match.

**The opt-out path changed silently.** `searcher.go:140` replaced `return cats` with a loop that strips the parent, so `[7000, 100060]` with the option off now sends `[100060]` where main sends both. That's arguably more consistent but it changes every non opted in mixed taxonomy indexer, and it's not in the PR body, the changelog, or any test. Either revert that branch or call it out.

**Worth knowing about, not a blocker:** `BookSearch`'s tiers 2 to 4 short circuit on any result at all, before relevance filtering. Broadening categories makes it likelier that an early text tier returns junk, stops the ladder, and then `filterRelevant` discards everything, so an opted in indexer can end up with *fewer* good results than a narrow list would have found. Ideally the opt-in would be confined to the structured tier 1 `t=book` query. At minimum it deserves a line in the doc comment.

**API decode.** The embedded struct plus `*bool` shadowing works, and the omit versus explicit false tests confirm it. But it's a per field patch on a general problem, and main's handler already zero values `FreeleechOnly` for any client that omits it. This does the same job in one line and fixes freeleech for free:

```go
idx := *existing
if err := json.Unmarshal(body, &idx); err != nil { ... }
```

Unmarshal only overwrites keys present in the body. It drops the embedded struct, the `io.ReadAll` and the `*bool`, and `failingReader` in the tests goes away with it.

Smaller stuff: drop the two English strings from `ko.json`, convention here is to omit untranslated keys and let i18next fall back. Match `FreeleechOnlyField`'s markup now that they'll sit next to each other, and consider moving the field under Categories since that's what it modifies. Rename the fragment to `1571-broad-parent-categories.md` with `(#1571)` in the entry.

On the sync side, your preservation claim holds. I traced it rather than trusting the test: the update path mutates the row loaded from the DB, which already carries the flag, and nothing zero values it. Two things you may want to note in the description: it isn't preserved if an indexer disappears from Prowlarr for a sync and gets recreated, and with the opt-in on, `filterCategoriesForMedia` returns categories verbatim so the 7000 to 7020 widening stops happening.

Tests are good, especially the httptest server asserting the literal outbound `cat=`. The gaps line up exactly with the bugs above: no cross media type case, no non standard taxonomy negative, and nothing covering the opt-out change.

Ping me when it's rebased and I'll take another look.

vavallee:
Landed in #1887, which is your three commits rebased onto current main with your authorship intact, plus one follow up commit for the review items. Closing this one because the branch had drifted a month behind and the migration number would have stopped the app booting, not because anything was wrong with the idea. Thank you for it, and sorry it sat this long before I picked it up.

What changed from your version, so nothing is a surprise:

**Migration renumbered to 073.** Yours was 064, which collided with `064_downloads_import_path.sql` and would have failed `assertUniqueMigrationVersions` at startup. Added the `-- +migrate Down` too.

**The parent is now chosen from what the indexer carries, not from the search.** One boolean gated both 7000 and 3000, so a books only indexer that opted in was sending `cat=3000` on every audiobook search and pulling that tracker's whole music surface. There is now a guard that only adds a parent the indexer actually has subcategories under.

**Non standard taxonomy indexers are left alone.** Opting in was injecting 7000 into MaM style lists, which is the thing the doc comment above that function argues against. The opt in can no longer reach that branch.

**Reverted the opt out change.** Your loop meant `[7000, 100060]` with the option off sent only `[100060]`, where main sends both. It was unrelated to the feature and untested, so I put it back.

**Simplified the update decode** to `idx := *existing` then decode over it, which does the same job as the embedded struct and the `*bool` in one line, and fixes the same omission bug for `freeleechOnly` for free.

I also dropped the Prowlarr syncer half. The search time guard rebuilds the parent from the surviving children, so a row whose parent Prowlarr stripped still gets it back at query time, and keeping it would give the option a second place to change behaviour. Say the word if you disagree, that one is a judgement call rather than a fix.

Three tests were added for the gaps: cross media type, non standard taxonomy, and the opt out case. Your httptest server asserting the literal outbound `cat=` survived, it was the best part of the test suite here.

Issue #563, opened 2026-05-09
Author matching: token-overlap false positives pull in books from same-named co-authors


vavallee:
## Reported behaviour

Items are being pulled in for an author when a *different* multi-author work happens to contain the right surname tokens. Example: monitored author "Rachel Reid" gets a book imported whose author field is "Rachel Larsen, Adam Reid, and Ozi Akturk" — the tokens "Rachel" and "Reid" both appear, but on different actual people.

## Suspect code paths (in descending likelihood)

After reading the code, three sites do token-style author matching that could produce this. Without a specific imported example I can't say which one fired here:

### 1. Indexer release filter — `internal/indexer/searcher.go:filterRelevant`

The release-relevance check requires the release to contain only the **surname** at a word boundary (line 304: `surname := AuthorSurname(author)`; line 320-323 calls `titleMatchesResult(... sn ...)`). So a release titled `Adam.Reid.Some.Book.epub` would pass the gate when `Author = "Rachel Reid"` because `"reid"` is in there. It would then be auto-grabbed by the searcher and ingested.

This is the most likely source if the user has indexer auto-grab on.

### 2. Library scanner — `internal/importer/scanner.go:authorMatch` (lines 981-991)

```go
parts := strings.Fields(strings.ToLower(parsedAuthor))
lastName := parts[len(parts)-1]
return len(lastName) >= 3 && strings.Contains(strings.ToLower(bookAuthor), lastName)
```

`strings.Contains` is substring match, not word-boundary, and only the last name is checked. So `parsedAuthor = "Rachel Reid"` resolves to `lastName = "reid"` and matches any `bookAuthor` containing "reid" anywhere — including `"Adam Reid"`, `"Reid Hoffman"`, etc.

This is the most likely source if the user is scanning a local library.

### 3. ABS importer — `internal/abs/import_author_matcher.go`

This one uses full-name comparisons via `textutil.MatchAuthorName` (Jaro-Winkler whole-string), so it should NOT false-positive on `"Rachel Reid"` vs `"Rachel Larsen, Adam Reid, and Ozi Akturk"` — those whole-strings are very different. Listing here for completeness so a fix doesn't accidentally tighten the ABS path.

## Likely fix shape (pending diagnosis)

Once we know which path fired:

- For (1): augment `titleMatchesResult` to also require the **first name** (or a reasonable initial form) at a word boundary when the requested author has more than one token. "Rachel" + "Reid" both must match, contiguously preferred.
[...]

Issue #409, opened 2026-04-28
Possessive author titles fail relevance matching (e.g. "Tom Clancy's Rainbow Six")


jhughes2466:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

When OpenLibrary stores a book title with a possessive author prefix (e.g. "Tom Clancy's Rainbow Six"), Bindery's relevance filter rejects valid releases named "Tom Clancy - Rainbow Six" with "title/author keywords did not match release name" even though the release is correct.

### Expected behaviour

Bindery should strip possessive prefixes from titles during matching and recognize "Tom Clancy - Rainbow Six" as a valid match for "Tom Clancy's Rainbow Six" by Tom Clancy.

### Steps to reproduce

Add Tom Clancy, search for Rainbow Six etc...

### Trace logs

```text
No log data, because it finds some results which it sees as "good" but fails to find specific ones, especially with torrent sites.  Confirmed issue with torrentday using both the bindery 3030, etc... and the torrentday matching category numbers.
```

### Deployment method

Docker (ghcr.io)

### Bindery version

sha-99e8ef8

cranium:
the possessive "Tom Clancy's" is pretty much only there for the videogames- the books were never marketed that way

jhughes2466:
So the issue I'm hitting is the relevancy filter is killing good results.  IE: I just tested and got 18 valid results from AudioBookBay but none of them filtered into the list.   Adding screenshots:

<img width="974" height="821" alt="Image" src="https://github.com/user-attachments/assets/575b70b4-dfaf-446d-98cc-197afa388637" />

<img width="1021" height="433" alt="Image" src="https://github.com/user-attachments/assets/9041a506-1087-4ef9-950d-6ed1465e262a" />


I haven't looked at the source code yet, but from what I can tell the relevancy filter is so strict that a large portion > 80% of valid results are also filtered out.

In this specific example I had to manually update the SQLite DB to add in the correct 3000, 3030 because the gui for some reason stored them oddly and found 0 results.  After I made the update and docker restart, it found 18 results, but all were filtered out.  

Love the idea, if I can help let me know.  I don't have quite as much time as I used to, but I can step on on some, but I don't want to overstep.  

If nothing else if I can fork the repo, I can do some localized testing and send results back.  Let me know.  I don't typically ever do anything like that without permission from the creator.

jhughes2466:
[...]

PR #1609, opened 2026-07-23
fix(indexer): normalize author names like release titles so dotted initials match (#1608)

### LINKED ISSUE #1608: Author names with dotted initials ("J.R.R. Tolkien") produce author tokens that can never match a release — single-keyword titles return zero results
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate (closest are #563, which introduced `authorTokens` initials-dropping, and #1539/#409 which touch the same filter but not this token-alphabet mismatch)
- [x] I'm running the latest version (reproduced on v1.26.2; the code path is unchanged on current `main` / v1.27.0)

### Current behaviour

For any author whose canonical metadata name spells initials in compact dotted form — OpenLibrary's **"J.R.R. Tolkien"** being the flagship case — every search for a book whose title reduces to a **single significant keyword** ("The Hobbit" → `hobbit`, "The Silmarillion" → `silmarillion`) returns **0 results**, even when the indexers return dozens of obviously-correct releases. Search debug shows all of them eaten at the relevance stage (`afterUsenetJunk=N`, `afterRelevance=0`).

Root cause (in `internal/indexer/searcher.go`):

- `filterRelevant` matches releases via `titleMatchesResult(NormalizeRelease(releaseTitle), sigWords, authorTokens(author), true)`.
- `NormalizeRelease` collapses **all** punctuation to spaces: `"J R R Tolkien - The Hobbit (retail) (azw3)"` → `"j r r tolkien the hobbit retail azw3"`. Dots never survive on the release side.
- `authorTokens` only trims punctuation from token **edges** (`strings.Trim(w, ".,;:()[]")`), so `"J.R.R. Tolkien"` tokenizes to `["j.r.r", "tolkien"]`. The token `j.r.r` keeps its interior dots and therefore **cannot appear in any normalized release string, ever**.
- For single-sig-word titles, `titleMatchesResult` hard-requires `authorMatchesRelease`, whose all-tokens fallback demands *every* significant token — including the unmatchable `j.r.r` — so every release is rejected. This includes a release literally titled `"The Hobbit - J. R. R. Tolkien"`.

The same interior-punctuation mismatch hits hyphenated first names: `authorTokens("Mary-Kate Olsen")` = `["mary-kate", "olsen"]`, and `\bmary\-kate\b` can never match the normalized haystack `"mary kate olsen …"` (the existing `TestAuthorMatchesReleaseHyphenated` passes only because its haystack is hand-written *unnormalized* text — production haystacks always come out of `NormalizeRelease`).

What makes this maddening to diagnose: **multi-word titles escape** through the anchored phrase-match path (which never needs the full author token set), so "The Lord of the Rings" works fine while "The Hobbit" silently returns nothing for the same author.

The spaced form behaves fine: `authorTokens("J. R. R. Tolkien")` drops the single-letter initials (per #563) leaving `["tolkien"]`, and everything matches. Our workaround in production was to merge the author into a duplicate with a spaced-out name.

### Expected behaviour

`"J.R.R. Tolkien"` should decompose exactly like `"J. R. R. Tolkien"` does — author tokens should live in the same alphabet as the normalized release string they are matched against. With author `"J.R.R. Tolkien"` and title `"The Hobbit"`, a release named `"J R R Tolkien - The Hobbit (retail) (azw3)"` must pass the relevance filter.

### Steps to reproduce

Deterministic unit-test repro against the repo (no indexer needed):

```go
func TestDottedInitialsRepro(t *testing.T) {
	toks := authorTokens("J.R.R. Tolkien") // = ["j.r.r", "tolkien"]
	releases := []string{
		"J R R Tolkien - The Hobbit (retail) (azw3)",
		"The Hobbit - J. R. R. Tolkien",
		"J.R.R. Tolkien - The Hobbit EPUB",
	}
	for _, r := range releases {
		got := filterRelevant([]newznab.SearchResult{{Title: r}}, "The Hobbit", "J.R.R. Tolkien", nil)
		if len(got) == 0 {
			t.Errorf("dropped %q", r) // fires for ALL of them today
		}
	}
}
```

All three releases are dropped; `titleMatchesResult` returns false for every one. Swap the author to `"J. R. R Tolkien"` and all three pass.

Live repro: add J.R.R. Tolkien from OpenLibrary (compact dotted form is his canonical OL name), mark "The Hobbit" wanted, run a search. Indexers return dozens of releases; Bindery reports zero.

### Trace logs

From a live deployment (v1.26.2, two Newznab indexers — NZBgeek and scenenzbs behind a Prowlarr proxy), sanitized: interactive search debug for "The Hobbit" / J.R.R. Tolkien shows the pipeline counts `rawCount=34 afterDedupe=29 afterUsenetJunk=29 afterRelevance=0`, with per-candidate `stage=relevance` rejections for every release including `The Hobbit - J. R. R. Tolkien`. The unit test above is the distilled, log-independent reproduction of the same run.

### Deployment method

Helm / Kubernetes

### Bindery version

v1.26.2 (code path unchanged at v1.27.0 / current `main`)

### Environment details

Two Usenet indexers (NZBgeek, and scenenzbs via a Prowlarr proxy). Not indexer-dependent — the filter drops the results after they come back.

### Proposed fix

Tokenize the author through the same normalization the release side uses — e.g. build `authorTokens` from `strings.Fields(NormalizeRelease(author))` — so both sides of the comparison share one alphabet by construction. That preserves the #563 initials-dropping behaviour (`"J.R.R." → "j r r"` → all tokens <3 chars → dropped, same as the spaced form) and also fixes the hyphenated-name variant. I have a PR with regression tests ready and will link it here.

helios57:
## Summary

Closes #1608.

**Problem** — for any author whose canonical metadata name uses compact dotted initials (OpenLibrary's "J.R.R. Tolkien" being the flagship case), every search for a book whose title reduces to a single significant keyword ("The Hobbit", "The Silmarillion") returns zero results, even when indexers return dozens of correct releases — including one literally titled `The Hobbit - J. R. R. Tolkien`. Multi-word titles escape via the anchored phrase path, which makes the failure look title-specific and maddening to diagnose.

**Root cause** — `authorTokens` only trimmed punctuation from token *edges* (`strings.Trim(w, ".,;:()[]")`), so `"J.R.R. Tolkien"` tokenized to `["j.r.r", "tolkien"]`. `NormalizeRelease` collapses all punctuation to spaces on the release side, so the token `j.r.r` can never occur in any haystack it is matched against; the all-tokens fallback in `authorMatchesRelease` therefore failed for every release, and single-sig-word titles (which hard-require the author) dropped everything. Hyphenated first names ("Mary-Kate" → token `mary-kate`) had the same defect against production (normalized) haystacks.

**Fix** — tokenize the author through `NormalizeRelease` so both sides of the comparison share one alphabet by construction. The #563 initials-dropping behaviour is preserved: `"J.R.R."` now decomposes exactly like the spaced `"J. R. R."` (sub-3-char initials drop, `["tolkien"]` remains). No change to matching semantics beyond the tokenization; umlaut transliteration and apostrophe stripping are inherited from `NormalizeRelease` (which also covers the Unicode `’` the old code missed).

Found while debugging a live deployment (v1.26.2) where every Tolkien single-keyword title auto-search returned empty; the code path is unchanged on current `main`.

## Checklist

- [x] Tests added or updated
- [ ] Wiki pages updated if user-facing behaviour changed — n/a (bug fix, no behaviour contract change)

## Test plan

- New `TestAuthorTokensDottedInitials` — compact dotted, spaced, and hyphenated forms produce the expected token sets
- New `TestFilterRelevantDottedInitialsAuthor` — end-to-end regression: author "J.R.R. Tolkien", title "The Hobbit", release `J R R Tolkien - The Hobbit (retail) (azw3)` (and punctuation variants) pass the filter; a wrong-author release (`The Hobbit by Jane Doe EPUB`) is still rejected
- Extended `TestAuthorMatchesReleaseHyphenated` — now also asserts the *normalized* (hyphen→space) haystack matches, which is what production actually feeds the matcher
- Existing behaviour guards stay green: `TestAuthorMatchesReleaseInitials` (#563 initials-dropping), `TestFilterRelevantReorderedKeywordCollisions`, possessive-prefix (#409), umlaut, and iso639 alias tests

### DIFF
--- a/changelog.d/1608-author-dotted-initials.md
+++ b/changelog.d/1608-author-dotted-initials.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Authors with dotted initials match releases again** (#1608) — a name like "J.R.R. Tolkien" produced an author token ("j.r.r") that can never appear in a normalized release name, so searches for single-keyword titles ("The Hobbit") returned zero results even when indexers found dozens. Author names are now normalized the same way release titles are before tokenizing; hyphenated first names ("Mary-Kate") are fixed by the same change.
--- a/internal/indexer/searcher.go
+++ b/internal/indexer/searcher.go
@@ -295,20 +295,28 @@ func stripPossessivePrefix(title, author string) string {
 }
 
 // authorTokens splits an author name into a (significant, all-lowercased)
-// token list suitable for word-boundary matching. Significant means >=3 chars
-// of letters/digits; shorter tokens (typically initials like "R." or "R")
-// are treated as optional and dropped. German umlauts are transliterated to
-// match NormalizeRelease. Returns nil for empty / all-initials input — the
-// caller should fall back to surname-only behaviour.
+// token list suitable for word-boundary matching. The name is passed through
+// NormalizeRelease first so the tokens live in the same alphabet as the
+// normalized release strings they are matched against: interior punctuation
+// becomes a token boundary exactly as it does on the release side. Previously
+// only edge punctuation was trimmed, so "J.R.R. Tolkien" produced the token
+// "j.r.r" — which can never occur in a NormalizeRelease haystack (dots
+// collapse to spaces there), making every all-tokens author match fail and
+// zeroing out single-keyword-title searches for such authors (#1608).
+// Hyphenated first names ("Mary-Kate") had the same defect.
+//
+// Significant means >=3 chars of letters/digits; shorter tokens (typically
+// initials like "R." or "R") are treated as optional and dropped, so the
+// compact dotted form decomposes exactly like the spaced "J. R. R. Tolkien".
+// German umlauts are transliterated by the normalization. Returns nil for
+// empty / all-initials input — the caller should fall back to surname-only
+// behaviour.
 func authorTokens(author string) []string {
 	if author == "" {
 		return nil
 	}
 	var out []string
-	for _, w := range strings.Fields(strings.ToLower(author)) {
-		w = strings.ReplaceAll(w, "'", "")
-		w = strings.Trim(w, ".,;:()[]")
-		w = transliterateUmlauts(w)
+	for _, w := range strings.Fields(NormalizeRelease(author)) {
 		if len(w) >= 3 {
 			out = append(out, w)
 		}
--- a/internal/indexer/searcher_test.go
+++ b/internal/indexer/searcher_test.go
@@ -1224,17 +1224,77 @@ func TestAuthorMatchesReleaseSingleName(t *testing.T) {
 }
 
 // TestAuthorMatchesReleaseHyphenated covers hyphenated names like
-// "Mary-Kate Olsen" (#563). The hyphen is non-word so the regex \bmary-kate\b
-// matches "mary-kate" in the release, and "olsen" matches separately.
+// "Mary-Kate Olsen" (#563, #1608). The hyphen is a token boundary — the name
+// tokenizes to ["mary", "kate", "olsen"], matching both the hyphenated raw
+// form (the \W+ phrase separator covers "-") and the NormalizeRelease form
+// where the hyphen has collapsed to a space.
 func TestAuthorMatchesReleaseHyphenated(t *testing.T) {
 	toks := authorTokens("Mary-Kate Olsen")
 	if !authorMatchesRelease("mary-kate olsen biography epub", toks) {
 		t.Errorf("hyphenated name should match: tokens=%v", toks)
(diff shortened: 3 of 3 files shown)


helios57:
@vavallee When you have a moment, a review would be much appreciated — this one silently empties search results for dotted-initials authors (J.R.R. Tolkien being the flagship case), so it may explain other "no results" reports too. Happy to adjust anything.

Issue #1374, opened 2026-07-01
authorSortKey panics under concurrent author writes (shared transform.Chain is not goroutine-safe)


vavallee:
## Summary

`authorSortKey` (added in #1347 / v1.23.1) can **panic** under concurrent author writes:

```
panic: runtime error: slice bounds out of range [27:10]
golang.org/x/text/transform.String(...)  transform/transform.go:650
github.com/vavallee/bindery/internal/db.authorSortKey  internal/db/sortkey.go:36
```

Seen in CI on PR #1371 (`TestImporter_RollbackRemovesCreatedBatch`, an ABS-import test that creates authors from parallel goroutines). Flaky — the same suite passed on other runs.

## Root cause

`accentStripper` is a **package-level** `transform.Chain(norm.NFD, runes.Remove(...), norm.NFC)` (`internal/db/sortkey.go:48`). `transform.Chain` returns a `*chain` whose `Transform` method mutates internal link buffers — it is **not safe for concurrent use**. Any two goroutines writing authors at the same time (ABS import, parallel author work discovery, concurrent API calls) race on that shared state; the corruption surfaces as slice-bounds panics inside `x/text/transform`.

This ships in **v1.23.1 and v1.23.2**: a concurrent import can crash the server, and CI's `validate (Go)` will keep flaking until fixed.

## Fix

Construct the transformer chain per call instead of sharing one instance. Author writes are low-frequency; the extra allocations are noise. (`strings.NewReplacer` is documented concurrency-safe and can stay package-level.)

Regression test: hammer `authorSortKey` from many goroutines under `-race`.

PR #1611, opened 2026-07-23
fix(indexer): transliterate umlaut book-search queries, with original-spelling retry (#1610)

### LINKED ISSUE #1610: Indexer queries sent with literal umlauts while Usenet release names are ASCII-transliterated — German titles with ä/ö/ü return ~zero results and never auto-grab
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate (#1573/#1574 are about result-side language filtering; #47/#665/#713 touch the query tiers but not character handling. Nothing covers umlauts in outgoing queries.)
- [x] I'm running the latest version (observed on v1.26.2; the query-construction path is unchanged on current `main` / v1.27.0)

### Current behaviour

Bindery builds Newznab queries from the book title + author **verbatim**, umlauts included. Usenet release names are almost universally ASCII-transliterated (`Phoenix`, `Heiligtuemer`, `zurueck`), and the indexers we tested match queries close to literally. The result: **every German book with ä/ö/ü in its title fails to auto-grab; every umlaut-free German title succeeds.**

Measured on a live deployment (v1.26.2, NZBgeek and scenenzbs behind a Prowlarr proxy), same category set, same moment:

| Query (`t=search`) | Hits |
|---|---|
| `Harry Potter Orden des Phönix` | 2 |
| `Harry Potter Orden des Phoenix` | 9 — including all the retail EPUBs Bindery wanted |

The asymmetry is that the **result-matching side already speaks ASCII**: `SigWords` / `NormalizeRelease` transliterate umlauts (`transliterateUmlauts`), and `umlautFlexRegex` even matches the compact `Phonix` form. So if the transliterated releases ever *reached* the filter they would match fine — they just never come back, because all four `BookSearch` query tiers (`internal/indexer/newznab/client.go`) send the raw metadata form:

- Tier 1 `t=book` sends `title=` / `author=` verbatim (modulo subtitle/paren stripping in `NormalizeQueryTitle`, which does not touch diacritics).
- Tiers 2–4 `t=search` send `q=Surname Title` / `q=Author Title` / `q=Title` verbatim.

### Expected behaviour

Outgoing book-search queries should be sent in the alphabet release names actually use: German umlauts expanded to their two-letter forms (`ä→ae`, `ö→oe`, `ü→ue`, `ß→ss`) and remaining Latin diacritics folded to the base letter (`é→e`, `ú→u`), leaving non-Latin scripts (CJK, Cyrillic) untouched. With that, "Harry Potter und der Orden des Phönix" queries as "…Orden des Phoenix" and the wanted retail EPUBs are found and grabbed.

### Steps to reproduce

1. Add a German-language book whose title contains an umlaut (e.g. "Harry Potter und der Orden des Phönix") and one without (e.g. "Harry Potter und der Gefangene von Askaban"); both exist as retail EPUB releases on the tested indexers.
2. Trigger a search for both.
3. The umlaut-free title returns releases and grabs; the umlaut title returns (near-)zero results — while the same query with `ö→oe` typed into the indexer's own web search returns the full result set.

### Trace logs

Sanitized DEBUG excerpt from the live deployment (indexer URL and apikey redacted):

```text
DEBUG indexer query tier=1 url=…/api?t=book&title=Harry+Potter+und+der+Orden+des+Ph%C3%B6nix&author=J.+K.+Rowling&cat=7020&limit=100&apikey=***
DEBUG indexer query tier fallthrough tier=1 items=0
DEBUG indexer query url=…/api?t=search&q=Rowling+Harry+Potter+und+der+Orden+des+Ph%C3%B6nix&cat=7020&limit=100&apikey=***
DEBUG indexer query url=…/api?t=search&q=J.+K.+Rowling+Harry+Potter+und+der+Orden+des+Ph%C3%B6nix&cat=7020&limit=100&apikey=***
DEBUG indexer query tier 4 (title only) title="Harry Potter und der Orden des Phönix"
INFO  indexer returned results indexer=… count=0
```

Replaying the tier-4 query by hand with `Ph%C3%B6nix` replaced by `Phoenix` returns 9 results including the wanted retail EPUBs.

### Deployment method

Helm / Kubernetes

### Bindery version

v1.26.2 (query construction unchanged at v1.27.0 / current `main`)

### Environment details

Two Usenet indexers: NZBgeek, and scenenzbs via a Prowlarr proxy. Both match query terms literally enough that `Phönix` ≠ `Phoenix`. German metadata from OpenLibrary/DNB.

### Proposed fix

Transliterate the query at the single choke point where `BookSearch` builds its tiers: expand German umlauts to two-letter forms first (so a generic diacritic fold can't collapse `ä` to `a` — the release side expects `ae`), then strip remaining combining marks via the `x/text` NFD → remove-Mn → NFC chain already used elsewhere in the codebase (`internal/db` sort keys, per-call construction per #1374). Free-text user searches (`SearchQuery`) stay untouched — the user controls that string.

Known tradeoff: a release that keeps a literal `Phönix` in its name would no longer be found by the transliterated query (2 of the 11 combined hits above). Those are a small minority and the alternative — sending both forms and merging — doubles indexer traffic for every query with a diacritic; if you'd prefer that design I'm happy to rework. PR with tests incoming; will link it here.

helios57:
## Summary

Closes #1610.

**Problem** — `BookSearch` sends every query tier with the raw metadata title/author, umlauts included. Usenet release names are ASCII-transliterated (`Phoenix`, `Heiligtuemer`) and indexers match query terms close to literally, so every German title containing ä/ö/ü returns (near-)zero results and never auto-grabs. Measured live (v1.26.2, NZBgeek + scenenzbs behind a Prowlarr proxy): `q=Harry Potter Orden des Phönix` → 2 hits; `q=Harry Potter Orden des Phoenix` → 9 hits including all the wanted retail EPUBs.

**Root cause** — the result-*matching* side already speaks ASCII (`SigWords`/`NormalizeRelease` transliterate, `umlautFlexRegex` even accepts the compact `Phonix` form), but the query-*construction* side (`BookSearch` tiers 1–4 in `internal/indexer/newznab/client.go`) never did, so the matchable releases were simply never returned by the indexer.

**Fix** — transliterate at the single choke point where `BookSearch` builds its tiers:

1. German umlauts expand to two-letter forms first (`ä→ae`, `ö→oe`, `ü→ue`, `ß→ss`, both cases) — a blind diacritic fold would collapse `ä` to `a`, but the release convention is `ae`.
2. Remaining combining marks are folded off **Latin base letters only** (`é→e`, `ú→u`, `ñ→n`, `ç→c`) via NFD → strip-Mn-on-Latin-base → NFC. Marks on non-Latin bases are kept — a blanket Mn-strip (as the `internal/db` sort-key folder legitimately uses for ordering) would corrupt Cyrillic (`й` is `и` + combining breve) and Greek (tonos) queries.

Free-text `SearchQuery` is deliberately untouched — the user controls that string and can type either form. Uses `norm.Form` values directly (stateless, concurrency-safe), not a shared `transform.Chain` (#1374).

Known tradeoff (also in the issue): a release keeping a literal `Phönix` in its name is no longer matched by the query. Those were 2 of the 11 combined hits in the live measurement, and none were the wanted releases. The alternative — querying both forms and merging — doubles indexer traffic for every diacritic query; happy to rework to that design if preferred.

## Checklist

- [x] Tests added or updated

## Test plan

- New `TestTransliterateQuery` — umlaut expansion (both cases, ß), Latin diacritic folding (é/ú/ñ/ç), and pass-through guarantees for ASCII, CJK, Cyrillic (`й` must survive), and Greek
- New `TestBookSearch_TransliteratesUmlautQueries` — end-to-end via httptest: all four query tiers for title "Harry Potter und der Orden des Phönix" / author "Jürgen Müller" carry exactly the transliterated `title=`/`author=`/`q=` params and no raw umlaut reaches the wire
- Existing `TestBookSearch_*` tier tests, `TestSigWords`, `TestNormalizeQueryTitle`, and the umlaut-flex matching tests stay green


helios57:
@vavallee Review appreciated on this one as well — non-ASCII titles (umlauts/diacritics) currently return near-zero indexer results; the fix transliterates at the BookSearch choke point only, with the scoping rationale in the PR description. Glad to rework the approach if you'd prefer the both-forms-merge variant mentioned in #1610.

vavallee (changes requested):
Thanks for this, and for the writeup in #1610. The diagnosis is right and the German core of the fix is the correct shape: doing the two letter expansion before the generic mark strip is exactly right, and putting it at query construction rather than on the match side is the right layer. The Cyrillic and Greek carve out in `foldLatinDiacritics` is a genuinely nice touch.

Three things I'd like sorted before this goes in.

**1. The non German half of the fold is a net regression.**

The comment in `client.go` says the relevance filter accepts both forms so this only widens recall. That holds for ä/ö/ü/ß, because both sides transliterate those. It does not hold for é/ñ/á/ç, because nothing on the match side folds them. `SearchBook` passes the original `c.Title` into `filterRelevant` (searcher.go:199), and `NormalizeRelease` only calls `transliterateUmlauts`, which is German only.

So for a Spanish or French title we now ask the indexer for the ASCII form and then throw away every ASCII release that comes back, while no longer querying the diacritic form that previously did match:

```
meta="Señor Núñez"  query="Senor Nunez"  release="Senor.Nunez.Retail.epub"
  keywords stay [señor núñez], allMatch=false, result discarded
```

That's strictly worse than today for those languages. Either fold both sides symmetrically, or scope this PR to the umlaut replacer and drop `foldLatinDiacritics` entirely. I'd take the second, it's the smaller change and it fixes the issue you actually filed.

**2. Decomposed input silently takes the naive path.**

`transliterateQuery` runs the replacer on the raw string, so if a title arrives NFD (`o` + U+0308) the precomposed `"ö"` key never matches and the mark strip then gives you `Phonix` instead of `Phoenix`. That's the exact collapse the doc comment says the ordering prevents. Titles come from DNB, OpenLibrary, Calibre and ABS, so this isn't hypothetical, and `dedup.go:33` already normalises to NFC right before `transliterateUmlauts` for the same reason.

```go
return foldLatinDiacritics(queryUmlautReplacer.Replace(norm.NFC.String(s)))
```

Worth an NFD test case too. The current suite only feeds precomposed input, so it goes green with the bug live.

**3. Keeping the original form is cheaper than you think.**

You mention losing 2 of 11 hits and argue that keeping both doubles indexer traffic. `BookSearch` is a sequential cascade that only falls through on zero results, so appending an original form retry after tier 4 costs nothing whenever the transliterated query works. Gate it on `transliterateQuery(x) != x` and non diacritic titles never pay at all. That gets your 2 hits back and also defuses point 1 for accented titles.

Smaller stuff, non blocking:

* The fold only handles decomposable characters, so `ø`, `æ`, `Ł`, `Đ` pass straight through. `"Snømannen"` stays as is. The doc comment claims more than the code does.
* An all ASCII early return in `foldLatinDiacritics` makes the common path zero alloc.
* `queryUmlautReplacer` sits right next to `transliterateUmlauts` doing near identical work with a different case contract. A cross reference comment would save the next reader some time.
* All the tests are in the `newznab` package, but the regression in point 1 is a cross package interaction with `internal/indexer.filterRelevant`. A test at the `internal/indexer` layer asserting that what the transliterated query finds still survives `filterRelevant` would have caught it, and that's the acceptance criterion I'd want on the fix.

Happy to take this as two PRs if you'd rather land the umlaut part now and look at the general diacritic story separately. The umlaut half on its own is close to mergeable.

helios57:
@vavallee rebased onto current `main` and addressed the review. Since the umlaut logic now lives in the shared `internal/textutil` package on `main`, `newznab/words.go` reconciles with that rather than re-defining it.

I took your **option 2** — scope the PR to the umlaut replacer and drop the general diacritic fold — since that's the change that fixes the issue I actually filed.

**1. Non-German fold was a net regression → dropped it.**
`foldLatinDiacritics` is gone. `transliterateQuery` now expands German umlauts only. I re-confirmed the asymmetry still exists on current `main`: `textutil.FoldForTitleMatch` expands ä→ae but leaves é/ñ/ç as-is, so a query-side fold of those would still ask for `Senor` and then discard every `Senor` release. Umlaut-only keeps the query and match sides in one alphabet. `TestTransliterateQuery` now pins é/ñ/ç as pass-through.

**2. Decomposed input took the naive path → NFC first.**
`transliterateQuery` now composes with `norm.NFC` before the replacer, so a decomposed `ö` (o + U+0308, which macOS/ABS emit) transliterates instead of querying a bare `o`. Added an explicit NFD case to the table test (it was all-precomposed before, so it went green with the bug live).

**3. Keeping the original form is cheap → added the retry.**
`BookSearch` runs the four-tier cascade with the transliterated spelling, and if that comes up empty **and** transliteration actually changed the query, it runs the cascade once more with the original umlaut spelling — gated on `queryTitle != origTitle || queryAuthor != origAuthor`, so non-umlaut titles (and every non-Latin script) never pay for a second pass. That recovers the ~2/11 releases that keep the literal umlaut. Cascade extracted into `bookSearchTiers` for the two passes.

**Smaller items:**
- Cross-reference comment added on `queryUmlautReplacer` explaining the case-contract difference vs `textutil.TransliterateUmlauts` (both-case, no lowercasing, because the outgoing query isn't lowercased).
- The `ø/æ/Ł/Đ`-pass-through and all-ASCII-early-return notes are moot now that the generic fold is gone.
- Added the acceptance test you asked for at the `internal/indexer` layer: `TestFilterRelevantAcceptsTransliteratedUmlautRelease` asserts a transliterated query's ASCII release survives `filterRelevant`.

`go test ./internal/indexer/... -race` green; `go build ./...`, `go vet ./internal/indexer/...`, `gofmt` clean. Changelog fragment updated to match the umlaut-only + retry behaviour.

vavallee:
Apologies for the wait on this — my `CHANGES_REQUESTED` from 26 July was submitted against `2517c924` and you revised on 4 August, so the block has been stale for weeks. I have re-reviewed the current head (`a68712fa`) and my original concern is properly addressed.

Verified rather than assumed:

- **French/Spanish accents are deliberately untouched, which is correct.** `FoldForTitleMatch` does not fold `é ñ ç`, so transliterating query-side only would ask for "Senor" and then discard every "Senor" release. Symmetry holds. Same reasoning covers `ø æ ł đ` passing through, since `FoldForTitleMatch` does not call `FoldNonDecomposableLatin`.
- **Proper German convention, not accent-stripping** — `ä→ae ö→oe ü→ue ß→ss` plus the uppercase forms. I property-tested `transliterateQuery` over a 15-title corpus in both NFC and NFD: unicode-form-stable and idempotent, zero drift. Cyrillic, CJK, Greek, Korean and Devanagari untouched.
- **The retry is conditional, not unconditional.** Gated on `len(results) == 0 && query != original`. I instrumented an httptest server and counted real requests: umlaut title that finds nothing anywhere is 8 instead of 4 per indexer; a plain ASCII miss stays at 4; NFC *and* NFD Spanish/Korean/Cyrillic misses never retry. I had suspected the gate would misfire on NFD input since `origTitle` is not NFC-normalised at the call site, but `NormalizeQueryTitle` already composes at `client.go:412`, so it holds.
- Merges cleanly onto current main (27 commits on), `go build` / `go vet` / `go test ./internal/indexer/...` all green. No new dependency — `golang.org/x/text` is already direct. No new network hosts, no auth or CI changes.

Not merging it into the release going out now, only because I am keeping that one tight. It is next, and none of the below is a blocker.

**Two bits of project-convention debt worth doing while it is open:**

1. **Register the transliteration in `internal/normdrift`.** That package exists precisely for this bug class (#1642/#1646/#1647), its `stringFolds` registry already includes `newznab.NormalizeQueryTitle`, and its doc says adding one there is how you opt into the properties. `transliterateQuery` is unexported so it would need exporting first. I confirmed it *would* pass both properties.
2. **Update the four-alphabet doc block in `internal/textutil/fold.go`.** That comment declares itself the single home for "FOUR distinct comparison alphabets… so the next person can tell which one they want". This adds a fifth reduction elsewhere without updating it.

**One accuracy fix:** the commit message and changelog fragment say the retry "recovers the ~2/11 releases that keep the literal umlaut". My measurements say that is not what happens in the case you sampled — `Phoenix` returns 9 hits, so the retry never fires and those 2 stay lost. The retry only helps in the all-or-nothing case. That makes the quota story *better* than advertised, but the claim should match.

**Optional:** a `BookSearch`-level test that an accented non-German title does not fire the retry. That invariant currently rests on two separate NFC normalisations; remove either and every accented title starts double-querying with nothing failing.

Also worth stripping the `Claude-Session:` trailer on squash — it appears zero times in the last 200 commits on main (`Co-authored-by: Claude` is normal and fine).

helios57:
Thanks for the thorough re-review. All four follow-ups are in (`438c926`), rebased onto current `main`:

1. **normdrift registration** — exported `TransliterateQuery` and added it to the `stringFolds` map. It passes both properties (the NFC-compose makes it Unicode-form-invariant; ASCII output makes it idempotent), and `TestNonLatinScriptsKeepDistinctKeys` confirms it doesn't collapse CJK/Cyrillic to a shared/empty key.
2. **`internal/textutil/fold.go` doc block** — added it as the fifth reduction, explicitly labelled a one-directional query reduction (*not* a comparison alphabet) that stays symmetric with alphabet (1), with a pointer to the normdrift registration. Bumped the intro and the "first four" line.
3. **Accuracy fix** — you're right, it's a zero-results fallback, not a merge. Reworded both the commit message and the changelog: the retry recovers releases only when the transliterated query finds nothing *at all*; when the ASCII form already returns hits (your Phoenix→9 sample), the retry never fires and any umlaut-only releases stay unfound. Also removed the stale "other Latin diacritics fold to their base letter" line in the changelog — leftover from the dropped generic fold, no longer true.
4. **No-retry test** — added `TestBookSearch_NoRetryForAccentedNonGermanTitle`, feeding a *decomposed* (NFD) `Señor Núñez` / `José García` and asserting exactly one 4-tier cascade. That's the case that regresses if either NFC normalisation (`NormalizeQueryTitle` or `TransliterateQuery`) is dropped — remove one and every accented title starts double-querying with nothing failing.

Also dropped the stray non-standard commit trailer you flagged.

vavallee (approved):
Clearing my own stale block. I said on 7 August that the original concern was properly addressed and that none of the follow-ups were blockers, but I wrote that as a comment rather than a review, so the July `CHANGES_REQUESTED` stayed on the record and kept this sitting in the blocked pile for another three days. That is on me, not on you.

Re-verified the current head before approving: rebases clean onto main, the four tier cascade still only falls through on zero results so the retry cannot fire on an auth failure or a timeout, an all ASCII title still issues exactly four requests on a total miss, and the suite fails correctly under mutation on all four properties (transliteration, the NFC compose, the retry, and the inequality gate). It also auto merges against both in flight branches that touch the same area.

The one red check is `SAST – Frontend`, which is failing on main itself and on the scheduled runs. Nothing to do with a Go only diff. The three required checks are green.

Thanks for the patience on this one, and for the measurements in #1610. Those made the whole thing straightforward to reason about.

Issue #1457, opened 2026-07-07
Multi-user tenancy: owner_user_id never stamped on create; series endpoints unscoped


vavallee:
Found during a codebase security sweep. This is the highest-impact finding and needs a design decision before implementation, so filing rather than PRing.

## Problem

The multi-user tenancy model has no write path for `owner_user_id` on several resources, and the read-scoping policy is inconsistent across resource tiers. Only matters when `BINDERY_ENFORCE_TENANCY` is on, so it is not shipping-broken today, but the multi-user feature is effectively broken for anyone who enables it.

### owner_user_id never stamped on create

`BookRepo.Create`/`Update` omit `owner_user_id` from the column list (`internal/db/books.go:442`), same for `DownloadRepo.Create` (`internal/db/downloads.go:74`), `RootFolderRepo` (`internal/db/root_folders.go:51`), and quality/metadata profiles. Only `AuthorRepo.CreateForUser` stamps an owner. The only writes to `books.owner_user_id` / `downloads.owner_user_id` anywhere are migration 025's one-time backfill to user 1. `multiuser_test.go:61` documents it: "Create uses no user scope yet."

Under `EnforceTenancy`:
- Every book created after migration 025 has NULL owner. Books lists treat NULL as world-visible (`QueryScopeForIncludingNull`), and `auth.CheckOwnership` receives 0, so any user can search / unmonitor / blocklist / delete any other user's post-migration books (`internal/api/bulk.go:450`).
- Downloads use strict scoping (`downloads.go:33`), and since no download ever gets an owner, a non-admin sees an empty queue including their own grabs (`internal/api/queue.go:135`).

### Series endpoints have no owner scoping at all

`SeriesRepo` has no `ForUser` variants; `ListWithBooks` joins `books` with no owner predicate (`internal/db/series.go:50`), and `SeriesHandler.List`/`Get` call it unscoped (`internal/api/series.go:162`). Any non-admin can enumerate every series and every other user's book titles, authors, statuses, and covers. Same class as the leak #1210 fixed for the series-diff view, but the main list/detail endpoints were never patched.

### Inconsistent NULL-owner semantics across tiers

Books/authors treat NULL owner as visible (`scope.go:40`); downloads use strict equality; history/pending scope through `book_id IN (SELECT ... WHERE owner_user_id = ?)`, which hides events for NULL-owner books the library list does show.

## Design decision needed

Pick one NULL-owner policy (strict vs include-NULL) and apply it consistently per tier, then:
[...]

Issue #891, opened 2026-06-01
OpenLibrary edition-sampling for work-level language so the allowed_languages filter catches the OL tail


vavallee:
## Problem

The Hardcover half of #889 is being closed by #890 (canonical-book `language` now flows through the supplemental query). The OpenLibrary half is still open: OL's `/authors/{id}/works.json` carries no language at all, and the `/search.json` enricher only backfills `language` for works it indexes. For prolific authors, a non-trivial tail of works (often translations) are absent from the search index and arrive at the filter with `Language=""`, falling through the default `unknown_language_behavior=pass` branch (`internal/api/authors.go:1182-1190`, `internal/models/language.go:41-50`).

Net effect: even after #890 lands, OL-primary setups (no Hardcover token) keep seeing foreign-language editions in author bibliographies when the active metadata profile restricts language.

This is the same root cause #224 first identified and #232 partially mitigated (added the `unknown_language_behavior` setting). Neither closed the gap; they just gave users a lossy switch.

## Proposed fix (issue #224 plan A, revisited)

Edition-sample the language when the work has no language after `/search.json` enrichment:

1. In `openlibrary.Client.GetAuthorWorks`, for each work that ends up with `Language == \"\"`, fetch `/works/{id}/editions.json?limit=5` and take the majority language (or first non-empty as cheaper fallback).
2. Cache the result keyed on work ID so refresh passes don't re-fetch.
3. Gate the extra fan-out on the active profile actually having an `allowed_languages` set — no point spending the round-trips when the filter is `any`.
4. Surface the count in the existing `slog.Debug(\"skipping non-allowed-language book\", ...)` line so we can confirm it's biting in production logs.

## Out of scope (deferred or rejected)

- Title-character heuristics (#224 plan C): too brittle to lean on as the primary mechanism. Could be a last-resort tie-breaker but not worth the maintenance until the edition-sampling layer is in place.
- Changing the default `unknown_language_behavior` to `fail`: breaks every user whose existing library accumulated empty-language English books pre-fix. Not worth the support burden.

## Related

- #889 — user report (Spanish editions of Connelly/Patterson/Gladwell leaking)
- #890 — Hardcover half of the fix
- #224 — original root-cause analysis
- #232 — `unknown_language_behavior` switch (mitigation)

Issue #804, opened 2026-05-24
AddBook can delete a new author before async sync inserts the selected book


magrhino:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings -> About)

Note: this was observed in a dev-container/source-build run. The runtime log reports `version=dev` and `commit=unknown`; the issue is being filed against `main`. I found closed issue #285 with a SQLite foreign-key error, but that was in the download/grab path, not the AddBook author-sync path shown here.

### Current behaviour

Trying to add `Twilight` by Stephenie Meyer through Add Book fails to create the selected book. The logs show Bindery starts fetching Stephenie Meyer's author works, then `AddBook` cleans up the newly created author as an orphan before the background author sync has inserted any books. The still-running background sync then tries to insert `Twilight` and the rest of the author's catalogue using the deleted author ID, so every insert fails with SQLite foreign-key errors.

Observed result from the trace:

- `AddBook` starts author sync for `Stephenie Meyer` / `OL1391085A`.
- 15 seconds later, orphan cleanup deletes `authorId=48` because the requested book has not appeared yet.
- The async sync then tries to create `Twilight` and many related works, each failing with `create book: constraint failed: FOREIGN KEY constraint failed (787)`.
- The sync reports `added=0` out of `total=175`.

Local DB inspection after the run showed no Stephenie Meyer author row and no `Twilight` book row; `PRAGMA foreign_key_check` was clean because the invalid book inserts were rejected.

Suspected code path:

- `internal/api/authors.go`: `AddBook` creates an unmonitored author and calls `fetchAuthorBooksAsync`.
- `fetchAuthorBooksAsync` snapshots the author, including its DB ID, before launching the goroutine.
- `AddBook` polls for 15 seconds, then returns `book not found after author sync` if the selected book is not inserted yet.
- The deferred orphan cleanup deletes the just-created author when no books exist yet.
- The background sync continues with the stale author snapshot and inserts books with the now-deleted `author_id`.

### Expected behaviour

Adding a single selected book should either:

- create the requested book and keep/link the author, or
- fail without leaving the in-flight author sync writing against a deleted author ID.

The orphan cleanup should not delete a newly created author while its async author-work sync is still running, or the AddBook path should directly insert the selected book instead of waiting for the full author sync to happen first.

### Steps to reproduce

1. Run a source/dev-container build with trace/debug logging enabled.
2. Open Add Book.
3. Search for `Twilight`.
[...]

Issue #667, opened 2026-05-15
feedback about DNB + more issues


zippoking:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Hey,

unfortunately i have to give a bit negative feedback. It has been over a month (#67) that i asked for DNB support. Since then and till today i am not able to add a SINGLE BOOK to bindery. Everytime something is broken. 

First DNB was only added as an enricher, which is fine for the beginning. As it turned out, OL cannot provide the data for DACH-region, so it was necessary to have DNB as a primary metadata provider. So we had issues with the isbn search in #284 then it was not possible to add books, because the button was greyed out #545 , right after that another popup with some erros messages comes up #608 and now another window pops up with an error message, that i havn´t reported yet (see below).

You push another version, i test the new version and i still cannot add a single book from DNB. 
It takes 5 seconds, literally 5 seconds, to grab one of my ISBNs that i provide, put it into the search and you would see, that your bugfixes are not sufficient. 

For me it looks like you are not even testing your own bugfixes. 

So my questions is.....what are we doing here? I want to use bindery in the future, i see your quality work in development, but the quality management of bugfixes/testing etc. is ridiculous. You are closing the issues super fast, you provide no prove that bugs have been fixed and as i already said, it looks like you are not even testing your bug fixes...

In the end it is your software, you can do whatever you want with it, but i am really frustrated at the moment.

I would love to hear your feedback, maybe i am totally in the wrong here. Please let me know.
________________

So here are multiples bugs i want ro report and unless i get some positive feedback from you, that you personally tested DNB intregration (provide screenshots that it works or whatever and i dare you to close this issue just with saying fixed in a new version!), i have to withdraw from this project.

_____________________________________________________

Here are the same ISBNs i provided in #608:
978-3-8373-9555-6
978-87-27-26711-1
978-3-8449-4284-2
978-3-7324-7563-6
978-3-96092-698-6
978-3-8371-6512-8
978-3-8449-3577-6
978-3-9597-2586-6
978-3-86717-354-4
978-3-8371-5566-2
978-3-8449-0704-9
978-3-8449-4504-1

### 1. Error message doesn´t let me add books:
<img width="1261" height="840" alt="Image" src="https://github.com/user-attachments/assets/ea49911d-50ed-4ae0-a7ff-b8e04048b176" />

[...]

PR #1613, opened 2026-07-23
fix(api): direct-insert explicitly added books for existing authors too (#1612)

### LINKED ISSUE #1612: AddBook 404s forever for specific OpenLibrary works when the author already exists (e.g. OL82537W, Chamber of Secrets) — poll waits for a row the catalogue sync refuses to create
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate (#804 fixed the just-created-author case with a synchronous direct insert; #1405/#1415 fixed re-linking of rows attached to a different author record. This is the remaining case: author already exists, the work's row simply never comes into existence, and AddBook has no fallback.)
- [x] I'm running the latest version (observed on v1.26.2; the AddBook flow described below is unchanged on current `main` / v1.27.0)

### Current behaviour

A specific OpenLibrary work can be permanently un-addable: `POST /api/v1/author/book` returns `404 {"error":"book not found after author sync — try again shortly"}` on **every** attempt — 8+ retries over 2 hours in our case — while sibling works of the same author add fine (some after 1–2 retries).

Concrete case, J. K. Rowling (`OL23919A`, ~500+ works, author already present in the library):

- `OL82537W` — *Harry Potter and the Chamber of Secrets* — **404 on every attempt, never succeeds**
- Siblings `OL82563W`, `OL82536W`, `OL82560W`, `OL82548W`, `OL82565W`, `OL82586W` — all succeeded

Meanwhile the metadata layer resolves the work fine: `GET /api/v1/book/lookup?isbn=9780747538493` returns full `OL82537W` metadata. The add still 404s immediately afterwards.

### Expected behaviour

An explicit user add of a specific work that the metadata provider can resolve (as the ISBN lookup proves) should create that book — or fail with a diagnosable error saying *why* the catalogue sync will never surface it. "Try again shortly" is wrong when no number of retries can ever succeed.

### Steps to reproduce

Against a library that already contains J. K. Rowling (`OL23919A`):

```bash
curl -s -X POST http://<bindery>/api/v1/author/book \
  -H 'Content-Type: application/json' \
  -d '{"foreignBookId":"OL82537W","foreignAuthorId":"OL23919A","authorName":"J. K. Rowling","searchOnAdd":false,"mediaType":"ebook"}'
# → 404 {"error":"book not found after author sync — try again shortly"}   (every time)

curl -s 'http://<bindery>/api/v1/book/lookup?isbn=9780747538493'
# → 200, resolves to OL82537W with full metadata

# repeat the POST → 404 again, on every retry, indefinitely
```

Swap `OL82537W` for `OL82563W` (Prisoner of Azkaban) and the same request succeeds.

### Trace logs

Sanitized API transcript from the live deployment (v1.26.2, Helm/Kubernetes):

```text
POST /api/v1/author/book {"foreignBookId":"OL82537W","foreignAuthorId":"OL23919A",...}  → 404 {"error":"book not found after author sync — try again shortly"}   (attempts 1..8+, spread over ~2 h)
POST /api/v1/author/book {"foreignBookId":"OL82563W",...}                               → 201 (2nd attempt)
POST /api/v1/author/book {"foreignBookId":"OL82536W",...}                               → 201
GET  /api/v1/book/lookup?isbn=9780747538493                                             → 200 (OL82537W metadata, correct title/author)
POST /api/v1/author/book {"foreignBookId":"OL82537W",...}                               → 404 (immediately after the successful lookup)
```

We were unable to capture the author-sync DEBUG window for the failing work in production; the relevant line to look for during a Rowling sync would be `skipping non-allowed-language book`/`skipping junk-title OL work` (`internal/api/authors.go`) naming `OL82537W`.

### Deployment method

Helm / Kubernetes

### Bindery version

v1.26.2

### Environment details

Metadata provider OpenLibrary; author added long before, catalogue previously synced. Book adds issued via the API (same payloads the UI sends).

### Analysis / hypothesis for the root cause

Reading the AddBook flow (`internal/api/authors.go`), when the **author already exists** the request:

1. takes none of the direct-insert paths — `directInsertNeeded` is only true for `authorWasJustCreated || authorMatchedByAlternateID || dnb: || resolvedByName` (#804/#667 fixes), and
2. does not itself trigger a catalogue sync; it only polls `GetByForeignID(foreignBookId)` for 15 s and 404s if no row appears.

So for an existing author, the add succeeds only if the async/scheduled catalogue sync has (or eventually will have) created the row. If the sync **refuses to create this work**, the add can never succeed, no matter how often it is retried — which is exactly the observed behaviour. Candidate reasons the sync skips `OL82537W` specifically:

- **Edition-sampled work language** (leading hypothesis): OpenLibrary works carry no language, so the sync derives one by sampling the first `editionLangSampleCap = 5` editions (#891). The most-translated works — and Chamber of Secrets is among the most-translated books in existence — can easily present 5 non-English editions first, deriving a language outside the profile's allowed set → `skipping non-allowed-language book` → row never created. This would also explain why it hits a famous work while obscure siblings sail through, and why it is deterministic across retries (the derived language is cached per work for the process lifetime).
- Junk-title filter (`skipping junk-title OL work`) — unlikely here, the work is properly titled.
- Row existing under a different author record (#1405) — the #1415 re-link should now cover that; listed for completeness.

Suggested fix direction: when the author already exists and the poll is about to 404 (or upfront), fall back to the same synchronous `h.meta.GetBook(foreignBookId)` direct insert that #804 added for just-created authors. An explicit, single-work user add is the strongest possible signal and should override catalogue-sync heuristics (language sampling, junk filters) — the user is looking at the exact work they want. Failing that, the 404 should at least surface the skip reason instead of advising a retry that can never succeed.

helios57:
## Summary

Fixes #1612.

**Problem** — a specific OpenLibrary work can be permanently un-addable when its author already exists in the library: `POST /api/v1/author/book` returns `404 "book not found after author sync — try again shortly"` on every attempt (8+ retries over 2 hours in the live report for `OL82537W`, *Harry Potter and the Chamber of Secrets*), while `GET /api/v1/book/lookup?isbn=…` resolves the same work fine and sibling works add normally.

**Root cause** — for an existing author, AddBook took none of the direct-insert paths (`directInsertNeeded` covered just-created / alternate-ID / `dnb:` / name-resolved only, per #667/#804) and triggered no sync: it only polled `GetByForeignID` for 15 s. The add could therefore only succeed if a catalogue sync had already created the row — and the sync can *deterministically* refuse a work (leading candidate: the work-level language sampled from the first `editionLangSampleCap = 5` editions of a heavily-translated work lands outside the profile's allowed set, #891). "Try again shortly" was advice that could never help.

**Fix** — make the #804 synchronous direct insert unconditional. The block already no-ops when the row exists, so the common existing-author path is unchanged (one cheap `GetByForeignID` the poll would have issued anyway); when the row is missing, the explicitly requested work is fetched via `h.meta.GetBook` and persisted before the poll, which then succeeds on its first iteration. Rationale: an explicit add of one specific work is the strongest possible user signal and must not be vetoed by catalogue-sync heuristics — which is already the established behaviour for just-created authors (#804 direct-inserts before any filter can run). Sync filters still govern everything the user did **not** explicitly pick. The now-purposeless `authorMatchedByAlternateID` gate variable is removed with the gate.

**Why not fix the language sampling instead** — I considered attacking the 5-edition sample mis-derivation directly, but (a) it is a heuristic-quality problem with no crisp correctness criterion (any sample size loses to a sufficiently translated work), (b) the deployment evidence identifies it only as the most probable of several possible skip reasons, and (c) even a perfect sample leaves the structural defect that an explicit add depends on catalogue-sync heuristics at all. The direct-insert extension fixes the whole failure class in one small, behaviour-preserving change consistent with the #804 precedent; sampling quality can be tuned independently if desired.

## Checklist

- [x] Tests added or updated

## Test plan

- New `TestAddBook_ExistingAuthorDirectInsertWhenSyncSkipsWork` — mirrors the bug report end to end: existing author, prior catalogue sync where the requested work is language-skipped (`added=1 skipped_language=1 skipped_junk=1`), then an explicit add. Asserts 201 (was 404 after the full 15 s poll before the fix — verified by running the test against unpatched `authors.go`), row linked to the existing author, `Monitored=true`, explicit media type wins, owner inherited (#1457)
- Same test pins the negative behaviour: the junk-titled and language-filtered works the user did *not* request remain skipped after the add, and no duplicate author row appears
- Existing guards stay green: `TestAddBook_DirectInsertCoversSlowAsyncSync` (#804), `TestAddBook_OrphanAuthorDeletedOnTimeout` (#667), `TestAddBook_DNBDirectInsertSucceeds`, tenancy suite (#1457/#947)


helios57:
@vavallee One more review when you get the chance — this closes the "book not found after author sync — try again shortly" dead end for works the catalogue sync deterministically skips (existing authors had no direct-insert fallback, so retrying could never help). If you would rather gate the direct insert differently, or attack the edition-language sampling instead, happy to rework.

vavallee (changes requested):
The analysis in #1612 is excellent, thank you. I confirmed the root cause independently: for an author that already exists, `fetchAuthorBooksAsync` only fires inside the just created branch (`authors.go:2362`), so "existing author plus missing book row" is a permanent, deterministic 404. Your fix direction is right, and the old `directInsertNeeded` OR really is a strict subset of unconditional, so nothing loses coverage.

One blocking problem before I can take it.

**The direct insert skips the normalized title dedup that every sync created book goes through** (`authors.go:1702`), and doesn't use the helper that exists for exactly this, `BookRepo.FindByAuthorAndDedupKey` (`internal/db/books.go:796`). Tested against your branch:

```
status=201 rows=2
  fid=calibre:book:7  title="Harry Potter and the Chamber of Secrets"
  fid=OL82537W        title="Harry Potter and the Chamber of Secrets"
```

On main that request is a 404 with one row. With the patch it's a 201 with two. There's no unique constraint problem (`books.foreign_id` is globally unique and the collision gets swallowed cleanly), it's a title level duplicate. Two ordinary ways to hit it:

* Calibre imported library, where the existing row carries a `calibre:` foreign ID. The sync would have upgraded that stub in place, the direct insert duplicates it.
* OpenLibrary's split ebook and audiobook Work records. The sync merges those into `media_type=both` and continues, so the second Work's foreign ID row never exists. Someone clicking add on that Work is hitting the exact bug you're fixing, and now gets a duplicate instead of a 404. Search results have no "already in library" marker, so the UI can't head it off.

The reason it's more than cosmetic is that the direct insert also skips `handleNewWantedBook` (`authors.go:1788`), so the new row gets no LibraryFinder check and no series linking. With `searchOnAdd` enabled bindery will go grab an ebook the user already has on disk, which is what #940 and migration 026 exist to prevent.

What I'd like:

1. Before `h.books.Create`, call `FindByAuthorAndDedupKey(ctx, author.ID, primary.Title)` and reuse or upgrade the match rather than inserting. Calibre stub adopts the foreign ID so the poll finds it, ebook plus audiobook becomes `media_type=both`.
2. Call `handleNewWantedBook` on the success path, the way `recommendations.go:239` and `series.go:1408` do.
3. Skip the insert when `strings.TrimSpace(primary.Title)` is empty or equals the author name. Right now a provider record with an empty title persists and gives you destination folders like `Jared M. Diamond/Jared M. Diamond ()`. The sync blocks this at `authors.go:1615` for that reason.

On tenancy, the ownership handling itself is correct: `primary.OwnerUserID = author.OwnerUserID` off a properly scoped `GetByForeignIDForUser`, matching the sync's own rule. But the test assertion for it is vacuous, `got.OwnerUserID != author.OwnerUserID` is comparing 0 to 0 since no user is created and no `auth.WithUserID` context is set. Since this is a project convention I'd like it actually pinned, there's a ready pattern at `authors_test.go:3632`.

Smaller things, none blocking:

* Worth calling out in the description that this also bypasses the strict media type clamp from #1575, not just the language filter. An ebook only user adding an audiobook only work with `mediaType` omitted now gets an ungrabbable row. Probably fine under "explicit user signal wins", but it shouldn't fall out silently.
* The comment about inheriting the owner "stamped from the requesting user when this request created the author" is stale now, since the author usually pre exists.
* The stub hands the handler back the same `*models.Book` pointer it then mutates. Harmless with one call, fragile with two.
* Changelog fragment is a full paragraph, release notes want one line.

The rest is genuinely good work. `TestAddBook_ExistingAuthorDirectInsertWhenSyncSkipsWork` running a real `FetchAuthorBooks` so the skip comes from the actual language filter, rather than stubbing it, is the right instinct, and the two race tests are deterministic with no sleeps. Add the dedup guard and this is a merge.

vavallee:
Superseded by #1745, which is this branch rebased onto current main with the three review items resolved on top. Your commits and tests went in as is, and the merge commit is fab002e5. Closing this one as merged-by-proxy rather than rejected. Thanks @helios57.

Issue #1642, opened 2026-07-26
Non-ASCII author/title tokens can never match: Go's \b is ASCII-only, so accented and non-Latin names return zero results


vavallee:
Found while sweeping for other instances of the #1608 bug class after #1609 merged. This is the same shape (two sides of a comparison reduced through different alphabets) but the cause is one level down, and it silently zeroes out entire language populations.

## Root cause

Go's `regexp` is RE2, where `\b` is defined against **ASCII** `\w`, i.e. `[0-9A-Za-z_]`. Meanwhile `NormalizeRelease` deliberately preserves `\p{L}` and `\p{N}`, and its doc comment says so explicitly: *"Unicode letters (\p{L}) and numbers (\p{N}) are preserved so CJK and accented-Latin release names survive"*.

Those two facts are incompatible. Any keyword whose **first or last rune is non-ASCII** produces a regex that cannot match anything, ever, because the position next to a non-ASCII letter is never an ASCII word boundary.

Affected: `WordBoundaryRegex` (`internal/indexer/release.go:104`), `phraseRegex` (`release.go:119`), `inOrderRegex` (`internal/indexer/inorder.go:29`).

## Proof

Running the real `NormalizeRelease` + `authorTokens` + `WordBoundaryRegex`, matching each author against a release name built from that same author, so this is the most favourable case possible:

```
Stieg Larsson           tokens=[stieg larsson]              selfMatch=true
Heinrich Böll           tokens=[heinrich boell]              selfMatch=true
Åsa Larsson             tokens=[åsa larsson]                 selfMatch=FALSE  dead=[åsa]
José Saramago           tokens=[josé saramago]               selfMatch=FALSE  dead=[josé]
Émile Zola              tokens=[émile zola]                  selfMatch=FALSE  dead=[émile]
Jo Nesbø                tokens=[nesbø]                       selfMatch=FALSE  dead=[nesbø]
Gabriel García Márquez  tokens=[gabriel garcía márquez]      selfMatch=true
Fyodor Достоевский      tokens=[fyodor достоевский]          selfMatch=FALSE  dead=[достоевский]
```

The rule is that a diacritic **at a token edge** kills the token, while an interior one survives. That's why `garcía` is fine (ends in `a`) and `josé` is not. German is unaffected because `transliterateUmlauts` converts ä/ö/ü/ß to ASCII before the regex is built, which is almost certainly why this has gone unnoticed.

[...]

PR #1657, opened 2026-07-27
fix(importer): reconcile authors and titles through the shared folds (#1646)

### LINKED ISSUE #1646: Library-scan author matching re-implements #1608: run-together dotted initials and NFD folder names never reconcile
Found sweeping for other instances of the #1608 bug class after #1609. `authorMatch` (`internal/importer/scanner.go:1770`) has the same defect #1609 just fixed in the indexer, plus two more, and it is the author predicate for the library scan reconcile tier and for `findExistingInDir`.

The ASCII `\b` half is filed separately in #1642. This issue covers the other two.

## 1. Run together dotted initials produce tokens that can never match

`significantAuthorTokens` (`scanner.go:1789`) does `strings.Trim(w, ".,;:()[]'\"")`, which trims only the **ends**. So `"J.R.R."` becomes `j.r.r`, length 5, which survives the `>= 3` gate that is supposed to discard initials. The docstring on `authorMatch` (`scanner.go:1762`) claims *"Initials (1-2 char tokens like "R." in "George R. R. Martin") are dropped"*, but that only holds when the initials are space separated.

This is #1608 exactly, re-implemented in a second package.

Proof, real `ScanLibrary`, folder author versus catalogue author:

```
catalogue "J. R. R. Tolkien"    folder "J.R.R. Tolkien"      reconciled=FALSE  tokens=[j.r.r tolkien]
catalogue "George R. R. Martin" folder "George R.R. Martin"  reconciled=FALSE  tokens=[george r.r martin]
catalogue "J.R.R. Tolkien"      folder "J. R. R. Tolkien"    reconciled=true   tokens=[tolkien]
```

and `findExistingInDir` agrees:

```
findExistingInDir(root, "The Hobbit", "J. R. R. Tolkien")  -> ""       (folder is "J.R.R. Tolkien")
findExistingInDir(root, "The Hobbit", "J.R.R. Tolkien")    -> found
```

It's **directional**. It fires when the parsed side (the author folder name, or an ID3 `Artist` tag, and audiobook releases overwhelmingly write `J.R.R. Tolkien`) has run together initials while the catalogue has the spaced form, which is the form OpenLibrary and DNB return.

## 2. No Unicode normalization anywhere in internal/importer

`grep -rn "norm.NF"` finds hits in `internal/indexer/dedup.go:33`, `internal/seriesmatch/match.go:77`, `internal/metadata/aggregator_canonical.go:319`, `internal/textutil/author.go:15` and `internal/db/sortkey.go:54`, and **zero** in `internal/importer`.

macOS hands filenames back decomposed, while OpenLibrary/DNB/Calibre/ABS hand back precomposed. `authorMatch` compares byte literals, so they never meet:

```
catalogue "Björn Andersen" (NFC), folder NFD("Björn Andersen")  ->  reconciled=FALSE
```

That author matches itself fine in NFC, so this is an independent trigger from #1642.

## Why it goes all the way to unmatched

`matchingAuthors` (`scanner.go:2110`) verifies every index candidate with `authorMatch` and returns an **empty non-nil** set. The caller at `scanner.go:2296` treats non-nil as "restrict candidates to this set", so the candidate list is empty and the title tier never runs. The book isn't considered at all, and it stays unmatched across rescans.

## Why nobody noticed

The manual import and bulk folder import paths use `lookupAuthorMatch` (Jaro-Winkler) instead, which is tolerant of all of the above: it scores 0.929 / 0.978 / 0.950 on the dotted initial pairs and returns true for NFD versus NFC. So manual import matches these books and only the library scan doesn't. That asymmetry between the two author matchers is itself worth a look.

## Related, same family

`lookupAuthorMatch` has no transliteration, so `lookupAuthorMatch("Boell, Heinrich", "Heinrich Böll")` is false at JW 0.441. The indexer normalizer does ä→ae, the importer's does not.

`titleMatch`'s two reductions also disagree. `normalizeTitle` (`scanner.go:1642`) does lowercase plus article handling only, while `sigTokens` (`scanner.go:1691`) keeps only `[a-z0-9]` and treats every non ASCII rune as a separator with a `>= 2` floor. Neither strips apostrophes or folds diacritics, so against real `lookupWith`:

```
file "Enders Game.epub"  -> none   bt=[ender game]   pt=[enders game]
file "Die Hoehle.epub"   -> none   bt=[die hle]      pt=[die hoehle]
file "Der Prozess.epub"  -> none   bt=[der proze]    pt=[der prozess]
```

`"Die Höhle"` reducing to `[die hle]` is worth staring at: the `ö` deletes the `h` before it, because the leftover one rune fragment falls under the floor.

## Suggested fix shape

One change kills #1642's importer half plus both defects here: route `authorMatch` through `textutil.NormalizeAuthorName` (`internal/textutil/author.go:14`), which already does NFD, strips `Mn` combining marks, keeps letters and digits only, lowercases and collapses whitespace. `NormalizeAuthorNameWithVariants` already handles compacted versus expanded initials and last-first inversion. Normalize **both** sides and match on token subset. `wordRunsInto` and `firstWordRun` would need rebuilding over the same normalized form to keep the index a superset.

The word run index itself is **not** at fault. I checked its superset guarantee against hyphens, apostrophes, dotted initials, leading diacritics and all Cyrillic, and it holds every time. Where the index and `authorMatch` disagree it is `authorMatch` that is wrong.

vavallee:
Second of three. **Stacked on #1656** — review that first, this PR retargets to `main` once it merges.

## The author half

`authorMatch` (`scanner.go`) is the author predicate for the library-scan reconcile tier and for `findExistingInDir`. It had three independent ways to never match:

1. **Run-together dotted initials.** `significantAuthorTokens` trimmed punctuation only at the **ends** of a whitespace-split token, so `"J.R.R."` survived as `j.r.r` — 5 characters, clearing the `>= 3` gate that exists to *discard* initials, and impossible for any author name to contain. This is #1608 re-implemented in a second package.

2. **No Unicode normalization anywhere in `internal/importer`.** Every provider that fills the catalogue emits NFC; macOS hands filenames back decomposed. The two byte forms of `Björn Andersen` never met.

3. **ASCII `\b` on the catalogue side** — #1642's importer half.

Defect 1 is directional, and the failing direction is the common one: the **parsed** side (an author folder, or an ID3 `Artist` tag, where audiobook releases overwhelmingly write `J.R.R. Tolkien`) has the run-together form while the catalogue has the spaced form OpenLibrary and DNB return.

None of these degrade gracefully. `matchingAuthors` verifies index candidates with this predicate and returns an **empty non-nil** set, which the caller reads as "restrict candidates to these" — so the candidate list is empty, the title tier never runs, and the book stays unmatched across every rescan.

Both sides now reduce through `textutil.NormalizeAuthorName` and compare as token sets. `authorWordIndex` is rekeyed on that same reduction, so its super-set guarantee holds **by construction** rather than by coincidence; `firstWordRun`/`wordRunsInto` are deleted. `TestAuthorWordIndexIsSuperset` re-proves the invariant, because if it ever breaks the scan silently drops books rather than slowing down.

## The title half

`titleMatch` had the same shape inside one function: `normalizeTitle` did lowercase + articles, while its neighbour `sigTokens` kept only `[a-z0-9]` with a 2-char floor. Two reductions, two alphabets.

```
"Die Höhle"  -> [die hle]     # ö splits the word, leftover "h" falls under the floor
"Der Prozeß" -> [der proze]
"Ender's Game" vs "Enders Game" -> no match
```

Both now go through `textutil.FoldForTitleMatch`.

## Also

`lookupAuthorMatch` (manual + bulk import) gains an exact-via-variants check ahead of its 0.80 Jaro-Winkler gate — raw scoring can never reach it for a transliterated pair, `"Boell, Heinrich"` vs `"Heinrich Böll"` scores 0.441. Strictly additive; the 0.80 check still runs.

## Verification

Wrote verbatim copies of the pre-fix `authorMatch`, `significantAuthorTokens` and `sigTokens` and asserted each new test case genuinely failed against them (including that `"Die Höhle"` really did reduce to `[die hle]`), then deleted the probe. Negative tests pin that the widening does not merge distinct authors (`Emily Brontë` vs `Charlotte Brontë`) or distinct books.

Full suite green, vet and gofmt clean.

Closes #1646


vavallee:
Superseded by #1661 (same branch, rebased onto main). GitHub auto-closed this when the base branch was deleted on merge of #1656.

Issue #1471, opened 2026-07-07
decision package: half the specs are dead code with latent bugs (wire in or delete)


vavallee:
Found during a codebase sweep.

## Problem

About half the `internal/decision` package is dead code, and the dead half has latent bugs. `QualityAllowed`, `QualityCutoff`, `SizeLimitSpec`, `LanguageFilterSpec`, and `CustomFormatScoreSpec` (plus `Release.CustomScore`) are constructed only in tests. Production (`scheduler.go:513`, `api/indexers.go:392`) uses only `BlocklistedSpec`, `DelayProfileSpec`, `AlreadyImportedSpec`.

Within the dead code:
- `matchCondition` (`specs.go:244`) matches `releaseGroup` patterns against the whole title (the parsed `ReleaseGroup` never reaches `decision.Release`).
- `CustomCondition.Required` is ignored.
- The documented `"size"` / `"indexerFlag"` condition types silently never match (`default: return false`), so any format using them is unmatchable.
- Scoring weight is slice position (`i+1`), not a user value.

This is maintenance surface that lies about the product's capabilities.

## Suggested fix

Either wire these specs into the real decision pipeline (and fix the bugs above), or delete them. If custom formats / quality profiles are a planned feature, track that separately and keep the dead code behind a clear "not wired yet" marker.

vavallee:
Revisited this in a sweep against `main` at `7917b7ec`. Most of it is now stale:

- `QualityAllowed` is wired. #1693 added it at `scheduler.go:802`, `api/indexers.go:551` and `importer/scanner.go:370`, and #2307 then narrowed it per media type.
- `QualityCutoff` is still dead, and it turns out the reason is bigger than "no constructor": there is no upgrade re-grab path anywhere for a cutoff to gate, and the Quality profile editor makes the cutoff mandatory while telling users it stops upgrade searching. Split that out into #2373 with both options written up, because it needs a product decision and #2333 is blocked behind it.
- `SizeLimitSpec`, `LanguageFilterSpec`, `CustomFormatScoreSpec` and the `matchCondition` / `CustomCondition.Required` / `"size"` and `"indexerFlag"` bugs listed here are still exactly as described.

Leaving this open for that remainder.

vavallee:
[...]

PR #1696, opened 2026-08-02
fix(indexer): enforce the quality profile allowed-formats list (#1693)

### LINKED ISSUE #1693: Quality profile "allowed formats" list is not enforced during search
### Pre-flight

## Summary

The `QualityAllowed` decision spec (`internal/decision/specs.go`), which is supposed to reject a release whose format isn't in a quality profile's allowed list, is never actually constructed or added to any `DecisionMaker` anywhere in the codebase. Confirmed by grepping the full source tree for `QualityAllowed` — it appears exactly 4 times, all within its own definition in `specs.go`. No other file references it.

## What's enforced instead

`internal/indexer/searcher.go` scores candidate releases using the global `models.QualityRank` constant (a fixed 1–10 scale across all formats) as a ranking *preference*, not the profile's per-format `allowed` boolean as a *filter*:

```go
score := float64(models.QualityRank[quality]) * 100
```

Separately, `QualityCutoff` *is* wired in (via `scheduler.go`'s `checkPendingReleases`), but it only prevents re-grabbing a worse copy of a book you already have — it does nothing to gate an *initial* grab.

## Impact

A quality profile's "Allowed formats" checkboxes read as a hard allow-list in the UI, but nothing currently stops a disallowed format (e.g. PDF, mobi) from being selected and grabbed if it's the best-scoring or only available candidate for a title. Users relying on this as a strict filter — which the UI strongly implies — are not actually protected.

## Suggested fix

Wire `QualityAllowed` into whatever spec list `DecisionMaker` evaluates during search, alongside the already-active `LanguageFilterSpec` / `QualityCutoff` / etc.

### Alternatives considered

_No response_

### Additional context

_No response_

vavallee:
## Summary

`decision.QualityAllowed` was written but never constructed anywhere in production code — grepping the tree finds it only in its own definition and its own unit test. A quality profile's **Allowed formats** checkboxes therefore did nothing at grab time. Closes #1693.

What was actually running is `models.QualityRank` scoring in `internal/indexer/searcher.go`, a fixed 1–10 preference across all formats. That ranks candidates; it does not filter them. A disallowed format was grabbed whenever it was the only candidate, or simply the best-scoring one — and since `azw3` outranks `epub`, an EPUB-only profile would still pick the AZW3.

## Implementation notes

Both search paths resolve the profile through one helper, `db.ResolveAuthorQualityProfile`, so auto-grab and interactive search cannot disagree about which profile governs a book. Disagreeing would mean a release the user sees marked "rejected" in the search dialog gets grabbed anyway by the next scheduled sweep.

| Path | Behaviour | Why |
|---|---|---|
| Scheduler (`searchAndGrabFormat`) | **Enforces** — rejected releases are never grabbed | No human in the loop |
| Interactive search (`SearchBook`) | **Annotates** — all results returned with `approved=false` + reason | The user may deliberately want a disallowed format for one book; hiding it would be worse than the bug |

The spec goes on the same `DecisionMaker` that `checkPendingReleases` re-uses, so a rejected release keeps failing re-evaluation instead of slipping through on a later sweep.

**A release with no parseable format always passes.** `ParseRelease` only sets `Format` when it finds a known token in the title, and plenty of legitimate Usenet releases carry none (`Author - Title (Year)`). Rejecting those would turn any ticked checkbox into a near-total grab blackout — the opposite of what the UI promises. The filter can only speak to formats it can actually see. This is the one semantic change to the spec itself and it is pinned by a test.

**No behaviour change for existing installs that haven't configured a profile:** an author with no `quality_profile_id` gets no spec, and the three seeded profiles ship with `items='[]'`, which already meant allow-all. Both are pinned by tests.

**`QualityCutoff` is still unwired.** Issue #1693 states it is active via `checkPendingReleases`; it is not — it is dead in exactly the same way. Out of scope here, and it belongs with #1471 (decision package: half the specs are dead code).

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — no env vars, config, or exported-symbol behaviour changes beyond the new godoc'd `db.ResolveAuthorQualityProfile`
- [ ] Wiki pages updated if user-facing behaviour changed — n/a (the UI already described this as an allow-list; it now is one)

## Test plan

- [x] Non-vacuity verified: dropping the spec append makes `RejectsDisallowedFormat` and `PicksAllowedOverBetterRankedDisallowed` fail, the latter grabbing the higher-ranked AZW3 exactly as reported
- New `internal/scheduler/scheduler_quality_test.go` — disallowed rejected, allowed grabbed, allowed beats better-ranked-disallowed, untagged passes, no-profile unfiltered, empty-items unfiltered
- New `TestSearchBook_QualityProfileAnnotatesDisallowedFormat` — all three results returned, only the PDF unapproved, rejection reason present


vavallee:
Pushed \`9802ff57\` for the codecov/patch flag.

The behaviour was already covered end to end through the scheduler and the search endpoint, but Go attributes coverage per package — lines in `internal/db` and `internal/decision` get no credit from tests living in `internal/scheduler` or `internal/api`, so both new files reported 0%.

Added direct unit tests, which are worth having regardless: the fail-open on an unparseable release format is the one semantic this PR changes and deserves a unit test rather than only an integration one, and the resolver has five paths that must all return nil (no profile, no repo, nil author, dangling id) because a profile that cannot be read must not become an accidental grab block. Also pinned the rejection string, since interactive search renders it verbatim.

The scheduler fixture now attaches the repo through `WithQualityProfiles` instead of setting the field directly, so the wiring `main.go` uses is exercised.

Issue #1691, opened 2026-08-02
ABS importer matches by title, so retitling an item in ABS + re-import mints a duplicate book record — including 'wanted' twins of owned books


gavinmcfall:
### Environment
- Bindery `v1.28.0` (commit `14788f4`), single instance, embedded SQLite.
- Source of truth = Audiobookshelf (2.x) via the ABS importer, 15 book libraries.
- Author monitoring **off** — Bindery used as an owned-tracker over an existing library.

### What happened
I cleaned up titles on ~100 items in ABS (stripped `(Unabridged)`, series suffixes, fixed casing — files and paths unchanged, only the ABS metadata title), then re-ran the ABS import so ownership stays current.

**Result:** every retitled item now has **two** book records — the old row under the old title and a freshly minted row under the new title, both `status: imported`, both pointing at the same on-disk files:

```
'The Botanist:'      (id 33,   imported)  /  'The Botanist'       (id 3726, imported)
'Chapter House Dune' (id 405,  imported)  /  'Chapterhouse: Dune' (id 1809, imported)
'The priests of Psi' (id 1804, imported)  /  'The Priests of Psi' (id 3860, imported)
```

An audit across my instance found **77 duplicate title+author groups** (out of ~5,200 book rows). They fall into three tiers, and the third is the dangerous one:

- 25 groups: both rows `imported` (the retitle pattern above — catalog bloat, stats inflated).
- 41 groups: both rows `wanted` (duplicate wanted rows — would double-grab when acquired).
- **11 groups: one row `imported`, its twin `wanted`** — e.g. `Archangel's Legion` (imported) vs `Archangels Legion` (wanted), `Brayan's Gold` (imported) vs `Brayans Gold` (wanted). These are standing orders to re-download books I already own the moment monitoring/auto-search is enabled.

The comma-inverted forms also pair up (`A Trace of Death` / `Trace of Death, A`), so it isn't only punctuation drift.

### Root cause (as I read it)
The ABS importer resolves "which book record is this item?" by **title (+author) matching**, not by any stable identity. ABS items carry a stable `libraryItemId`, and the files on disk have stable paths — but a title change breaks the match, so the importer mints a new record and the old one is never reconciled or removed. Author-catalogue refreshes (OpenLibrary) then independently create near-miss `wanted` rows (`Brayans Gold`) that never merge with the owned spelling.

[...]

PR #1697, opened 2026-08-02
fix(abs): prune stale book_files rows and add a row-scoped deregister (#1692)

### LINKED ISSUE #1692: ABS re-import appends a new book_files row when a book's path changes but never removes the old one — no safe way to prune the stale row
### Environment
- Bindery `v1.28.0`, single instance, embedded SQLite, distroless image.
- Source of truth = Audiobookshelf (ABS 2.35.1) via the ABS importer, 14
  book libraries, one genre folder each. `BINDERY_LIBRARY_DIR` /
  `BINDERY_AUDIOBOOK_DIR` = the shared colocated library root.
- Author monitoring is **off** (Bindery used purely as an owned-tracker).

### What happened
I re-filed ~130 owned books into a new top-level genre folder (a new ABS
library) — e.g. `…/Books/Fantasy/Sarah J. Maas/…` → `…/Books/Romantasy/Sarah J.
Maas/…`. In ABS this is a clean move (old item goes missing, new item created in
the new library). I then re-ran the ABS import in Bindery to re-point ownership.

**Result:** each moved book is now correctly owned via a **new** `book_files`
row pointing at the new path — but it **also retains the old row** pointing at
the now-nonexistent old path. Example (`GET /api/v1/book/263`):

```
bookFiles:
  1039 ebook      /media/Library/Books/Books/Fantasy/Sarah J. Maas/…/….epub    ← STALE (file moved away)
  1040 audiobook  /media/Library/Books/Books/Fantasy/Sarah J. Maas/…            ← STALE
  4288 ebook      /media/Library/Books/Books/Romantasy/Sarah J. Maas/…/….epub   ← valid
  4289 audiobook  /media/Library/Books/Books/Romantasy/Sarah J. Maas/…          ← valid
```

Owned detection still works (`status: imported`), but the book's top-level
`ebookFilePath` surfaces the **stale** path, and ~130 books now carry dead
rows. (Separately, ~36 books that had a copy under a now-deleted sandbox tree
show the same pattern.)

### Root cause (as I read the source)
- `internal/abs/import_files.go reconcileFormatPath` → `db.BookRepo.SetFormatFilePath`,
  which is `return r.AddBookFile(...)` — i.e. the new path is **added**, the old
  row is not reconciled/removed.
- `internal/importer/scanner.go isReconcileCandidate` returns `false` for an
  `imported` book as soon as **any** tracked path resolves, so a filesystem
  `ScanLibrary` never revisits a book that still has one good file — the stale
  row is never pruned. (`reconciled=0` in the scan result for these books.)

### What I tried (and why each doesn't fit)
- `POST /api/v1/library/scan` — read-only reconciliation report
  (`files_found` / `already_tracked` / `unmatched`); does not prune stale rows
  for still-owned books (see `isReconcileCandidate`).
- `DELETE /api/v1/book/{id}/file?format=ebook` — **format-scoped**: it runs the
  on-disk delete for *every* ebook path of the book (`removeBookPathScoped` +
  stem-sibling sweep + empty-dir removal), which would destroy the **valid**
  new file. There is no `?path=` / per-`book_files`-row variant.
- `PUT /api/v1/book/{id}` — ignores `ebookFilePath` / `audiobookFilePath`
  (derived fields).
- Direct SQLite edit — image is distroless and the config volume is RWO, so
  there's no in-place maintenance path without downtime.

### Ask
What's the recommended way to prune a stale `book_files` row from a book that
still has a valid file? Specifically, would you consider any of:
1. Having the ABS importer **reconcile** an existing format's `book_files` row
   to the new path (replace/dedupe) rather than append, when a matched item's
   path changes; and/or pruning rows whose path no longer exists on disk.
2. Making `ScanLibrary` prune `book_files` rows that fail `os.Stat`, even when
   the book retains another valid file.
3. A row-scoped deregister endpoint — e.g. `DELETE /api/v1/book/{id}/file?path=…`
   (or by `book_files.id`) that removes the DB row **without** the on-disk
   delete, for exactly this "file relocated out from under me" case.

Happy to provide a full `bookFiles` dump or test against a build.

vavallee:
## Summary

`BookRepo.SetFormatFilePath` is literally `return r.AddBookFile(...)` — it appends. So re-filing books into a different folder or ABS library and re-importing left each book owning both the new `book_files` row and the dead old one, and the book's derived `ebookFilePath` could surface a path that no longer exists. Closes #1692.

Nothing cleaned it up afterwards: `isReconcileCandidate` returns false for an `imported` book as soon as **any** tracked path resolves, so a library scan never revisits a book that still has one good file. The reporter ended up with ~166 affected books and no safe way out — the format-scoped delete runs the on-disk delete for *every* path of that format plus the stem-sibling sweep, which would have destroyed the valid new file.

## Implementation notes

Two halves, matching the two problems (recurrence and existing damage):

**1. `pruneVanishedFormatPaths` in the ABS importer** — after a new path for a format is stat-verified and stored, drop this book's *other* rows of that same format whose files are confirmed gone.

Deliberately narrow, because "the file is missing" is not always "the file is gone" — an unmounted NFS share makes every path vanish at once:

| Guard | Effect |
|---|---|
| Scoped to the book just reconciled | A book nothing replaced keeps all its rows, however stale |
| Scoped to the format just written | A missing audiobook survives an ebook reconciliation; the formats move independently |
| Only after a new path was stored | No prune without a confirmed replacement |
| The freshly written path is never a candidate | Cannot delete the row just committed |
| Only a definite `ENOENT` prunes | Permission / I/O / stalled-mount errors leave the row alone |

Best-effort: every error is logged and swallowed, so it cannot fail an import that already succeeded.

**2. `DELETE /api/v1/book/{id}/file?path=…`** — deregisters one tracked path, DB-only, no filesystem access at all. This is the issue's option 3, and it is what unblocks instances that already accumulated dead rows.

The path must already be tracked against the book. Combined with the existing tier-1 ownership check that runs before this branch, a caller can only ever detach a row from a book they already own. Comparison is on `filepath.Clean` both sides so a path copied out of a JSON dump with a trailing slash still matches, but the row is deleted by its exact stored spelling.

The issue's **option 2** — have `ScanLibrary` prune any row failing `os.Stat` — is not implemented. It is the same operation without any of the guards above: an unmounted share during a scheduled scan would clear a library's file rows wholesale. If it's wanted later it needs a mount-liveness check first.

**#1691** (retitling mints a duplicate *book* record) is the sibling report and is not addressed here. It needs an identity-keying decision (ABS `libraryItemId` vs. path), not a patch.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — `docs/API.md` documents the new `?path=` mode

## Test plan

- New `internal/abs/import_files_prune_test.go` — prunes the dead row; keeps BOTH rows when both files exist; leaves other formats alone; never prunes the keep-path
- New `TestBookDeleteFile_DeregisterByPath` / `_DeregisterUntrackedPath` / `_DeregisterLastFileFlipsToWanted` — the valid file survives, an untracked path 404s without touching the book's own rows, and dropping the last row refreshes status to `wanted` with the file still on disk


vavallee:
Pushed \`ca8ef4d9\` for the codecov/patch flag.

Added the tests that were worth adding on their own terms — the non-ENOENT guard was asserted in the PR body but not in a test, which is backwards for the one behaviour deciding whether an unmounted share clears a library. A path whose parent is a regular file stats with ENOTDIR (not `os.IsNotExist`), so it reproduces without root or a real mount. Also pinned the `deregistered` history flag and trailing-slash path matching, both claimed in comments but unchecked.

The lines still showing uncovered in both functions are repo-error branches — `ListFiles`, `RemoveBookFile`, and `GetByID` failing. Reaching those needs a mocked `*db.BookRepo`, which the project testing guidance rules out ("mocked DB tests pass when migrations are broken"). Flagging rather than contorting the tests to chase the number.

PR #1711, opened 2026-08-03
fix(series): apply genre override before adding books

### LINKED ISSUE #1709: "Set genre" is only offered after a book has been added
From discussion #1682 (Aiz0): "(also noticed that \"set genre\" is only added after a book is added, shouldn't that always be available?)"

### Problem
In the series view, the genre control only appears once a book has been added, so you cannot set it as part of adding. For a series being filled in bulk that means going back over each book afterwards.

### Proposed solution
Offer the control before the add as well, so the choice applies to the books being created.

### Additional context
Discussion #1682, which reports several problems with adding light novel series. The volume-collapse bug from the same discussion is fixed in #1704.

w3lld1:
## Summary

I made the series genre override available before a series contains any books. The override is now persisted on the series and applied and locked on books subsequently created through the Hardcover fill flow.

Closes #1709.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` not applicable (no environment, configuration, or upgrade-path changes)
- [x] Updated the Hardcover series documentation

## Test plan

- [x] `go test -p 2 -count=1 ./internal/api ./internal/db`
- [x] `go vet ./internal/api ./internal/db ./internal/models`
- [x] `cd web && npm test` (446 tests)

I did not run the repository-wide `make check`; its full Go race/lint/build matrix is deferred to CI. The changed backend modules, complete frontend suite, frontend production build, typecheck, and lint completed locally.

### DIFF
--- a/changelog.d/1709-series-genre-before-add.md
+++ b/changelog.d/1709-series-genre-before-add.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Series genre override** (#1709) — allow setting genres before adding books and apply the choice to newly created series books.
--- a/docs/Hardcover-Series-Wiki.md
+++ b/docs/Hardcover-Series-Wiki.md
@@ -36,7 +36,8 @@ If either requirement is missing, Bindery hides the enhanced controls and the en
 4. Use **Search** to find a matching Hardcover series.
 5. Pick the correct Hardcover result, or remove a bad existing link.
 6. Open the linked series to load the Hardcover catalog diff.
-7. Use **add** on a single missing row, or **add all** / **Fill gaps** to queue every missing catalog entry.
+7. Optionally use **Set genre** before adding books. The override is applied to current and newly added books in the series and locked against metadata refresh.
+8. Use **add** on a single missing row, or **add all** / **Fill gaps** to queue every missing catalog entry.
 
 Added Hardcover books are created as wanted and monitored rows. Bindery then queues indexer searches the same way it does for other wanted books.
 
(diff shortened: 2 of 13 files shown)


vavallee:
Validated the review finding against this branch rather than by reading, and it reproduces:

```
GET /series/{id} -> 200 {"id":1,...,"monitored":false,"createdAt":"..."}
RESULT 1: override is NOT visible to the client
RESULT 3: after applying an empty list -> GenreOverrideSet=true GenreOverride=[]string{}
RESULT 3: override CANNOT be cleared
```

Both model fields are `json:"-"` and only `GetByIDForUser` selects the column, so the series page never learns an override exists. The prompt opens blank, nothing marks the series, and `SeriesPage.tsx:81` returns early on empty input, so once an override is set there is no way to remove it from the UI.

Rather than ask you to rework this, I put the follow-up in #1722, stacked on your branch: exposes the field, selects it in the remaining list queries, adds `DELETE /series/{id}/genres`, and makes the prompt an edit rather than a blank slate. Merge this one first and #1722 collapses to a single commit on top.

The core of this PR is right, and I checked the part that worried me most: `ensureHardcoverCatalogBook` returns at series.go:1406 for a matched existing book, so the override only ever touches newly created rows and cannot clobber genres on books already owned. Migration `067` is also clear of collisions.

Two small things not worth blocking on. The `Down` section says SQLite cannot `DROP COLUMN`, but `066_indexer_freeleech_only.sql` does exactly that. It is moot either way since `parseMigration` (internal/db/db.go:621) truncates everything after `-- +migrate Down` and never runs it, so bindery is forward-only. And this quietly changes what the existing Set genre button means, from a one-shot bulk edit into a standing policy for future books; the wiki update covers it but nothing in the UI signals it, which #1722 addresses with the button state.

Nice work, and thanks for the thorough test plan.

PR #1770, opened 2026-08-04
fix(web): note restart requirement on Google Books key and primary provider

vavallee:
## Interim for the two boot-only settings

`googlebooks.apiKey` (builds the Google Books enricher) and `metadata.primary_provider` (picks the aggregator's primary) are both read once at boot, so saving a new value in Settings has no effect until a restart — with no indication why (audit finding).

The proper fix is a live-reconfigurable aggregator, which is a larger, high-blast-radius change. This is the safe interim: surface the restart requirement so the fields don't read as if the change applied immediately.

## Change
Add the existing `"Takes effect after the next Bindery restart."` hint (generalised to `settings.general.restartRequired`, same copy as the wanted-search-interval note) under both fields. Pure UI copy — no behavior change.

- `web/src/pages/settings/ApiKeysTab.tsx` — hint under the Google Books API key.
- `web/src/pages/settings/MetadataTab.tsx` — hint under the primary provider selector.
- `web/src/i18n/locales/en.json` — new `restartRequired` key.

`tsc` clean; the SettingsPage suite (55 tests) passes.

### DIFF
--- a/changelog.d/restart-hints.md
+++ b/changelog.d/restart-hints.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Google Books API key and primary metadata provider now show a "restart required" hint** — both are read once at boot, so a change only takes effect after a Bindery restart. The settings fields now say so (matching the existing wanted-search-interval note), instead of silently appearing to save with no effect.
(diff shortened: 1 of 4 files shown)


vavallee:
Interim only. The real fix (live-reconfigurable aggregator) is tracked in https://github.com/vavallee/bindery/issues/1771 — keeping this PR open until that lands or is deferred.

Issue #1732, opened 2026-08-04
Hardcover list pinned to Ebook creates books as 'both' — edition hydration overwrites the pinned media type


vavallee:
### Problem

A Hardcover import list pinned to Media Type = Ebook creates most of its books as `both`. Reported on Discord #bug-reports by ballflartcallmop (2026-07-23) with the asymmetry that makes it confusing: an Audiobook-pinned list works correctly, an Ebook-pinned list doesn't. Verified unchanged on `main` @ 96b70c73.

(The Discord thread's reply dealt with bulk-removing the mis-created queue items — that shipped as #1622 — but the media-type defect itself was never filed. This is it.)

### Root cause

`syncList` (`internal/hardcoverlistsyncer/syncer.go:245`) sets `book.MediaType = il.MediaType` before creating the book, then immediately calls `hydrateHardcoverEditions` (:259), which looks for an audio-shaped edition (`isLikelyAudioEdition`: "audio"/"audible"/"mp3"/"cd"/"cassette" in the format). If one exists, `deriveAudiobookMetadataFromEdition` (`internal/bookhydrate/hardcover.go:201`) runs:

```go
switch book.MediaType {
case "":
    book.MediaType = models.MediaTypeAudiobook
case models.MediaTypeEbook:
    book.MediaType = models.MediaTypeBoth
}
```

The `ebook → both` arm treats "deliberately pinned to ebook" the same as "no format decided yet" and overwrites the list's choice the moment any audio edition exists for the work on Hardcover — which is true for most popular titles. The row is persisted wrong immediately after creation.

Audiobook-pinned lists are immune only because a book created as `audiobook` already satisfies `bookAcceptsAudiobookASIN`, so the switch never runs. That accident is the whole reason one list type works and the other doesn't.

The reporter also correctly notes this undermines `syncList`'s own comment promising "two single-format lists never combine into both" — that guarantee only holds against re-sync collisions, not this same-request hydration step.

### Fix

`deriveAudiobookMetadataFromEdition` needs to distinguish a pinned MediaType from a zero value — e.g. skip the `ebook → both` promotion when the book was created via an import list with a non-empty MediaType override. The `"" → audiobook` arm is fine and should stay.

PR #1784, opened 2026-08-05
fix(hardcover): stop per-book edition fan-out during list sync (#1694)

### LINKED ISSUE #1694: Hardcover list sync unconditionally fetches every tracked edition per book, causing cascading timeouts on high-edition-count titles
### Pre-flight

## Summary

`hardcoverlistsyncer.syncList` calls `hydrateHardcoverEditions` for every book it imports, which calls `client.GetEditions` — a fully paginated GraphQL query that fetches **every edition Hardcover tracks for that book**, with no cap:

```go
const slugGQL = `query GetEditions($slug: String!, $limit: Int!, $offset: Int!) {
    editions(where: {book: {slug: {_eq: $slug}}}, limit: $limit, offset: $offset, ...)
```

For a book with many published editions (translations, formats, regional printings), this means dozens of sequential local DB upserts per book, all within one sync run's shared time budget.

## Observed impact

On a real library sync (~560 books), a single title — "The Two Towers" — triggered **50+ edition upserts, all failing with `context deadline exceeded` within 1.2ms of each other** (i.e. the parent context was already expired, so every subsequent attempt in the batch failed immediately without even trying). This cascades onto *unrelated* books queued in the same batch — we observed completely unrelated titles (light novels, other unrelated authors) failing in the same burst purely because they shared the exhausted context.

Removing the offending title from the source list measurably improved throughput on the next sync, and the same pattern immediately recurred on the next high-edition-count book ("Fourth Wing," 20+ editions). This is a general pattern affecting any popular/classic title, not a one-off.

## The fix doesn't require removing the feature

The book-level query already used by `GetListBooks`/`getShelfBooks` returns `default_ebook_edition_id` and `default_audio_edition_id` in the same response:

```go
book {
    ...
    default_audio_edition_id
    default_ebook_edition_id
```

List sync's actual job is "mark this book wanted" — it doesn't need the full edition catalog to do that. Using the default edition ID(s) already present in the response, instead of unconditionally calling `GetEditions` for the complete list, would eliminate the unbounded fan-out entirely for this code path. Full multi-edition hydration may still make sense elsewhere (e.g. manual per-book edition browsing in the UI) — this report is specifically about it running unconditionally during list sync, where it isn't needed.

## Suggested fix

In `hydrateHardcoverEditions` (or upstream in `syncList`), use `book.DefaultEbookEditionID`/`DefaultAudioEditionID` when available instead of always calling `client.GetEditions`, or at minimum cap/paginate-with-early-exit so one edition-heavy title can't exhaust the shared sync context and cascade failures onto unrelated books in the same batch.

### Alternatives considered

_No response_

### Additional context

_No response_

koseozgur:
## Summary

`syncList` called `GetEditions` for every newly imported book — a fully paginated GraphQL query (up to 1,000 editions per book), with each edition then upserted individually. All of that shares the single 60-second context the API handler bounds the whole sync with (`internal/api/import_lists.go:296`), so a **first** sync of a large list only ever gets through a handful of books before the deadline trips.

Everything after that fails instantly against the dead context, and since per-book failures are logged-and-`continue`d, `syncList` returns `nil` and the handler still replies `{"status":"ok"}`. `UpdateLastSyncAt` fails against the same dead context and is only logged — so `last_sync_at` doesn't move either. There is no signal anywhere that the sync did almost nothing.

The two things hydration actually contributed on this path — an audiobook **ASIN** and the **audiobook/both media type** — are both obtainable from the list query itself. It already selects `default_audio_edition_id` / `default_ebook_edition_id`; this adds a server-filtered slice of ASIN-bearing audio editions next to them, and `toBook` lifts the ASIN off that. An inlined audio edition also implies the work has an audiobook, which preserves the media-type promotion the fan-out performed.

Closes #1694.

## Measured

Against a real 1,660-book "Want to Read" shelf, same machine, same account, only the binary differs:

| | before | after |
|---|---|---|
| books imported | 216 | **1,485** |
| wall clock | 60s (deadline) | **6s** |
| `context deadline exceeded` | 1,444 | **0** |
| books with an ASIN | 18 | **330** |
| edition rows written | 2,910 | 0 |

The remaining 175 of 1,660 are fully accounted for: 174 collapsed by the existing `#1223` dedup (`FindByAuthorAndDedupKey`) and 1 skipped for having no author metadata. Verified at debug log level — nothing unexplained.

**More ASINs than before**, because filtering server-side on `reading_format = "Listened"` finds audio editions the old `edition_format`-string heuristic missed. The added selection is effectively free — a 100-book page went **0.854s → 0.876s**.

## Revised after review (see comments)

The review round found the original description understated what hydration provided: it also filled `Language` (whose absence can silently drop books under `UnknownLanguageBehavior == fail`) and triggered Audnex enrichment. Both are restored in the follow-up commit — language via the default-edition relations inline on the list query (the `books` type has no `language` field), Audnex via `WithAudiobookEnricher` on the syncer. Fresh 3-way measurements against v1.30.0 are in the review reply below.

## Behaviour changes worth calling out

- **Editions are no longer persisted during list sync.** ~~Nothing reads that table for list-synced books today~~ **Correction (review):** `resolveCalibreEdition` in the importer does read it for the Calibre handoff — for list-synced books that path now degrades gracefully (nil-guarded, confirmed passing) and a proper fix is proposed as a follow-up. The four on-demand hydration call sites (book rebind, author add, series link, recommendations) are untouched.
- **`#1732`'s asymmetry is preserved.** A list pinned to a non-audio format now drops an ASIN the list response carried, matching the pre-existing rule that an ebook-pinned book never takes the audio edition's ASIN.
- `ListSyncer.WithEditionHydration` and its fields are removed rather than left dead.

## Test plan

- [x] `TestSync_DoesNotFanOutToEditions` — new regression guard; the fake client now counts `GetEditions` calls and the test asserts **0** during list sync.
- [x] `TestSync_AudiobookPinnedListKeepsASIN` / `TestSync_UnpinnedListKeepsDerivedMediaTypeAndASIN` — rewritten to assert the ASIN and media type from the list response survive, and that no fan-out occurs.
- [x] `TestSync_EbookPinnedListSurvivesAudioEditionHydration` — unchanged, still passes (`#1732` guard).
- [x] `TestToBook_InlineAudioEditionSuppliesASIN` — new, covers ASIN normalisation, blank-entry skipping, media-type implication, and the no-inline-editions case.
- [x] `gofmt`, `go vet`, `golangci-lint run` (0 issues, full repo), `govulncheck` (0 vulnerabilities)
- [x] `go test ./cmd/... ./internal/...` — full suite passes
- [x] End-to-end against the live Hardcover API with the numbers above


vavallee (changes requested):
Thanks for this — the diagnosis is right and the perf argument is convincing. Swapping a per-book paginated `GetEditions` fan-out for a server-filtered inline selection is the correct shape, and I can see why it resolves so many more ASINs. I want it in. Three things first, one of which I think you will want to know about regardless.

## 1. The mechanism has no test coverage

I deleted **both** `editions(...)` selections from `lists.go` — the entire fix — and re-ran the suite:

```
ok   internal/metadata/hardcover  0.034s
ok   internal/hardcoverlistsyncer 1.787s
```

Green. `TestToBook_InlineAudioEditionSuppliesASIN` only constructs `hcBook{AudioEditions: ...}` by hand, and the syncer tests inject `ASIN: "B123PINNED"` straight into the fake client fixture, so the whole seam from **query text → JSON decode → `toBook`** is unguarded.

That matters more than usual here because this filter fails *silently*: a wrong predicate returns an empty array, not an error. No log line, no failure, just zero ASINs forever.

The fix is cheap — `TestGetListBooks_PositiveID` (`client_test.go:2067`) already builds a full JSON envelope over httptest. Adding `"editions": [{"asin": "B0..."}]` to it and asserting `books[0].ASIN` pins both the decode path and the field name.

## 2. `"Listened"` contradicts the repo's own fixtures

The filter uses `reading_format: {format: {_eq: "Listened"}}`, but existing fixtures in this repo use a different vocabulary for that field — `client_test.go:1129` has `"reading_format": {"format": "Audiobook"}`, and `:1123`/`:1039` use `"Ebook"`/`"Physical"`.

Those are hand-written mocks rather than captured responses, so they are not authoritative, and your 330 live ASINs are strong evidence `"Listened"` is correct. But the contradiction is unresolved inside the codebase and nothing in CI can catch it either way. Could you paste a captured response fragment showing the enum value? That would settle it and make a good fixture.

## 3. Two claims in the PR body are not accurate

Flagging these because they change whether the trade is worth making — not to nitpick.

**"Nothing reads that table for list-synced books today."** The endpoint half is true, but `internal/importer/scanner_handoff.go:589` calls `s.editions.ListByBook(ctx, book.ID)` in `resolveCalibreEdition`, feeding ISBN13/ISBN10/ASIN, publisher, publish date, language and cover into the Calibre handoff. A list-synced book later downloaded into Calibre loses that. It degrades gracefully — every consumer is nil-guarded, and I confirmed `./internal/importer/...` passes — but it is real data loss.

**The removed hydration did more than ASIN + media type.** Reading `internal/bookhydrate/hardcover.go:204-237`, `deriveAudiobookMetadataFromEdition` also filled `book.Language` and `book.ImageURL`, and `hardcover.go:143` ran `EnrichAudiobook` (Audnex: narrator, refined duration, summary) whenever an ASIN was promoted.

- `ImageURL` is fine — the list query already selects `image { url }`.
- **`Language` is a real loss.** Neither list query selects `language`, so edition hydration was its only source for list-synced books. Empty language is "unknown" to `models.IsLanguageAllowed`, which **drops the book** when the profile has `UnknownLanguageBehavior == fail`.
- **Audnex enrichment stops running on list sync entirely.** `WithEditionHydration` at `main.go:476` was the only scheduler-path wiring; the sole remaining trigger is the manual `POST /book/{id}/enrich-audiobook`. Since this PR resolves ~18× more ASINs, that is arguably the more valuable path going dark — and it is a straightforward add-back: keep the enricher on the syncer and call `EnrichAudiobook` when `book.ASIN != ""`, with no edition fetch needed.

## Smaller things

- The nested selection uses `limit: 2` with no `order_by`, and `toBook` takes the first non-blank. The old path scored editions via `preferredEditionASIN` to pick the best. For a title with several Audible editions (regional, abridged) the ASIN you get is now arbitrary and can differ between syncs. An explicit `order_by` costs nothing.
- `hcASINEdition.EditionFormat` (`client.go:653`) is selected in both queries and never read.

## What I checked and found clean

No `go.mod`/`go.sum` change, no new imports, no new network hosts (same configured Hardcover endpoint), no auth/ownership/tenancy changes, no CI or build-config edits. `go build`, `go vet`, and the hardcover / listsyncer / bookhydrate / importer suites all pass on the branch. The `#1732` ASIN asymmetry is genuinely preserved — `syncList` still clears the ASIN when the list pins a non-audio format.

Holding this out of the current release so it is not rushed. Items 1 and 2 are what I need to merge; item 3 is mostly a question of whether you want to fold the Audnex re-wiring in here or leave it as a follow-up. Happy either way.

koseozgur:
Thanks for the review — every finding held up, and chasing them surfaced one more real problem beyond what you flagged. Point by point, rebased onto v1.30.0:

## 1. Seam coverage — done, and your experiment now fails

`TestGetListBooks_PositiveID` decodes the inline `editions` + language from the raw JSON envelope over httptest and asserts the query text for both selections; `TestGetListBooks_BuiltinShelfInlinesASINAndLanguage` does the same for the shelf path's separate query string. I re-ran exactly your experiment — deleted both selections from `lists.go` — and the suite now fails 4 assertions instead of staying green.

## 2. `"Listened"` — captured evidence

The full `reading_formats` enum from the live API:

```json
{"data": {"reading_formats": [
  {"id": 1, "format": "Read"},
  {"id": 2, "format": "Listened"},
  {"id": 3, "format": "Both"},
  {"id": 4, "format": "Ebook"}
]}}
```

And a captured fragment filtered to `edition_format: "Audible Audio"` — every row returns `"reading_format": {"id": 2, "format": "Listened"}`:

```json
{"id": 30581272, "asin": "0007496788", "edition_format": "Audible Audio",
 "reading_format": {"id": 2, "format": "Listened"}}
```

So the existing fixtures' `"Audiobook"` / `"Physical"` values don't occur in the API at all — happy to file a follow-up correcting them.

## 3. The metadata losses — fixed, with one discovery you'll want to know about

**Language:** first attempt selected `language { language }` at book level, unit tests all green — and the live sync failed instantly with `GraphQL: field 'language' not found in type: 'books' (validation-failed)`. The `books` type has no language field; the three existing `language` selections in `client.go` are all on *edition* queries. So the fix reads it off the relations instead: `default_ebook_edition { language { language } }`, `default_audio_edition { ... }`, plus `language` on the inline audio editions as a last fallback, ISO-normalized in `toBook`. Measured against a real 100-book shelf page: **92/100** books resolve a language from at least one source (85 ebook-default, 45 audio-default, 33 inline). This also means the pre-#1694 hydration path was the only thing that could ever set language for list books precisely *because* the book type can't carry it — worth knowing for any future list-query work.

**Audnex:** `WithAudiobookEnricher` on the syncer restores enrichment exactly as you suggested — fires when a created book's ASIN arrived inline, persists after, failures log-and-continue, placed after the #1732 pinning so an ebook-pinned book never enriches. Tests cover the call, the persist, failure tolerance, and that the fake client still sees zero `GetEditions` calls.

**Calibre handoff:** you're right and the PR body was wrong — `resolveCalibreEdition` does read the editions table. I'd rather not half-fix it here: persisting the inline editions would store ASIN-only stubs (no ISBN/publisher — the slim selection doesn't carry them), which pollutes the table to partially feed one consumer. Proposal: leave it degrading gracefully as you confirmed it does, and handle it in a follow-up — either hydrate-on-demand when the Calibre handoff actually runs (it has the book ID and a client), or accept the gap for list-synced books. Your call on which.

## 4. Fresh measurements — and a deadline question the numbers force

Same 1,660-book real shelf, all three builds current:

| | v1.30.0 baseline | branch, no enrichment | branch + Audnex re-wire |
|---|---|---|---|
| manual sync: books imported | 111 | 1,485 | 519 |
| wall clock | 60s (deadline) | 6s | 60s (deadline) |
| `context deadline exceeded` | 1,540 | 0 | 1,110 |
| books with ASIN | 41 | 330 | 184 |
| books with language | 43 | 0 → *(the bug you caught)* | 309 |
| books with narrator | 18 | 0 | 38 |

Two things fall out:

1. Even with the fan-out gone, **any per-book network call under the request deadline re-collapses at scale** — the Audnex re-wire alone drags the manual path from 1,485 books/6s back to 519 books/60s. The scheduled path is unaffected (`Sync(s.ctx())` has no deadline, so the nightly job completes the whole list, enrichment included). It's specifically the manual "Sync now" — the path a new user hits on first import, which #1694 is about — that can't absorb per-book work.

2. Repeated manual syncs still converge (each pass imports the next chunk), and 519/pass beats the baseline's 111/pass — so this is strictly better than both the current release and the pre-review branch on every metric except raw first-pass count. But it's worth deciding deliberately rather than by accident.

Options as I see them, no strong attachment: (a) ship as-is and treat "move list sync off the request lifetime" (202 + background job, or a spawned task with its own context) as the structural follow-up; (b) skip enrichment when the context carries a deadline — manual syncs stay 6s/1,485, but books they import never get narrator data since sync-time is the only enrichment trigger for list books; (c) raise or drop the deadline for this endpoint specifically. I lean (a) — it keeps this PR reviewable and puts the real fix where the real problem is — but you know the design intent for that 60s bound better than I do.

Branch is rebased on v1.30.0; `gofmt`/`vet`/`golangci-lint`/full suite green throughout.

vavallee (approved):
All three addressed. The seam experiment fails now the way it should, and the language finding is a genuinely useful thing to have written down: `books` having no language field is why the old hydration path was the only thing that could ever set it for list books. That belongs in the record.

On the two you left open:

**Calibre handoff.** Agreed, don't half fix it. ASIN only stubs with no ISBN or publisher would be worse than the gap, since they'd look like real edition rows to everything that reads the table. Leaving it degrading gracefully and filing hydrate on demand at handoff time as a follow up.

**Deadline.** Option (a). That 60s bound is a generic HTTP server timeout, not a considered statement about how long a sync should take, so tuning what fits inside it is solving the wrong problem. List sync should be off the request lifetime entirely. Filing that separately.

519 per pass against the current 111, converging on repeat syncs, with language coverage going from 43 to 309, is strictly better than what ships today on everything that matters. Going in as is. The measurement table made this an easy call, thanks for doing that work.

Issue #940, opened 2026-06-02
ABS + Calibre import: duplicate book rows for the same book, one with files, one empty


vavallee:
## Source

Discord report from ibsfox on 2026-06-02:

> The Bindery import from ABS and CWA seems to work OK, for many books I get two entries — one which shows both the book and audiobook linked, then a second entry which shows no files linked. It seems like it is recognizing the linked book but still creating a second entry in the database on the second system import. If I delete the \"empty\" record, it just re-creates the next time I run an import. How do I stop it from happening?

ibsfox's stack:
- ABS for audiobooks at \`/data/media/audiobooks\`
- CWA (Calibre-Web-Automated) for ebooks
- Bindery imports from both

## What we know

Both importers have a title+author fallback (Calibre at \`importer.go:628\`, ABS at \`import_upserts.go:610\`-ish), so a same-title same-author book *should* merge. The fact that a second empty row keeps re-appearing on re-import means dedup is failing in at least one direction, repeatably.

Candidate causes:

1. **Author name mismatch** between ABS's normalised name and Calibre's. ABS often gives \"Stephen R. Donaldson\" where Calibre has \"Stephen Donaldson\" (or vice versa). The book ends up under two different authors and \`FindByAuthorAndTitle\` can't reconcile.
2. **Empty row is a metadata-provider wanted stub.** When ABS adds an author, Bindery fetches the upstream catalogue and creates wanted stubs (\`internal/api/authors.go::FetchAuthorBooks\`). Calibre's import then creates a separate row for the same title because of subtle differences (subtitle, edition info, ISBN normalisation).
3. **One importer uses the foreign_id namespace the other doesn't recognise.** \`calibre:book:N\` vs \`abs:book:libID:itemID\` are different ID spaces; only the title+author fallback bridges them, and that's the failure point.

## Reproduce

Need from ibsfox:

1. SQL dump for one affected book pair:
   \`\`\`
   sqlite3 /config/bindery.db \"
     SELECT id, foreign_id, title, sort_title, author_id, media_type, file_path
     FROM books
     WHERE title LIKE '%KNOWN-DUP-TITLE%';
   \"
   \`\`\`
2. Author rows:
   \`\`\`
   sqlite3 /config/bindery.db \"
     SELECT id, name, sort_name, foreign_id
     FROM authors
     WHERE id IN (RESULTS-FROM-STEP-1);
   \"
   \`\`\`

[...]

PR #1786, opened 2026-08-06
fix(dedup): keep distinct series volumes from collapsing onto one key

### LINKED ISSUE #1785: Series volumes sharing a base title collapse onto one dedup key ("Series: Volume" merges every volume)
### Summary

`indexer.CanonicalDedupKey` (via `stripSubtitle` in `internal/indexer/dedup.go`) strips everything after the first `": "` in a title. For a series whose volumes are titled `Series: Volume` — e.g. `A Thousand Li: The First Step`, `A Thousand Li: The First War` — every volume collapses to the single key `a thousand li`. The key is the authoritative "same book" signal for every import and add path, so distinct volumes get merged.

Reported on Discord (Stankonator) with two symptoms from one root cause.

### Symptom 1 — ABS import creates fewer books than dry run

Dry run: 902 items, 902 created. Live: 902 items, 824 created, 73 linked, 5 queued for review (824 + 73 + 5 = 902).

`internal/abs/import_upserts.go` `upsertBook` never calls `books.Create` during a dry run, so `findBookByNormalizedTitle` finds zero prior rows and every item lands in the "created" branch. Live persists volume 1; volume 2 then matches it by dedup key and is **linked**; volume 3+ finds 2+ matches and is routed to **review** as ambiguous. The dry-run "created" count is inflated by design; the live numbers reflect the real collapse.

### Symptom 2 — manual add by OLID returns "book not found after author sync — try again shortly"

`internal/api/authors.go` `AddBook` direct-insert path: `FindByAuthorAndDedupKey` matches the already-collapsed row, `adoptDirectInsertMatch` no-ops (the row is not a `calibre:` stub and needs no field change), so the requested foreign id is never bound to any row. The 15s poll never finds it → permanent 404. The user saw the book "minus the subtitle" because that is the base-title row it collapsed onto.

Same root cause behind the reported Gears of War titles and the Tolstoy/Dostoevsky collections merging.

### Fix

Make the dedup match series-aware: a candidate that carries the **same series but a different sequence number** is a distinct work and must not be treated as a dedup match. This is a pure "prevent bad merge" guard keyed on positive evidence (different sequence), so it never splits a genuine ebook/audiobook pair (same sequence, or no series).

Known limitation: the manual-add fix only fires when the metadata provider supplies a series sequence for the requested work. OpenLibrary series data is inconsistent, so some manual-add cases may still need the provider to report position.

vavallee:
Fixes #1785. Reported on Discord (Stankonator).

## Problem

`indexer.CanonicalDedupKey` strips a `": subtitle"` tail (`stripSubtitle`, `internal/indexer/dedup.go`), so a series titled `Series: Volume` collapses every volume onto one key. That key is the authoritative "same book" signal, so distinct volumes were merged. Two symptoms, one root cause:

1. **ABS import undercount.** In `internal/abs/import_upserts.go`, `upsertBook` never persists during a dry run, so `findBookByNormalizedTitle` finds nothing and every item is "created" (902). Live persists volume 1; volume 2 links onto it; volume 3+ hits ambiguity and is queued for review. A 902-item library created only ~824 books (824 + 73 linked + 5 review = 902). Same collapse behind the reported Gears of War titles and the Tolstoy/Dostoevsky collections.
2. **Manual add 404.** In `internal/api/authors.go`, the `AddBook` direct insert matched the collapsed row via `FindByAuthorAndDedupKey`, `adoptDirectInsertMatch` no-op'd (not a `calibre:` stub, no field change), so the requested foreign id was never bound and the 15s poll returned `book not found after author sync — try again shortly` forever.

## Fix

Make the dedup match series aware: a candidate that shares a series with the incoming work but sits at a **different sequence number** is a distinct work and must not be treated as a dedup match. Positive-evidence guard only, so it never splits a genuine same-volume ebook/audiobook pair (same sequence, or no series), preserving the #940 cross-source behavior.

- `findBookByNormalizedTitle` gains an `incomingSeries` parameter and drops conflicting candidates via `candidateSeriesConflicts` (compares each candidate's primary series title + position against the incoming item's series). The hardcover-catalog caller passes `nil` to keep its behavior unchanged.
- `AddBook` direct insert skips the adopt and creates a distinct row when `directInsertSeriesConflict` sees the matched row is a different volume of one of the requested work's series refs.

## Tests

- `TestImporter_SeriesVolumesWithSharedBaseTitleStayDistinct` — 3 volumes sharing a base title create 3 rows (was 1 created + 2 linked before the fix).
- `TestImporter_SameVolumeDifferentFormatStillLinks` — a subtitle-dropped audiobook of the same volume still links to 1 row.
- `TestAddBook_DirectInsertCreatesDistinctSeriesVolume` — adding volume 2 by id returns 201 with a distinct row (reproduces the exact 404 without the fix).
- `TestAddBook_DirectInsertAdoptsSameSequenceEdition` — same-sequence edition still adopts the existing row.

Both fixes verified non-vacuous by neutering the guard and confirming the pre-fix failure.

## Known limitation

The manual-add fix only fires when the metadata provider reports a series sequence for the requested work. OpenLibrary series data is inconsistent, so some manual-add cases may still need the provider to supply position — noted in #1785.


vavallee:
Holding this out of the current release — not because it is wrong, but because it fixes less than the issue asks for and I would rather that be a decision than a surprise. Both tests are genuinely non-vacuous: I neutered each guard in a scratch tree and got "expected 3 distinct volumes, got 1" and the literal reported `404 book not found after author sync — try again shortly`. The reported symptom is reproduced and fixed.

Two gaps:

## The guard is inert against OpenLibrary-created rows

Both conflict checks depend on the candidate having a `primary_series` link. `handleNewWantedBook` links with `ref.Primary`, and **OpenLibrary never sets it** — `internal/metadata/openlibrary/client.go:982` has no `Primary` field, while Hardcover does (`client.go:1046,1069`).

So every row created by an OL catalogue sync, or by a prior OL add, has no primary series and the guard skips it. Probe: add vol 2 by OL id (row lands with `primarySeries=""`), then add vol 3 → **404 "book not found after author sync"**. That is the original bug, unfixed, and reproducible rather than theoretical. The reported case only works because the collapsed rows came from ABS.

Fix shape: a repo method returning *all* `(series title, position)` rows for a book, used by both guards instead of `GetPrimarySeriesForBook`. Worth noting that method is `LIMIT 1` with no `ORDER BY` while ABS links every membership `primary=1`, so for a multi-series book the row it returns is nondeterministic anyway.

## Re-importing does not repair already-collapsed data

In `upsertBook` the ABS provenance lookup (`import_upserts.go:600-627`) runs *before* the dedup match, and the collapsing run recorded provenance for every merged item pointing at the survivor row. Probe: collapse 3 volumes into 1, re-run the import with this fix and full series metadata → still **1 book**, `booksUpdated=3`.

Anyone who already hit this — including the 902-item reporter on #1785 — has to delete the collapsed rows or their provenance links and re-import. Nothing in the PR or the changelog fragment says so, and the fragment reads as though it just gets better. That will generate follow-up reports.

## Smaller

- `AddBook` judges only the first row of `FindByAuthorAndDedupKey` (no `ORDER BY`); with several collapsed siblings it can declare a conflict against the wrong one and create a duplicate. The ABS path filters the full list — the add path should use `FindAllByAuthorAndDedupKey` too.
- Series titles compare with bare `strings.EqualFold` — no NFC normalisation, no umlaut folding, unlike the dedup key itself. `normalizeSeriesName` already exists in `import_hardcover_series.go`. NFD vs NFC titles silently miss the conflict (fails safe, i.e. collapses).
- `sequencesEqual` / `seriesSequencesEqual` are byte-identical duplicates across two packages.
- `internal/calibre/importer.go:703` and `internal/hardcoverlistsyncer/syncer.go:225` still use the raw collapsing lookup, so a Calibre library of `Series: Volume` titles still collapses. #1785 says the key should be authoritative for every import and add path.
- The PR body's "never splits a genuine ebook/audiobook pair" is slightly too strong — it holds only when both sides agree on the sequence. Two sources disagreeing on position (prequel offsets, `0.5` vs `1`) now split where they previously merged.

None of this is a regression against current `main` — the worst new failure mode is a duplicate row, which beats a permanent 404. If you would rather ship the ABS half now, I am fine with that, but the release notes need the remediation step spelled out and #1785 should stay open for the OpenLibrary path.

PR #1795, opened 2026-08-07
refactor(web): restructure the Author and Book detail pages

vavallee:
Step 3 — the layout work. Structure and consistency only: no gradients, shadows or animation added, and no other page touched. Four commits, each independently revertable.

**Based on `fix/detail-page-tailwind-bugs` (#1791)**, not `main`, so the diff here shows only the layout work. Retarget to main once #1791 merges.

## Commits

| | |
|---|---|
| `3100885` | Extract CoverPlaceholder, MoreMenu, Section/Card (no page uses them yet) |
| `1bf0861` | Restructure BookDetailPage |
| `6c48211` | Restructure AuthorDetailPage |
| `dbf72e1` | Changelog fragment |

## Container width

Both pages now use `max-w-7xl`. They previously disagreed — Author was effectively 7xl, Book was 4xl — so author → book collapsed the content 384px and left-aligned it mid-navigation. Rather than narrowing the pages to fix the reading measure, the descriptions get `max-w-prose` locally, so the Author page's book grid keeps its width.

## BookDetailPage

Cover `w-44` → `w-32`. Series name + position added to the meta row. Description swapped from a bare `<p>` to `MarkdownDescription`, so long blurbs clamp. **Edit** moved out of the File card into the header — it edits metadata, not the file. File card keeps Download / Re-bind / Fix match on the row and puts Exclude / Rename files behind **More**. **Delete file** drops from solid red to `btn.danger`; solid red now appears exactly once per page, on the genuinely irreversible **Delete book + files**.

**On series:** there is no book→series endpoint, and adding one would be an API change, so this reuses the existing `GET /author/{id}/series` and picks out this book's entries. It runs after the book loads, never blocks render, and fails silently. Cost is one extra request per book page; `ListByAuthor` returns a trimmed book projection, so the payload is small. Say the word if you'd rather have it on the book payload instead — that's the cleaner shape, but it's a backend change.

## AuthorDetailPage

Stats become a fixed four-cell strip (Books / In library / Wanted / Audiobooks) — the audiobook count used to vanish at zero, changing the row's shape between authors. Action row goes from eight controls to five, with Rename files / Merge / Link metadata / Delete behind **More**; it no longer wraps, so Delete is no longer orphaned onto its own line. Three chip groups (ten buttons) become three selects on one toolbar line, now exposing `downloading` and `skipped`. Grid card text blocks get a fixed height. The `{book.releaseDate &&}` guard is dropped since `fmtPublishedYear` already returns an em dash. Table rows use `navigate()` instead of `window.location.href`.

**One behaviour change worth flagging:** "Show excluded" folds into the status select as an `excluded` option, as requested. It used to *add* excluded books to the current view; it now *narrows to* only them, consistent with every other option in that select. It also persists to localStorage now, which the checkbox never did. Easy to change if you want the additive semantics back.

## Shared components

`CoverPlaceholder` paints a per-book ground colour hashed from the book id with the title set large and ranged left. The palette is pinned to Tailwind's 700 level by three properties enforced in `contrast.test.ts`: white title ≥ 4.5:1 on every ground (AA for *normal* text, not just the 3:1 large-text allowance), and ≥ 3:1 against both page surfaces so the tile edge doesn't dissolve. That third one is why there's no deep blue/indigo/violet — at the saturation needed for white text they land within 3:1 of zinc-950 — and it's exactly what the old flat `bg-zinc-800` box was failing. A regression case pins it.

`MoreMenu` is keyboard accessible: Escape and outside-click both close and return focus to the trigger; Tab closes *without* stealing focus back, so keyboard users aren't yanked backwards out of the tab order. Arrow keys skip disabled items, Home/End jump to the ends.

`Section`/`Card` extracts the pattern BookDetailPage had repeated inline four times and AuthorDetailPage's Books section didn't follow at all.

Button drift fixed: the local `actionBtnCls`/`dangerBtnCls` and `chipCls` are gone, and the hand-rolled `bg-slate-200 dark:bg-zinc-800 hover:…` buttons go through `btn.secondary`. Added `btn.dangerSolid` to buttons.ts rather than hand-rolling solid red again — "solid red means no undo" is a rule that belongs with the vocabulary.

## Preserved

localStorage filter/sort persistence, bulk multi-select scoping to the filtered view, group-by-series, and i18n all still work and are still covered.

## Checks

`npm run lint` 0 errors (6 pre-existing react-hooks warnings unchanged) · `typecheck` clean · `npm test` **504/504** · `build` clean, with all 12 palette classes confirmed present in the compiled CSS.

Tests added: 12 for MoreMenu, 26 for the cover palette contrast, plus new cases for the filter selects, filter persistence, the fixed stat cells, series rendering (including its failure path), and client-side row navigation.

**Note for future tests:** a coverless grid card renders its title twice — card heading plus the aria-hidden CoverPlaceholder. Screen readers get it once; `getByText` matches both, so query the heading.

## New i18n keys

39 new keys, `en.json` only — other locales untouched. Listed in the thread below for translation.


vavallee:
## New i18n keys for translation

**39 added, 0 changed, 0 removed.** `en.json` only — `de/es/fr/id/ko/nl/tl` are untouched, verified against `origin/main`.

⚠️ **One key needs translator care:** `authorDetail.actions.searchWanted` uses `{{count}}`, which i18next treats as a plural trigger. English gets away with a single form because there is no noun after the number ("Search 3 wanted"). Languages with plural categories will want explicit `_one` / `_few` / `_other` variants rather than a single string.

```json
{
  "common": { "more": "More" },

  "bookDetail": {
    "seriesPosition": "{{series}} #{{position}}",
    "description": { "showMore": "Show more", "showLess": "Show less" },
    "renameFiles": { "button": "Rename files" }
  },

  "authorDetail": {
    "booksHeading": "Books",
    "filteredCount": "{{shown}} of {{total}}",
    "groupBySeries": "Group by series",
    "showFlatList": "Show flat list",

    "stats": {
      "books": "Books",
      "inLibrary": "In library",
      "wanted": "Wanted",
      "audiobooks": "Audiobooks"
    },

    "actions": {
      "searchWanted": "Search {{count}} wanted",
      "searchWantedNone": "No wanted books to search",
      "searching": "Searching…",
      "refresh": "Refresh",
      "refreshing": "Refreshing…",
      "edit": "Edit",
      "editHint": "Edit quality, metadata, and root folder",
      "renameFiles": "Rename files",
      "renameFilesHint": "Move this author’s files to match the current naming template",
      "merge": "Merge…",
      "mergeHint": "Merge another author into this one",
      "delete": "Delete"
    },

    "filters": {
      "type": "Type",
      "status": "Status",
      "published": "Published",
      "allTypes": "All types",
      "allStatuses": "All statuses",
      "allPublished": "Any date",
      "released": "Released",
      "upcoming": "Upcoming",
      "status_wanted": "Wanted",
      "status_downloading": "Downloading",
      "status_downloaded": "Downloaded",
      "status_imported": "Imported",
      "status_skipped": "Skipped",
      "status_excluded": "Excluded"
    }
  }
}
```

Notes for whoever picks these up:

- `stats.*` are column labels in a fixed four-cell strip — short is better, they sit above the numbers.
- `filters.status_*` mirror the existing `bookStatus.*` vocabulary; reuse those translations where they already exist rather than inventing new wording.
- `actions.*Hint` are `title` tooltips, so they can run longer than the labels.
- `filters.allPublished` is "Any date" rather than "All" because it filters a date range, not a category.

Issue #1736, opened 2026-08-04
Hardcover list-sync authors are created with null metadataProfileId, bypassing the default stamping every other path applies


vavallee:
### Problem

Authors created by Hardcover list-sync carry a NULL `metadataProfileId`, unlike every other creation path. Reported on Discord #bug-reports by .v.e.g.a. (2026-08-03): an 821-author bulk migration all NULL, then the recurring list-sync produced 146 more overnight, confirmed via API. Verified unchanged on `main` @ 96b70c73.

### One correction to the report

The report concludes "null means no filter regardless of what the profile actually says". That's not what the code does: every reader falls back to `models.DefaultMetadataProfileID` (= 1, the seeded 'Standard' profile with `allowed_languages='eng'`) — `internal/api/authors.go:2520`, `internal/api/indexers.go:507`, `internal/scheduler/scheduler.go:481`. So a NULL-profile author is language-filtered as profile 1, not unfiltered.

The real defects are quieter:

- **Bypassed stamping.** `ensureAuthor` (`internal/hardcoverlistsyncer/syncer.go`) builds the author by hand — sets `Monitored`, `MonitorMode`, `MetadataProvider`, `SortName` — and calls `CreateForUser` directly, skipping `applyAuthorCreateOptions` (`internal/api/authors.go:508`), which is where every API-created author gets its profile defaulted.
- **Silent divergence from operator intent.** An operator who uses a custom profile everywhere still gets profile-1 semantics on every list-sync author, and the UI shows those authors with no profile selected, so there's nothing to notice or correct short of editing hundreds of rows (which is exactly what the reporter did, twice).
- **Fragile fallback if profile 1 is ever deleted:** the three readers then diverge — `resolveAllowedLanguages` in the API returns a hardcoded `[]string{"eng"}` while the scheduler's returns nil (no filter). Same author, different filter depending on which code path runs.

### Fix

Set `author.MetadataProfileID` to the same default `applyAuthorCreateOptions` uses before `CreateForUser` in `ensureAuthor`, and backfill existing NULL rows in a migration so affected libraries (the reporter has ~967 such authors) converge without hand-editing. Longer term, one shared resolve helper instead of three hand-rolled fallbacks would stop the divergence class.

vavallee:
Closed by #1783, which fixes the Hardcover list-sync path specifically and backfills existing NULLs via migration `071`.

[...]

PR #1800, opened 2026-08-07
docs(changelog): assemble v1.30.0

vavallee:
## Summary

Assembles the three `changelog.d/` fragments into a `## [v1.30.0]` section and clears the directory. This is the pre-tag step — the release job reads `CHANGELOG.md` at the tagged commit and aborts if the section is missing, so this has to land on `main` before any tag is pushed.

**No tag pushed.** That is yours to trigger.

## Why minor, not patch

Per the SemVer table in the changelog header, patch is "bug fixes, doc-only updates, security backports". This release goes past that — two things are newly *reachable*, not merely fixed:

- **Series membership on the book detail page.** `series_books` has been populated since v0.7.0; the page never rendered it.
- **The `downloading` and `skipped` status filters.** Both have been in the filter type since it was written, but the chip row only ever offered All / Wanted / Downloaded / Imported, so there was no way to see books in either state.

Cover placeholders are also new user-visible behaviour rather than a fix.

## Behaviour change flagged in the entry

**"Show excluded" changed from additive to narrowing.** It used to add excluded books to whatever you were viewing; folded into the status select, it now shows only them — consistent with every other option in that list. It also persists between visits now, which the checkbox never did. Called out explicitly in the Changed section, because otherwise it arrives as a bug report.

## What is in the release

| PR | |
|---|---|
| #1793 | js-yaml 4.3.1 (GHSA-5p4m-2wfm-xmqj) — build tooling only, no running instance exposed |
| #1791 | Two silently-broken Tailwind classes, plus two lint rules and a regression test |
| #1795 | Author + Book detail page restructure, in five reviewable commits |

Every non-bot commit since `v1.29.1` is represented. Nothing deliberately skipped.

## Checklist

- [x] Tests added or updated — covered by the PRs above; this PR is docs-only
- [x] Doc-update gate cleared — `CHANGELOG.md` is the doc being updated; no `docs/`, README, godoc or Helm values changes ride along
- [ ] Wiki pages updated if user-facing behaviour changed — **the "Show excluded" change is worth a wiki note if the filters are documented there; I have not touched the wiki**

## Test plan

- [x] `cd web && npm run typecheck` — clean
- [x] `cd web && npm run lint` — 0 errors (6 pre-existing react-hooks warnings)
- [x] `cd web && npm test` — 504/504
- [x] `cd web && npm run build` — clean
- [ ] `go test ./cmd/... ./internal/...` — not run locally; this PR touches no Go, and CI covers it on `main`

## Before you tag

Two paths in this release are reasoned-correct but not verified against real data — worth knowing since they land in public release notes:

1. **Series on the book detail page** is covered by unit tests with mocked responses, but I have not seen it against a real book that is actually in a series. The screenshot you shared was of a standalone book.
2. **The grid-card badge fix** addresses an overflow that only triggers with a long translated status label plus three badges. English does not reproduce it, so I verified the reasoning and the deployed CSS, not the visual outcome.

Neither is a reason to hold the release; both are cheap to eyeball on bindery-dev, which is currently serving this exact tree.


vavallee:
## Updated: release scope widened

Four more PRs merged to `main` since this was opened, so the v1.30.0 entry now covers them too and `changelog.d/` is cleared again.

| Merged | |
|---|---|
| `cc1248ec` | #1787 — `BINDERY_TRUSTED_PROXY` forwarded scheme/host docs |
| `f0ed7311` | #1788 — ABS author-lookup memoization |
| `7a686935` | #1783 — Hardcover list-sync default metadata profile (closes #1736) |
| `520d0c24` | #1790 — optional backup label |

Still a **minor** bump. #1790 adds a feature and an optional request body; the rest are fixes and docs.

### Three entries deliberately worded more narrowly than their fragments

Review turned up that the original fragments overstated in ways that would have generated follow-up reports:

- **#1790** claimed labelled *and manually renamed (same-shape)* backups now restore and delete correctly. Only half true — `List` shows any `.db` file while `Restore`/`Delete` still require the `bindery_` prefix and a restricted charset, so a backup renamed outside that shape still lists with buttons that 400. Written up as the limitation it is.
- **#1790** also silently drops a label that sanitises to nothing (an all-CJK one, say). The tests assert that as intended, but it is invisible from outside the code, so the entry says so.
- **#1783** fixes only the Hardcover list-sync path. Five other author-creation paths still insert no profile, so migration `071` is a one-shot cleanup rather than a settled fix — the entry says so and points at #1803.

#1788's entry also names a trade its PR body did not: a provider that degrades mid-import stays degraded for that author until the run ends, where each book previously got a fresh attempt.

### Not in this release

- **#1786** (series-volume dedup) — fixes the ABS case but is inert against OpenLibrary-created rows, and does not repair already-collapsed data. #1785 stays open. Details on the PR.
- **#1784** (Hardcover edition fan-out) — changes requested. Deleting the entire fix leaves CI green, and `"Listened"` contradicts the repo's own `"Audiobook"` fixtures.
- **#1611** (umlaut transliteration) — re-reviewed and sound; your July 26 `CHANGES_REQUESTED` was stale against a commit the contributor has since revised. Left for the next release.

### Test plan

- [x] `make changelog` — `changelog.d/` empty apart from README
- [x] Rebased onto current `main`; CHANGELOG is the only file changed beyond the fragment deletions

Still no tag pushed. That is yours.

PR #1820, opened 2026-08-10
fix(telemetry-server): reliable daily activity via per-day ledger + real WAL pragmas

vavallee:
## Problem

The /stats daily charts showed sudden spikes (Aug 8: 207, Aug 9: 466) against a flat ~600-pings/day reality, and the nightly snapshot/sweep jobs failed intermittently with SQLITE_BUSY.

Two root causes:

1. **DSN pragmas never applied.** `?_journal=WAL&_timeout=5000` is mattn/go-sqlite3 syntax; the modernc.org/sqlite driver silently ignores it. The DB has run with journal DELETE and busy_timeout 0 since day one — any write/write overlap surfaced as SQLITE_BUSY (failed snapshots, failed sweeps, pings dropped with 500).

2. **Daily activity measured dormancy, not activity.** The chart grouped installs by `substr(last_seen,1,10)`; last_seen rolls forward on every ping, so each install counted toward only its most recent ping day. Historical days degraded to "installs that went dormant that day" (~5% of truth) while the newest 1–2 days absorbed everyone else — a permanent structural spike at the right edge that rolls forward daily. Version mix had the same flaw.

## Fix

- `sqliteDSN()`: `_pragma=journal_mode(WAL)`, `busy_timeout(5000)`, `synchronous(NORMAL)` — with a regression test that opens a file-backed DB and asserts `PRAGMA journal_mode` / `busy_timeout` actually took.
- **`daily_activity` ledger**: one `(day, install_id, version)` row per active day, written in the same transaction as the installs upsert. Never rolls forward — per-day distinct actives stay queryable forever (400-day ledger retention; `daily_global`/`daily_version` aggregates remain perpetual).
- Daily-activity chart, version-mix chart, and `snapshotDay` all read the ledger. The aggregate loop re-snapshots a trailing 7-day heal window, so a failed or missed tick self-corrects.
- One-time seed from `installs.last_seen` on first boot + `meta` epoch row; the dashboard footnotes pre-epoch days while the 30-day window still spans them (those days undercount — the per-day data was never recorded).
- `daily_features` membership joins the ledger instead of matching `last_seen`.
- Test fixture now shares production DDL via `createAggregateTables` so schemas can't drift.

## Notes for review

- Existing daily numbers for pre-ledger days are not rewritten — they were recorded under the old semantics and the true per-day counts are unrecoverable. The footnote covers the transition window.
- `handlePing` now does a 2-statement transaction instead of a single exec; with WAL + busy_timeout this is uncontended at current volume (~1 ping/2 min).

### DIFF
--- a/cmd/telemetry-server/main.go
+++ b/cmd/telemetry-server/main.go
@@ -183,7 +183,125 @@ func sweepStaleAndDev(ctx context.Context, db *sql.DB) (int64, error) {
 		return 0, err
 	}
 	n, _ := res.RowsAffected()
-	return n, nil
+	// Ledger rows age out on their own, longer, clock — they are the
+	// per-day history the aggregates are rebuilt from, not live state.
+	ledgerCutoff := time.Now().UTC().Add(-ledgerRetentionWindow).Format("2006-01-02")
+	resL, err := db.ExecContext(ctx,
+		`DELETE FROM daily_activity WHERE day < ?`, ledgerCutoff)
+	if err != nil {
+		return 0, err
+	}
+	nL, _ := resL.RowsAffected()
+	return n + nL, nil
+}
+
+// sqliteDSN builds the production connection string. modernc.org/sqlite
+// takes pragmas as `_pragma=name(value)` query params — the previous
+// `_journal=WAL&_timeout=5000` was mattn/go-sqlite3 syntax, which modernc
+// silently ignored, so the DB ran with journal DELETE and busy_timeout 0
+// and every concurrent write surfaced as SQLITE_BUSY (the nightly
+// snapshot/sweep failures and dropped pings). Applied per connection.
+func sqliteDSN(dbPath string) string {
+	return dbPath +
+		"?_pragma=journal_mode(WAL)" +
+		"&_pragma=busy_timeout(5000)" +
+		"&_pragma=synchronous(NORMAL)"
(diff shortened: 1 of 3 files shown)


vavallee:
Added historical recovery in 9781e222: the daily-activity history is reconstructed, not just fixed going forward.

Sources: the 60 archived daily `/api/backup` snapshots (2026-05-06 → 07-10, union of last_seen-day attribution) + Loki ping log lines (07-11 → 08-09; every 8-char log id prefix resolved to a full UUID against the snapshot+live install sets — 9020/9020, zero prefix collisions). 14,693 rows, 110 KB gzipped.

Loader: `LEDGER_BACKFILL_PATH` env → CSV(.gz) import at boot, re-derives `active_day`/`daily_version` per imported day, lowers the epoch (footnote retracts), meta-guarded idempotent, silent no-op when absent. The CSV contains install UUIDs so it ships as an out-of-band ConfigMap (`ledger-backfill`, mounted `optional`), never via git.

Dry-run against a copy of the production DB: 13,607 rows imported (rest deduped against the boot seed), epoch → 2026-05-06, reconstructed series matches the offline extraction exactly. True curve: ~85/day (June) → ~110 (early Jul) → 459 peak Jul 19 → ~210 plateau → 676 on Aug 8.

PR #1843, opened 2026-08-10
test(authors): make AddBook owner assertion non-vacuous

### LINKED ISSUE #1760: Vacuous owner assertion in the AddBook direct-insert test: passes even if tenancy inheritance is removed
Split out of the #1745 review. Test-only, no user-facing impact, but the assertion is currently worthless and reads as if it covers something.

## What

In `internal/api/authors_test.go` around line 4302:

```go
// Tenancy (#1457): the direct-inserted row inherits the author's owner.
if got.OwnerUserID != author.OwnerUserID {
    t.Errorf("book owner = %d, want the author's owner %d", got.OwnerUserID, author.OwnerUserID)
}
```

The `author` under test is built without an owner:

```go
author := &models.Author{
    ForeignID: "OL23919A", Name: "J. K. Rowling", SortName: "Rowling, J. K.",
    MetadataProvider: "openlibrary", Monitored: true,
}
```

So `author.OwnerUserID` is the zero value, and the comparison is `0 != 0`. The assertion passes no matter what the production code does. It would still pass if `primary.OwnerUserID = author.OwnerUserID` were deleted outright, or replaced with a hardcoded zero.

## Why it matters

Tenancy (#1457) is a cross-user isolation property — the class of thing where a test that cannot fail is worse than no test, because it makes the property look covered. `AddBook`'s direct insert is exactly where a book could be created under the wrong owner.

## Fix

Give the test author a non-zero owner and assert against that literal rather than against the author struct, so the assertion has something to distinguish:

```go
const ownerID = 7
author := &models.Author{..., OwnerUserID: ownerID}
...
if got.OwnerUserID != ownerID {
    t.Errorf("book owner = %d, want %d (inherited from the author)", got.OwnerUserID, ownerID)
}
```

Confirm non-vacuity by removing the `primary.OwnerUserID = author.OwnerUserID` line in `internal/api/authors.go` and checking the test fails.

Worth grepping the other tenancy assertions for the same shape while in there — comparing a result field against a struct field that is also zero is an easy pattern to repeat.

AshSgDe29071999:
## Summary
- Seed the AddBook direct-insert fixture author with `OwnerUserID: 7`.
- Assert the persisted book owner against that literal so removing tenancy inheritance fails the test.

Closes #1760

### DIFF
--- a/internal/api/authors_test.go
+++ b/internal/api/authors_test.go
@@ -4376,9 +4376,11 @@ func TestAddBook_ExistingAuthorDirectInsertWhenSyncSkipsWork(t *testing.T) {
 
 	ctx := context.Background()
 	// The author is already in the library, added long before this request.
+	const ownerID = 7
 	author := &models.Author{
 		ForeignID: "OL23919A", Name: "J. K. Rowling", SortName: "Rowling, J. K.",
 		MetadataProvider: "openlibrary", Monitored: true,
+		OwnerUserID: ownerID,
 	}
 	if err := authorRepo.Create(ctx, author); err != nil {
 		t.Fatal(err)
@@ -4453,8 +4455,10 @@ func TestAddBook_ExistingAuthorDirectInsertWhenSyncSkipsWork(t *testing.T) {
 		t.Errorf("explicit media type should win, got %q", got.MediaType)
 	}
 	// Tenancy (#1457): the direct-inserted row inherits the author's owner.
-	if got.OwnerUserID != author.OwnerUserID {
-		t.Errorf("book owner = %d, want the author's owner %d", got.OwnerUserID, author.OwnerUserID)
+	// Assert against the literal ownerID (not author.OwnerUserID) so a zero
+	// author owner cannot make this pass vacuously.
+	if got.OwnerUserID != ownerID {
+		t.Errorf("book owner = %d, want %d (inherited from the author)", got.OwnerUserID, ownerID)
 	}
 
 	// The explicit add must not resurrect anything the user did NOT pick:


vavallee:
Thanks for this, and sorry for the slow look. The change is right and I want it in.

Heads up on the red `validate (Go)`: it is not your PR being wrong, it is your PR working. Giving the fixture author a real owner so the assertion stops passing vacuously is exactly what #1760 asked for, and it turns out that a non zero owner also breaks an earlier precondition in the same test, before your assertion is even reached.

I isolated it by flipping only that one constant and changing nothing else:

| `author.OwnerUserID` | result |
|---|---|
| 0 | pass |
| 7 | FAIL, sync logs `added=0 skipped_language=1 skipped_junk=1 total=3` |

So an author with a real owner loses a book to the allowed languages filter that a zero owned author keeps. That is a real bug on any multi user install and I have filed it as #1872 with what I ruled out so far (profile resolution and `GetByID` are both owner agnostic, so it is not the obvious explanation).

Nothing for you to do here. I would rather land #1872's fix first and then merge this on a green run, so your test lands as a passing guard rather than a known red. Leaving this open and linked in the meantime. Good catch, this is the most useful kind of test change.

vavallee:
Saw you closed this. For what it's worth the change was right, and I'll take it once #1872 lands so it goes in on a green run. Reopen whenever suits you, or I'll ping you then.

vavallee:
Landed in #1898, closing out #1760.

You were right, and the red CI run was your PR doing its job. Setting `OwnerUserID: 7` on the struct doesn't stick, because `AuthorRepo.Create` writes NULL no matter what the field says. So the fixture ended up claiming owner 7 against a row that owned nothing, no `users` row 7 existed, and every book insert died on the `books.owner_user_id` foreign key. The `precondition: sync should have created the allowed sibling` failure was that, not anything you got wrong.

Chasing it down turned up a real bug behind it. `AuthorRepo.CreateForUser` was writing `owner_user_id` to the row but never setting it on the struct it returned, unlike the quality and metadata profile repos which both already did. The author sync stamps that struct's owner onto every book it creates, so an author added by a real user produced a whole catalogue of NULL owned books, which per user scoping treats as shared. On a multi-user install that meant one user's library was visible to everyone else. There's a migration in #1898 to repair the rows already written.

So your test found a tenancy leak, which is a better outcome than the vacuous assertion you set out to fix. Thanks for filing it.

The fixture needed a real `users` row plus `CreateForUser` to work, so I folded that into #1898 rather than asking you to rebase onto a fix that didn't exist yet. Your assertion is in there and it's genuinely non-vacuous now, verified by deleting the tenancy inheritance line and watching it fail.

PR #1852, opened 2026-08-10
fix(telemetry-server): gate the funnel cohort on v1.30.1, not v1.31.0

vavallee:
The setup-funnel fields (#1823) ship in **1.30.1**, a patch release — but `funnelMinVersion` was pinned to `1.31.0` on the assumption of a minor bump.

Left alone this fails closed and silently: `funnelCapable` would reject every install that actually reports funnel data, so the **Setup funnel** section on /stats would render its "no data yet" note forever while the data sat in the DB.

- `funnelMinVersion` → `{1, 30, 1}`. No logic change needed — `funnelCapable` already compares all three components — but the test now pins the `1.30.0`/`1.30.1` pair explicitly, because a major/minor-only comparison would admit the whole 1.30.x line, which is exactly the cohort pollution the gate exists to prevent.
- Both user-visible version mentions updated: the `/telemetry-fields` schema page and the empty-state note on `/stats`.
- The "funnel-incapable" fixture in `TestComputeSetupFunnel` is `1.30.0` — now the immediately-preceding release, so it tests the real boundary rather than a distant one.

### DIFF
--- a/changelog.d/funnel-min-version.md
+++ b/changelog.d/funnel-min-version.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Setup-funnel cohort gate corrected to v1.30.1** — it was pinned to 1.31.0 on the assumption the funnel fields would ship in a minor release. With the fields landing in 1.30.1, the old gate would have excluded every install that actually reports them, leaving the /stats setup-funnel section permanently empty.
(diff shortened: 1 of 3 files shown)


vavallee:
Re the pre-merge check: there is nothing to include retroactively. 1.30.1 is unreleased (latest tag is v1.30.0, 2026-08-07) and the production installs table has **0 rows** on 1.30.1, so the cohort is empty until that release ships — the funnel section will keep showing its "no data yet" note until real 1.30.1 installs start pinging.

That does mean the gate can't be verified against live data before release. The check that matters is after 1.30.1 is out: the funnel table should populate with a cohort matching the count of 1.30.1 installs, not the whole 1.30.x line.

PR #1863, opened 2026-08-10
fix(importer): refuse directory moves whose destination is inside the source

### LINKED ISSUE #1809: Reorganize hangs when destination is inside source
### Pre-flight

### Current behaviour

When reorganize attempts to move a book directory to a destination that is inside the source path, the copy-based move path recursively descends into the newly-created destination and gets stuck in a loop. The container/operation must be manually killed to prevent filling up drive.

### Expected behaviour

If the destination path is a child of the source, expected behavior is one of the following:

- Fail fast with a clear error when the computed destination is inside the source path.
- Complete the reorganize using a safe alternative mechanism (e.g., rename source to a temporary sibling, create destination parents, then rename into final destination), and attempting best-effort restore on failure.

### Steps to reproduce

1. Configure a naming/layout change that causes a tracked book folder's computed destination to be inside its current source directory. Examples:
   - Change layout from /library/Author/Title/ to /library/Author/Series/Title/ where a series name equals a book title.
   - Reorganize flat author dir /library/Author/BookTitle.m4b → /library/Author/BookTitle/BookTitle.m4b.
2. Trigger a reorganize that uses the copy-based move.
3. Observe the recursive copy descending into the newly-created destination and never completing, creating infinite child dirs.

### Trace logs

```text
No relevant DEBUG logging to add.
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v1.30.0

vavallee:
Fixes #1809.

## Root cause

Destination-inside-source makes the same-filesystem `os.Rename` fail with `EINVAL`, so the cross-filesystem copy path takes over — and `copyDirContext` creates the destination *inside the directory it is about to walk*, then recurses into its own output, one level deeper per pass, until the disk fills (or the path passes `NAME_MAX`; ~8.6s in the regression test's measurement).

## Fix

- `copyDirContext` refuses a nested destination before creating anything (`ErrDestInsideSource`). It is the shared engine behind `MoveDir`/`CopyDir`, so one guard covers every copy-based directory move.
- The reorganize preview classifies the containment case as an **error row** — the fail-fast the issue asked for; `applyOne` never reaches the move.

## Provenance, plainly

The patch was written by `claude-opus-5` running headless in a network-restricted, tokenless sandbox (the bindhelper phase-4 trial), against this issue's text marked as untrusted input. It was gate-checked there (gofmt/build/vet/lint/full tests), then human-reviewed and re-tested before this PR. Its own review notes, verified and kept:

- `dirContains` is **lexical only** — case-insensitive filesystems can still evade it; fixing that means teaching the copy engine to skip the destination inode, out of scope here.
- `moveDirCtx` still runs `MkdirAll(filepath.Dir(dst))` before the refusal, which can leave one empty intermediate dir for a *direct* `MoveDir` caller (reorganize itself now errors in the preview first).
- **`HardlinkDir` has the identical structural hazard**, unreachable from reorganize — deliberate follow-up, not scope creep.

## Tests

`TestMoveDirRefusesDstInsideSrc` (both `MoveDir` and `CopyDir`: sentinel returned, source intact, nothing created) and `TestReorganize_DestInsideSourceFolder` (preview reports `error`, apply moves nothing, `book_files` still points at the source). Both fail against unfixed code by reproducing the runaway.


vavallee:
Superseded by #1865, which covers the same fix more completely: the guard is applied to the copy, hardlink and move entry points rather than move alone, and it carries the end to end reorganize test plus the symlinked source and shared prefix sibling cases. Closing this one to keep a single review surface.

Issue #1731, opened 2026-08-04
Single-word titles accept releases that extend past the keyword: 'Treasure' auto-grabbed 'The Treasure of Khan'


vavallee:
### Problem

A book with a single-significant-word title accepts releases whose title extends past that word into a different book. Reported on Discord #bug-reports by ballflartcallmop (2026-07-27): their "Treasure" (Clive Cussler, Dirk Pitt 9) auto-grabbed and silently imported `Clive Cussler - Dirk Pitt Universe Bk 29 - The Treasure of Khan 64 kbps` — a different book they already tracked separately, which then received its own correct release minutes later. The wrong audiobook landed under the right book record with no error. Verified unchanged on `main` @ 96b70c73.

### Root cause

`titleMatchesResult` (`internal/indexer/searcher.go`), `case 1` branch: a single-keyword title requires only the keyword at a word boundary plus the author somewhere in the release. `The Treasure of Khan` + `Clive Cussler` satisfies both. Nothing checks whether the release title continues past the keyword into a longer, different title.

This is the residual case after #1539 and #1063: both tightened the multi-word paths with author corroboration, but here the author is **correct** — same author, different book — so author checking structurally cannot separate them. And `matchAnchored`, the #1539 remedy, only inspects words to the *left* of the match and never runs for single-keyword titles at all. The shortest titles get the weakest protection, and exposure grows with the size of the author's catalogue.

### Fix directions (reporter's, both sound)

1. For single-keyword titles, look right as well as left: reject when the keyword is immediately followed by connectives extending it into a longer title (`of khan`).
2. Stronger: cross-check against the library itself. "Treasure of Khan" was already a `books` row under the same author — if a release matches book A but matches sibling book B more specifically, don't accept it for A. That signal is local, free, and catches exactly the same-author collisions that author corroboration cannot.

Option 2 also generalises: it would protect multi-word titles that prefix a longer sibling title ("Ender's Game" vs "Ender's Game: Speaker for the Dead" style collisions), so it's worth considering as the primary fix with option 1 as the cheap first line.

Issue #1724, opened 2026-08-03
ISBN is stored but never passed to search, so the ISBN exact-match boost can never fire


vavallee:
### Problem

Bindery fetches ISBNs, stores them, and has an ISBN exact-match boost in the release ranker, but never connects the two. The boost cannot fire.

Reported on Discord by vegapunk as "ISBN is not even being fetched ever". The fetching part is not accurate, it is fetched. The consequence they were seeing is real: ISBN plays no part in matching a release to a book.

### Current state

The ranker has the comparison, at `internal/indexer/searcher.go:718`:

```go
if c.ISBN != "" && p.ISBN != "" && strings.EqualFold(p.ISBN, c.ISBN) {
```

The release side works. `internal/indexer/release.go:163-174` parses an ISBN out of the release name and normalises it to digits, with care taken over separators.

The criteria side is never populated. There are exactly two places that build a real `MatchCriteria`, and neither sets `ISBN`:

* `internal/api/indexers.go:342` (interactive search) sets `Title`, `Author`, `MediaType`, `ASIN`, `AllowedLanguages`, `AuthorAliases`, and `Year`
* `internal/scheduler/scheduler.go:567` (automatic grabs) sets the same list

So `c.ISBN` is always `""`, the condition short circuits on its first term every time, and the boost is unreachable. `ASIN` sitting right next to it in both constructions makes the omission look accidental rather than deliberate.

The data is available. `models.Book` carries `ISBNs []string`, `models.Edition` carries `ISBN13` and `ISBN10`, and Hardcover, Google Books, OpenLibrary and DNB all populate them.

### Proposed solution

Populate `crit.ISBN` in both constructions from the book's ISBN, then the existing boost starts working with no ranker changes.

Two things to decide:

1. **Which ISBN.** `Book.ISBNs` is a list and editions carry their own. The edition being targeted is the more precise choice, but the book-level list is what is reachable at both call sites today.
2. **ISBN-13 vs ISBN-10.** A release name may carry either, and they are different strings for the same book. `internal/isbnutil` already exists and should be checked for a conversion helper before adding one, so a book stored as ISBN-13 still matches a release named with its ISBN-10.

This also unblocks `SkipMissingISBN` in the companion metadata-profile issue, which is the other half of what was reported.

### Note on scope

[...]

Issue #1515, opened 2026-07-10
Add Delay to Search for Numerous Wanted Titles by Author


DigNits:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I checked the [roadmap](https://github.com/vavallee/bindery/blob/main/docs/ROADMAP.md) and this isn't already planned

### Problem

An issue I'm having is when I am monitoring an author for many books (30+), and click "search all wanted" under the author's metadata, Bindery does not download any books. This appears to be due to the burst of requests to Prowlarr which overloads it or times it out. Once this happens Bindery logs indicate there is no response from my indexers due to 30+ requests at once. Additionally, I do not have any limiter settings enabled in my Prowlarr indexers.

### Proposed solution

Can we add a delay between each books search when there are many? If I want 30 Brandon Sanderson books, could it search 1 (5 second timeout), then 2 (5 second timeout), etc? Hopefully this is feasible.

vavallee:
Makes sense. A burst of 30+ searches hitting Prowlarr at once is a real problem, and search all wanted should pace itself rather than fan out all at once. Tracking this as a wanted improvement: serialize or throttle bulk author searches with a gap between each. No timeline yet but it is a small, sensible change. Thanks for the clear writeup.

vavallee:
Fixed and merged. The bulk search paths (search all wanted, series fill, per-author auto-search, and the scheduled loop) now pace themselves with a 3 second gap between each book's search instead of firing as fast as slots free up, so a big author no longer bursts Prowlarr into dropping requests. A 30 book author will work through its searches over a couple of minutes in the background. Ships in the next release. Thanks for the clear report.

PR #1870, opened 2026-08-10
fix(indexer): stop the bulk wanted-search query flood (#1814)

### LINKED ISSUE #1814: “Search all wanted” floods Prowlarr with hundreds of queries and times out
### Pre-flight

### Current behaviour

Clicking **Search 26 wanted** for Lemony Snicket caused Bindery to send approximately 294 searches to DrunkenSlug through Prowlarr. This included repeated searches for the same books using different author/title combinations and 76 searches with an empty search term.

Prowlarr received the requests, but after approximately 60 seconds Bindery began reporting `context deadline exceeded` for the DrunkenSlug searches. No books were grabbed.

The Open Library author refresh completed successfully immediately before the search, so the metadata connection was working.

### Expected behaviour

Bindery should search each of the 26 wanted books without overwhelming Prowlarr or DrunkenSlug. Searches should use limited concurrency, avoid duplicate and empty-term queries, and complete without timing out.

### Steps to reproduce

1. Configure DrunkenSlug through Prowlarr as a Bindery indexer.
2. Open an author with multiple wanted books.
3. Click **Search 26 wanted** on the author page.
4. Check the Bindery and Prowlarr logs.
5. Observe hundreds of repeated indexer searches, including empty search terms.
6. After approximately 60 seconds, observe `context deadline exceeded` warnings in Bindery.

### Trace logs

```text
Bindery:


{"time":"2026-08-09T05:57:59.523794785-04:00","level":"INFO","msg":"fetching books for author","author":"Lemony Snicket","foreignId":"OL20588A"}

{"time":"2026-08-09T05:58:00.07320232-04:00","level":"INFO","msg":"refreshed author profile from metadata provider","author":"Lemony Snicket","foreignId":"OL20588A"}

{"time":"2026-08-09T05:58:58.34377418-04:00","level":"WARN","msg":"indexer search failed","indexer":"DrunkenSlug","error":"search: Get \"http://192.168.1.51:9696/2/api?apikey=REDACTED&cat=3010%2C3030%2C3040%2C3050%2C3060&limit=100&q=Hurry+Up+and+Wait&t=search\": context deadline exceeded"}

{"time":"2026-08-09T05:59:02.353197746-04:00","level":"WARN","msg":"indexer search failed","indexer":"DrunkenSlug","error":"search: Get \"http://192.168.1.51:9696/2/api?apikey=REDACTED&cat=3010%2C3030%2C3040%2C3050%2C3060&limit=100&q=All+the+dirty+parts&t=search\": context deadline exceeded"}

{"time":"2026-08-09T05:59:07.869870926-04:00","level":"WARN","msg":"indexer search failed","indexer":"DrunkenSlug","error":"search: Get \"http://192.168.1.51:9696/2/api?apikey=REDACTED&cat=7020&limit=100&q=IFFYThe+Basic+Eight&t=search\": context deadline exceeded"}


Representative Prowlarr entries:


[Info] ReleaseSearchService: Searching indexer(s): [DrunkenSlug] for Term: [Lemony Snicket The Carnivorous Carnival], Offset: 0, Limit: 100, Categories: [7020]

[Info] ReleaseSearchService: Searching indexer(s): [DrunkenSlug] for Term: [Snicket The Carnivorous Carnival], Offset: 0, Limit: 100, Categories: [3010, 3030, 3040, 3050, 3060]

[Info] ReleaseSearchService: Searching indexer(s): [DrunkenSlug] for Term: [The Carnivorous Carnival], Offset: 0, Limit: 100, Categories: [7020]

[Info] ReleaseSearchService: Searching indexer(s): [DrunkenSlug] for Term: [], Offset: 0, Limit: 100, Categories: [7020]

[Info] ReleaseSearchService: Searching indexer(s): [DrunkenSlug] for Term: [], Offset: 0, Limit: 100, Categories: [3010, 3030, 3040, 3050, 3060]


The captured 15-minute Prowlarr output contained approximately 294 DrunkenSlug search entries, including 76 entries with an empty search term.
```

### Deployment method

Docker (ghcr.io)

### Bindery version

`sha-52786f1` Commit: `52786f136451225eeded23b9b90250a837d46b27`

### Environment details

* Host OS: Unraid
* Deployment: Docker
* Bindery image: `ghcr.io/vavallee/bindery:latest`
* Bindery container status: Healthy, zero restarts
* Prowlarr container: `binhex-prowlarr`
* Indexer: DrunkenSlug through Prowlarr
* Metadata provider: Open Library
* Enrichment providers enabled: Google Books, Hardcover and DNB
* Bindery port: `8787`
* Prowlarr port: `9696`

vavallee:
## Summary

Clicking **Search 26 wanted** on an author page sent ~294 queries to one indexer over 15 minutes and grabbed nothing — everything past the first minute died on `context deadline exceeded`. This fixes the three things that multiplied the query count. The deadline itself is unchanged: it was the symptom, not the cause. Closes #1814.

## Where the multiplication comes from

The per-book cascade is still sequential and still stops at the first tier that returns results — that part is fine. The volume is the product of everything above it:

| Axis | Multiplier |
|---|---|
| Books in the sweep | 26 |
| Formats per book (`media_type = both` fires ebook and audiobook independently) | ×2 |
| Cascade tiers when nothing matches (`t=book` → `Surname Title` → `Author Title` → `Title`) | ×4 |
| Enabled indexers | ×1 here |

That is 208 queries for one clean sweep. The reporter saw 294 with 76 empty-term entries, which is ~76 cascades — one full sweep plus about half of another. The Open Library author refresh that finished at `05:58:00`, one second before the click, kicks off its own bounded auto-search over the same wanted books, so the two sweeps overlapped and re-asked the indexer the same questions.

On the 76 empty terms: those are the tier-1 **structured `t=book`** requests. They carry `title=` and `author=` but no `q=`, and Prowlarr renders `SearchCriteriaBase` as `Term: [{SearchTerm}]` where `SearchTerm` comes from `q` — so a perfectly well-formed book query logs as `Term: []`. 76 of them lines up exactly with 76 cascades, and `294 − 76 = 218` lines up with the ≤3 text tiers those cascades then ran. Bindery does also have a genuine empty-`q` path, fixed below, but it is not where all 76 came from — worth knowing before reading the Prowlarr log as 76 wasted requests.

## What changed

**1. Empty terms are never sent** (`internal/indexer/newznab/client.go`)

`Search` refuses a blank or whitespace-only query outright, and `BookSearch` skips the whole cascade when the title normalises to nothing searchable. `NormalizeQueryTitle` strips a trailing parenthesised qualifier, so a catalogue row titled `(German Edition)` or `(Unabridged)` — Open Library produces these — reduced to `""` and then ran four queries: `t=book` with `title=`, `q=Snicket `, `q=Lemony Snicket `, and a bare `q=`. Same for a blank title, and for punctuation-only titles.

The guard is `hasSearchableText` (at least one letter or digit), deliberately weaker than `SigWords`: `It` and `Us` are real book titles and must stay searchable.

**2. Identical queries are deduplicated** (`internal/indexer/newznab/querycache.go`, new)

A small per-client cache keyed on the fully built request URL with the api key redacted. The searcher already pools one `*newznab.Client` per (indexer URL, api key), so this is per-indexer for free, and the key is exact — only the same `t=`, same term, same category set is ever served from it.

It handles both shapes the report shows. Two catalogue rows for one work (`The Carnivorous Carnival` and `The Carnivorous Carnival (Deluxe Edition)`) normalise to a byte-identical cascade, and an overlapping sweep repeats every query the first one made. Concurrent duplicates wait on the in-flight request rather than issuing their own; a completed response is reused for 90 s.

| Property | Choice |
|---|---|
| TTL | 90 s — outlives one sweep (the bulk fan-out paces launches 3 s apart) |
| Bound | 4 MiB of response bodies, then expired entries are dropped, then all completed ones |
| Failures | never cached — a transient timeout must not become a 90 s outage |
| Caps / Test / Probe | **not** cached — an admin must see the result of a config fix on the next click |

An interactive search repeated inside the window is served from the cache too. That is the intended trade: an indexer's answer to the same question does not change in 90 seconds, and re-asking it is what flooded the reporter's indexer.

**3. The bulk fan-out bound** (`internal/api/bulk.go`)

`bulkSearchConcurrency` 8 → `authorAutoSearchConcurrency` (4). Concurrency was not unbounded — `RunBoundedPaced` has capped and paced these fan-outs since #1515 — but bulk at 8 was the outlier: per-author auto-search and series fill run at 4, the scheduled wanted loop at 2. It is also the one a user triggers by hand, potentially on top of a refresh's auto-search still running at 4, which is exactly what happened here. One slot is not one request, so 8 slots was 8 concurrent cascades against a single indexer.

Taking the bound from the codebase rather than inventing a number, per the sibling constants.

## Query count for the reporter's scenario

| | Before | After |
|---|---|---|
| One clean sweep, 26 books, dual format, no matches | 208 | 208 (unchanged — this is the cascade, and it is out of scope here) |
| ...with edition-duplicate rows in the catalogue | 208 + 8 per duplicate work | 208 (duplicates ride the cache) |
| ...with an unsearchable-title row | +8 per row | 0 per row |
| The observed overlapping second sweep | +86 | ~0 |
| **Reporter's observed total** | **~294** | **≤208, lower with duplicate or junk rows** |
| Peak concurrent queries to one indexer | 8 (12 with the refresh sweep) | 4 (still 4 with the refresh sweep — the duplicates collapse) |

The honest headline is roughly a 30% cut in raw query count plus a halving of the burst rate, and a repeat click now costs almost nothing. Getting below 208 means changing the four-tier cascade itself, which #1610 is already working in — deliberately untouched here.

Three other branches are live in `internal/indexer/searcher.go` (#1724, #1731, #1610). **This PR does not touch `searcher.go` at all** — the fan-out concern turned out to live entirely in the newznab client and the bulk handler, so the cascade, match logic and criteria construction are untouched and a rebase should be clean.

- `Searcher.SearchBook` still fans out one goroutine per enabled indexer with no cap. Bounded in practice by the admin-configured indexer count (single digits), and not what the reporter hit with one indexer, so left alone rather than editing the file three other branches are in.
- Nothing coordinates the bulk sweep with the auto-search a catalogue refresh starts. The query cache makes the overlap cheap rather than preventing it; a real fix would be a per-author "search in progress" gate.
- The report's `q=IFFYThe+Basic+Eight` shows a corrupt title in the catalogue (`The Basic Eight` is a real Daniel Handler novel). Metadata-ingestion problem, unrelated to search.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (`docs/ARCHITECTURE.md` indexer package row; no env/config/auth/Helm surface touched)
- [ ] Wiki pages updated if user-facing behaviour changed — n/a, no UI or workflow change

## Test plan

- [x] `go test ./... -count=1` (full suite green)
- [x] `go test -race ./internal/indexer/... -count=1`
- [x] `go build ./... && go vet ./...`
- [ ] `cd web && npm run build` — not run; no frontend changes in this PR

New tests, and the revert-check that each one actually bites:

| Test | Guards | Fails when reverted |
|---|---|---|
| `TestBookSearch_NeverIssuesEmptyTerm` | blank / whitespace / `(German Edition)` / `(Unabridged)` / `:` titles, with and without an author | 4 queries issued, want 0 |
| `TestSearch_EmptyQueryIssuesNoRequest` | the floor under every caller | 3 empty-term requests reach the indexer |
| `TestBookSearch_MissingAuthorStillSearchesTitleOnly` | control — an author-less book stays searchable | — |
| `TestSweep_IssuesNoEmptyTerms` | same, driven end to end through `Searcher` | empty `title=` and empty `q=` on the wire |
| `TestBookSearch_DedupesIdenticalQueries` | three rows for one work | 12 queries, want 4 |
| `TestBookSearch_CollapsesConcurrentDuplicates` | dedupe under a race | 24 queries, want 4 |
| `TestSweep_IssuesNoDuplicateQueries` | no repeated (term, categories) pair in a sweep | 8 pairs repeated ×2 |
| `TestSweep_QueryCountIsBounded` | **the count itself** — a realistic multi-book sweep issues exactly 24 | count jumps |
| `TestSweep_ConcurrentBooksShareQueries` | 6 overlapping sweeps | 192 queries, want 24 |
| `TestQueryCache_DoesNotCacheFailures` | a 500 must not stick for 90 s | — |
| `TestCaps_NotCached` | admin Test button stays live | — |
| `TestAuthorsBulk_Search_SharesTheSearchFanOutBound` | peak in-flight ≤ the sibling bound | peak = 8, want ≤ 4 |


vavallee:
Good catch, this was a real one. Fixed in 74141313: the TTL branch now checks `e.err == nil` before serving, so a failed entry found in that window falls through to `dropLocked` and gets refetched.

Added `TestQueryCache_FailedEntryStillInMapIsNotServedAsSuccess` for it. The window is not reachable by racing two real requests, so the test plants the entry directly in the state the write side leaves behind. Confirmed it fails without the guard: `a failed entry must not be served from the cache; fetch was never retried`.

The existing `TestQueryCache_DoesNotCacheFailures` never covered this because it is sequential, so the failed entry is already deleted by the time the second call runs.

PR #1873, opened 2026-08-11
fix(importer): guard the hardlink path too, and compare case-insensitively

vavallee:
Both follow-ups #1863's review deliberately left open. **Stacked on `fix/1809-reorganize-dest-inside-source`** — merge that one first.

## 1. `HardlinkDir` had the identical hazard

`hardlinkDirRooted` MkdirAlls the destination and then walks the source, so a nested destination recurses into the tree it is creating. Not reachable from reorganize (which never hardlinks), which is why it was out of scope for the fix — but the primitive is exported, and the guard is one comparison against a disk-filling failure.

## 2. `dirContains` now compares case-insensitively

The filesystem decides what "same directory" means and this code cannot see it. On APFS or a Windows bind mount, `/library/Author` and `/library/author` are **one directory** — so a case-only difference between source and destination walked straight past the guard into the runaway it exists to stop.

The cost on a case-sensitive filesystem is refusing a move between two genuinely distinct directories whose paths differ only in case. That's a pathological layout, and refusing it is the safe direction; the alternative is filling the disk.

## Tests

`TestHardlinkDirRefusesDstInsideSrc` — sentinel returned, **nothing created** (the guard runs before `MkdirAll`), source intact.

`TestDirContainsIsCaseInsensitiveToo` — eight cases including the ones that must stay `false`: same dir, sibling, **sibling differing only in case** (`/library/Author` vs `/library/AUTHORS`), parent, unrelated.

Full `internal/importer` package green (12.3s), `go vet` clean.

### DIFF
--- a/changelog.d/1809-hardlink-and-case-guard.md
+++ b/changelog.d/1809-hardlink-and-case-guard.md
@@ -1,0 +1,6 @@
+- **Directory moves and hardlinks refuse a destination inside the source**
+  ([#1809](https://github.com/vavallee/bindery/issues/1809)) — the copy-based
+  move and `HardlinkDir` both walked into the directory they were creating and
+  recursed until the disk filled. Containment is now checked case-insensitively
+  as well, so a case-only difference cannot slip past on APFS or a Windows
+  mount.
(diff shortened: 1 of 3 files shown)


vavallee:
Superseded by the case-folding PR against current main. #1865 already guards HardlinkDir at the entry point, so half of this was redundant, and the other half targeted dirContains rather than checkDestNotInsideSource.

PR #1900, opened 2026-08-11
fix(calibre): show edition names in rollback preview (#1896)

### LINKED ISSUE #1896: rollbackDisplayName looks up editions with book ID 0, so edition rows render blank
Found while fixing #1635. Pre-existing, unrelated to that fix.

`rollbackDisplayName` in `internal/calibre/import_rollback.go` resolves the human-readable name shown for each planned rollback action. Its edition branch calls (`import_rollback.go:645`):

```go
eds, err := editions.ListByBook(ctx, 0)
```

Book ID `0` is hardcoded. No book has id 0, so this returns an empty slice on every call, and the linear scan that follows for a matching edition id never finds anything. Edition actions almost certainly render with a blank display name in the rollback preview.

### Impact

Cosmetic but user-facing at a bad moment: the rollback preview is what someone reads to decide whether to undo an import, and blank rows make it harder to tell what is about to be removed.

### Note

The `0` looks like a placeholder that was never filled in — the surrounding code has the edition id available, so the fix is probably to look the edition up directly rather than listing a book's editions and scanning. Worth checking whether an edition-by-id repo method already exists before adding one.

floze-the-genius:
## Summary

- add a transaction-aware edition lookup by local ID
- use that lookup when labelling Calibre rollback actions so created editions no longer have blank display names
- cover found, missing, and real rollback-preview behavior and add the required changelog fragment

Closes #1896

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` is unchanged because there are no environment, config, or upgrade-path changes
- [x] Wiki updates are not required because no documented workflow or configuration changed

## Test plan

- [x] RED: new tests failed before the implementation because `EditionRepo.GetByID` did not exist
- [x] `go test -race ./internal/db ./internal/calibre`
- [x] `go test -coverprofile=coverage.out -covermode=atomic ./cmd/... ./internal/...`
- [x] `go vet ./...`
- [x] `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run --timeout=5m ./...` (`0 issues`)
- [x] `go run golang.org/x/vuln/cmd/govulncheck@d1f380186385b4f64e00313f31743df8e4b89a77 ./...` (no reachable vulnerabilities)
- [x] `make changelog`
- [x] `git diff --check`
- [ ] `make test` exercised the changed `internal/db` and `internal/calibre` packages successfully, but the untouched `internal/api` package exhausted the local 10-minute package timeout while applying SQLite migrations. The full non-race package set passed; CI shards API race tests with a 15-minute timeout.

Frontend and HTTP smoke suites were not run because this change touches only repository lookup and existing Calibre rollback-preview logic.

## AI assistance

Implementation and test preparation were assisted by OpenAI Codex. The final diff was independently reviewed against the transaction and rollback contracts.

### DIFF
--- a/changelog.d/1896-rollback-edition-display-name.md
+++ b/changelog.d/1896-rollback-edition-display-name.md
@@ -1,0 +1,5 @@
+### Fixed
+- **Calibre rollback previews show edition names**
+  ([#1896](https://github.com/vavallee/bindery/issues/1896)) — created-edition
+  rollback actions now look up the edition directly, so their display names are
+  no longer blank.
--- a/internal/calibre/import_rollback.go
+++ b/internal/calibre/import_rollback.go
@@ -642,16 +642,11 @@ func rollbackDisplayName(ctx context.Context, books *db.BookRepo, authors *db.Au
 		if editions == nil {
 			return ""
 		}
-		eds, err := editions.ListByBook(ctx, 0)
-		if err != nil {
+		e, err := editions.GetByID(ctx, entity.LocalID)
+		if err != nil || e == nil {
 			return ""
 		}
-		for _, e := range eds {
-			if e.ID == entity.LocalID {
-				return strings.TrimSpace(e.Title)
-			}
-		}
-		return ""
+		return strings.TrimSpace(e.Title)
 	default:
 		return ""
 	}
--- a/internal/calibre/import_rollback_test.go
+++ b/internal/calibre/import_rollback_test.go
@@ -139,6 +139,7 @@ func TestImporter_PreviewRollback_ReturnsExpectedActions(t *testing.T) {
 	}
 	// Preview must touch deletes for created authors/books/editions.
 	var sawAuthorDelete, sawBookDelete, sawEditionDelete bool
+	editionNames := make(map[string]bool)
 	for _, a := range preview.Actions {
 		switch {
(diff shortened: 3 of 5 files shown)


vavallee:
Merged, thanks. Clean fix and the reasoning in `GetByID` is the right one.

Routing it through `r.exec` rather than `r.db` is the part that matters and it would have been easy to miss. `rollbackDisplayName` gets called from the apply path too, where the repos are tx wrapped, so a pool read there would block behind the open SQLite writer instead of just returning stale data. Your comment says exactly that.

Also correct that the edition snapshot's `local_id` is the edition row id, so the list and scan was never needed even with a working book id.

One thing you might find interesting: if you swap `r.exec` back to `r.db` and run the package, it does not fail, it hangs until the test binary panics at the timeout. Worth knowing that is the shape that deadlock takes if you hit it again elsewhere.

Related and not yours to fix: `GetByForeignID` right above your new method still reads through `r.db`, so it is not tx aware despite `WithTx` existing on the repo. Nothing calls it from inside a transaction today so it is latent rather than broken. Filed separately.

Sorry for the overlap, I had someone looking at this in parallel and did not spot your comment claiming it until after. Yours got there first and is what shipped.

Issue #1729, opened 2026-08-04
Language filter compares raw provider codes: a profile allowing 'eng' rejects books arriving as 'en' or 'en-US'


vavallee:
### Problem

A metadata profile set to `allowed_languages=eng` rejects English books that arrive with the code `en` or `en-US`. The skip is `slog.Debug` ("skipping non-allowed-language book"), so at the default log level the book simply never appears. Reported on Discord #bug-reports by ballflartcallmop (2026-07-27, `main` @ 90a2334, 15 affected rows in their own library); verified unchanged on `main` @ 96b70c73.

This contradicts the v1.28.0 release notes for #1160, which say `dc:language` is canonicalised to ISO 639-2/B "along with provider-supplied codes". Only the EPUB path is: `NormalizeLanguageCode` has exactly one caller, `internal/importer/epubmeta.go:144`.

### Root cause

`IsLanguageAllowed` (`internal/models/language.go:77`) lowercases and trims, then does a raw `slices.Contains(allowed, code)`. The canonicaliser sits unused in the same file.

Concrete path: Google Books assigns `vi.Language` straight through (`internal/metadata/googlebooks/client.go:127`) and Google returns ISO 639-1 (`en`, `en-US`). The other providers only dodge this by hand-rolling their own normalisation — Hardcover via `hardcoverLanguageAliases`, OpenLibrary by stripping the `/languages/` prefix, DNB by emitting 3-char codes natively. The filter trusts its input to already be canonical, which is the actual defect; Google Books is just the provider that exposes it.

### Fix

Call `NormalizeLanguageCode` inside `IsLanguageAllowed`, on both the incoming code and the parsed allowed list. That makes the filter correct regardless of which provider supplied the value and gives the per-provider alias tables something to converge on.

### Regression guard

The reporter's follow-up points at the right home for the invariant: `internal/normdrift` exists to catch exactly this family (its package doc cites #940 / #871 / #1608 / #1642–#1647 as "the same shape"), but its eight properties cover the title, author, series and slug alphabets — language is the one member of the family it doesn't watch. The missing property: for any code `c` and allowed list `a`, `IsLanguageAllowed(c, a, _)` must agree with `IsLanguageAllowed(NormalizeLanguageCode(c), map(NormalizeLanguageCode, a), _)` — the filter invariant under its own canonicaliser.

Issue #1160, opened 2026-06-15
Detect and manage ebook language (read dc:language, optional content detection, editable + filterable)


vavallee:
## Source

Community ask (Reddit, Kedryn73, 2026-06-13): a user migrating ~500 books from Readarr/Calibre has many books **with no language tag** and, being Italian, wants to find which books are in the wrong language so he can re-download them. Bindery has no way to surface this today.

## Current state (verified)

Language handling is thin and metadata-only:

1. **Embedded tag is ignored on import.** `EpubMetadata` (`internal/importer/epubmeta.go`) extracts only Title/Author/ISBN from the OPF — it never reads `dc:language`, which most epubs carry. The OPF XML is already parsed here, so the tag is right there for the taking.
2. **`Book.Language` is populated from online metadata only** — Hardcover (`internal/metadata/hardcover/client.go:457`), OpenLibrary (`internal/metadata/openlibrary/client.go:310,530,567`), ABS, and Calibre import (`internal/calibre/importer.go:713`). Books those providers don't supply a language for stay blank.
3. **Language is not user-editable.** The book PATCH handler (`internal/api/books.go:~405`) accepts `monitored/status/mediaType/asin/narrator` only — no `language`. So a user can't even correct it by hand.
4. **No content-based detection** anywhere (no `lingua`/`whatlanggo`/equivalent).

So a user with files that lack provider-supplied language has no signal and no manual recourse.

## Proposed scope (three escalating pieces, each independently shippable)

- [ ] **Read `dc:language` from the epub OPF at import/scan time.** Add a `Language` field to `EpubMetadata`, parse the `dc:language` element, normalise to ISO 639-1, and use it to populate `Book.Language` when the metadata provider left it blank. Cheap, deterministic, covers the majority of well-tagged epubs. (Decide precedence vs provider language — probably provider wins when present, embedded tag fills the gap.)
- [ ] **Optional content-based detection fallback** for files with no embedded tag — a local statistical detector (e.g. `lingua-go`/`whatlanggo`) over a sample of extracted text. Offline, fast, no external/AI service required. Gate behind a setting since it costs a text extract per book. Store the result with a `language_source` marker (`metadata` | `embedded` | `detected`) so users know how confident to be.
[...]

PR #1941, opened 2026-08-12
fix(importer): trust the file's language over the provider's

### LINKED ISSUE #1933: Imported file's language is never compared to the catalogue's: a Spanish EPUB displays as English forever
## Symptom

A book whose file is Spanish displays as **English** in Bindery, permanently, with no signal anywhere that the two disagree.

Field case: *Redshirts* by John Scalzi. Grabbed 2026-04-22 from the release `John.Scalzi-Redshirts.2014.Spanish.Retail.EPUB.eBook-BitBook`, imported to `.../John Scalzi/Redshirts (2012)/Redshirts - John Scalzi.epub`. The book detail page shows `English` on 1.30.2. The EPUB is Spanish. The only trace of the truth is the release name in the history row.

The grab itself is a pre-fix leftover and is **not** what this issue is about: `d996c94e` (#1576, 2026-07-19) taught auto-grab to apply the metadata profile's `allowed_languages`, and `releaseLanguageTags` maps `spanish` → `spa`, so that release would be rejected against an `eng` profile today. Before that commit `searchAndGrabFormat` filtered only on the global `search.preferredLanguage`, and `indexer.FilterByLanguage` is a no-op unless that value is exactly `"en"` — an unsaved setting means an empty value and no filtering at all.

What is still broken is everything after the grab.

1. **The embedded tag is read only when the catalogue left the field blank.** `internal/importer/scanner.go:1188`:

   ```go
   fillLanguage := book != nil && book.Language == "" && !book.IsFieldLocked(models.BookFieldLanguage)
   ```

   The EPUB is opened for `dc:language` only when `fillLanguage` is true (`scanner.go:1206`). Redshirts already carried `eng` from OpenLibrary, so the file's `es` was never read, never compared, never stored. This is #1160's stated "provider wins when present, embedded tag fills the gap" rule, decided before anyone hit the case where the two disagree and the file is the one telling the truth.

2. **The parse is already there and costs nothing extra.** `internal/importer/epubmeta.go:139-144` reads the first `dc:language` from the OPF and normalises it to ISO 639-2/B via `models.NormalizeLanguageCode`, in the same pass that already extracts title/author/ISBN. The gate above is the only reason it goes unused.

3. **`Book.Language` is the *work's* language, not the *file's*.** Every writer is a provider (`hardcover`, `openlibrary`, `abs`, `calibre`) or the user. `web/src/pages/BookDetailPage.tsx:427` renders `book.language` directly; `:523` shows an amber "unknown" badge when it is empty. There is no state for "we hold a file that disagrees" — the display is confidently wrong rather than unsure.

4. **Nothing in the import path checks language at all.** `FilterByAllowedLanguages` is applied at search time only (`internal/scheduler/scheduler.go:621`, `internal/api/indexers.go:451`). A Spanish EPUB hardlinked or dropped into the download folder by hand imports silently against an `eng`-only profile on current `main`.

5. **Manual edits already lock the field.** `internal/api/books.go:510-512` sets `book.Language` and calls `book.LockField(models.BookFieldLanguage)`, so any automatic correction must respect the lock — the user's word beats the file, the file beats the provider.

## Proposed scope

- [ ] Read `dc:language` on **every** EPUB import, not only when `Book.Language` is empty. Same call, same pass, drop the `book.Language == ""` half of the `fillLanguage` gate; keep the `IsFieldLocked` half.
- [ ] When the embedded tag and the stored language disagree, **trust the file**: persist the embedded value via `BookRepo.SetLanguage` (`internal/db/books.go:736`). A provider's work-level language describes an abstract work with editions in many languages; the file on disk is a specific edition and is the thing the user opens.
- [ ] Record the correction as a history event (new constant alongside `internal/models/settings.go:79-88`) carrying both codes, so the change is auditable rather than a silent overwrite.
- [ ] Do nothing when the field is locked — log at debug and leave the user's value alone.
- [ ] Tests: provider `eng` + embedded `es` → stored `spa` and one history row; provider `eng` + embedded `en` → no write, no event; locked field → no write; non-EPUB → unchanged; unreadable/absent `dc:language` → unchanged.

## Out of scope

- Content-based detection for files with no embedded tag (#1160 item 2) — that needs a detector dependency and a setting; this issue is purely the tag that is already parsed.
- Blocking or failing an import on a language mismatch. Correcting the record is enough; the user decides whether to replace the file, and they can only make that decision once Bindery stops telling them the file is English.
- Audiobooks — no equivalent embedded tag worth reading.

## Relation to #1160

#1160 item 1 is what shipped as the empty-field backfill. This is the same item done for the case it deliberately skipped, plus the mismatch signal. #1160 stays open for content detection (item 2) and the language filter/sort on the Books page — language is already editable via PATCH (`internal/api/books.go:446`), so that part of item 3 is done.

vavallee:
Closes #1933.

## The defect

The language on a book page is `Book.Language`, written only by metadata providers. A provider describes an abstract *work*, which has editions in many languages; the file on disk is one specific edition. When the two disagreed, nothing noticed.

`internal/importer/scanner.go:1188` read the embedded `dc:language` only when the catalogue had supplied nothing:

```go
fillLanguage := book != nil && book.Language == "" && !book.IsFieldLocked(models.BookFieldLanguage)
```

So a Spanish EPUB imported against an English OpenLibrary record displayed **English** indefinitely. Reported from the field on *Redshirts* by John Scalzi, imported from `John.Scalzi-Redshirts.2014.Spanish.Retail.EPUB.eBook-BitBook` — the release name in the history row was the only trace of the truth anywhere in the product.

## The change

- The tag is read on **every** EPUB import. The parse already existed (`epubmeta.go:139-144`, same pass that extracts title/author/ISBN), so this costs one zip open on an import that is already touching the file.
- When the embedded value disagrees with the stored one, the file wins and the book is updated.
- A disagreement is recorded as a `bookLanguageCorrected` history event carrying `from` and `to`, so the change is auditable rather than a silent overwrite. The History page gets a label, an amber badge, and a `from → to` detail line; unknown types already degraded gracefully, but a curated label makes it look deliberate.

**Precedence is user > file > provider.** A hand-set language locks the field (`api/books.go:510-512` already does this) and short-circuits before the EPUB is opened.

## Three deliberate non-changes

- **A plain fill stays silent.** An empty catalogue language being populated is the #1160 backfill, not a correction — nothing disagreed. OpenLibrary supplies no work-level language often enough that an event per import would be pure noise.
- **Both sides are normalised before comparing.** The catalogue is *not* normalised on write, so a provider's `en` against a file's `eng` would otherwise read as a conflict and "correct" every English book to English on every import (#1729's problem, one layer up).
- **Nothing is written when the import failed.** The tag is read while the source file still exists, which is necessarily before we know whether anything landed, so the reconcile stays gated on `imported > 0`.

## Verification

`go test ./... -count=1` green; `golangci-lint run ./internal/...` 0 issues; `npx tsc --noEmit` clean; frontend 535 tests across 56 files pass.

Six mutations, each confirmed to fail the test that should catch it:

| Mutation | Test that fails |
|---|---|
| restore the `book.Language == ""` gate | `TestTryImportInternal_CorrectsLanguageWhenEpubDisagrees` |
| compare raw strings, no normalisation | `TestTryImportInternal_NoCorrectionWhenLanguagesAgree` |
| emit a correction event for a plain fill too | `TestTryImportInternal_FillEmitsNoCorrectionEvent` |
| ignore the field lock | `TestTryImportInternal_DoesNotOverwriteLockedLanguage` |
| drop the `imported > 0` guard | `TestTryImportInternal_NoLanguageWriteWhenNothingImported` |
| never persist, only log | three of the above |

Two of these did not fail on the first attempt and the tests were fixed, not the claim:

- The agree-test stored `"eng"`, which is already canonical, so it never exercised normalisation at all. It now runs over `en`, `eng`, and `en-US` — the shapes providers actually supply.
- The `imported > 0` guard had **no** test; a failed import happily rewrote the catalogue from a file that never landed. `TestTryImportInternal_NoLanguageWriteWhenNothingImported` covers it by rooting the library under a regular file, so every `MkdirAll` fails with `ENOTDIR` for any user.

## Not addressed here

#1933 also notes that nothing in the import path checks a file's language against the metadata profile at all, so a hand-dropped Spanish EPUB still imports against an `eng`-only profile. That is a policy question (block? warn? import and flag?) rather than a correctness one, and this PR deliberately only fixes the record being wrong.

### DIFF
--- a/changelog.d/1933-embedded-language-correction.md
+++ b/changelog.d/1933-embedded-language-correction.md
@@ -1,0 +1,24 @@
+### Fixed
+- **A book whose file is in a different language than the catalogue says now
+  gets corrected on import**
+  ([#1933](https://github.com/vavallee/bindery/issues/1933)) — the language on a
+  book page came from the metadata provider, which describes the abstract
+  *work*, not the file you actually hold. The embedded `dc:language` tag was
+  read only when the provider had supplied nothing at all, so a Spanish EPUB
+  imported against an English OpenLibrary record displayed "English"
+  indefinitely, with the release name buried in a history row the only hint
+  otherwise. The tag is now read on every EPUB import and, when it disagrees
+  with the stored value, the file wins: a work has editions in many languages,
+  but the file on disk is one specific edition and is the thing you open. The
+  correction is recorded as a **Language Corrected** history event showing both
+  codes, so "why does my English book read as Spanish" has an answer in the
+  place you would look for it rather than only in a log line.
+
+  Precedence is user, then file, then provider. A language you set by hand locks
+  the field, and a locked field is left alone — the EPUB is not even opened.
+  Filling a language the catalogue never had is unchanged and stays silent: it
+  is a gap being filled, not a disagreement, and OpenLibrary routinely supplies
+  no work-level language at all. Comparison normalises both sides first, so a
+  provider's `en` against a file's `eng` is not mistaken for a conflict, and
+  nothing is written when an import fails — a book that never landed must not
+  rewrite its own catalogue entry.
--- a/internal/importer/scanner.go
+++ b/internal/importer/scanner.go
@@ -566,6 +566,56 @@ func (s *Scanner) createHistoryEvent(ctx context.Context, eventType string, sour
 	}
 }
 
+// applyEmbeddedLanguage reconciles the language recorded for a book with the
+// one embedded in the EPUB just imported for it.
+//
+// Precedence is user > file > provider. A user's own edit locks the field and
+// the caller never even opens the EPUB. Otherwise the file wins: a provider
+// describes an abstract *work* whose editions exist in many languages, while
+// the file on disk is one specific edition and is the thing the user opens.
+// Before #1933 the provider won unconditionally, so a Spanish EPUB imported
+// against an English OpenLibrary record displayed "English" indefinitely, with
+// the release name buried in a history row the only hint otherwise.
+//
+// detected is expected already normalised to ISO 639-2/B (ReadEpubMetadata does
+// this); book.Language is normalised here before comparing because provider
+// values arrive in both two- and three-letter forms (#1729).
+//
+// Best-effort throughout: a persistence failure is logged and leaves the
+// existing value alone rather than failing an import that otherwise succeeded.
+func (s *Scanner) applyEmbeddedLanguage(ctx context.Context, book *models.Book, detected, sourceTitle string) {
+	if book == nil || detected == "" {
+		return
+	}
+	previous := models.NormalizeLanguageCode(book.Language)
+	if previous == detected {
+		return
+	}
+	if err := s.books.SetLanguage(ctx, book.ID, detected); err != nil {
+		slog.Warn("failed to persist EPUB-detected language", "bookID", book.ID, "language", detected, "error", err)
+		return
+	}
+	book.Language = detected
+
+	if previous == "" {
+		// The catalogue had nothing; this is the #1160 backfill, and there is
+		// no disagreement to report.
+		slog.Info("filled book language from embedded EPUB metadata", "bookID", book.ID, "language", detected)
+		return
+	}
+	// A genuine disagreement means the file is not the edition the catalogue
+	// described. That is worth a history row: it is the answer to "why does my
+	// English book read as Spanish", and the language field alone cannot say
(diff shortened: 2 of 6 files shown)


vavallee:
Before implementing verify the ways in which this change affects performance for large libraries.

PR #1961, opened 2026-08-13
fix(importer): scope the already-in-library check to the download's format (#1885)

### LINKED ISSUE #1885: Already-in-library check isn't format-scoped: audiobook import marks the pending ebook download of the same title as imported
### Summary

The "book already in library" short-circuit in the failed-import retry path is not format-scoped: importing the **audiobook** of a title caused the still-pending **ebook** download of the same title to be marked imported and abandoned. Since each title holds independent ebook/audiobook slots with separate pipelines, an import of one format should never satisfy a pending download of the other.

### Environment

Bindery `sha-90a07d0` (ghcr.io/vavallee/bindery:latest), Docker on Synology DSM 7.3, qBittorrent 5.2.3.

### Timeline (from logs, single auto-grab sweep)

```
17:42:05  auto-grabbing book   format=ebook      (seedpool)
17:42:19  auto-grabbing book   format=audiobook  (MyAnonamouse)
17:42:25  ebook import retry attempt=1/3 → fails (torrent still downloading, see #1884)
17:42:26  audiobook imported → /data/media/audiobooks/…/We Who Wrestle with God ()
17:42:40  ebook import retry attempt=2/3 →
          INFO "no book files at download path but book already in library — marking as imported"
```

The ebook download was still in progress in qBittorrent, but retry #2 saw the book present in the library — present only because the **audiobook** had just imported 14 seconds earlier — and marked the ebook download as imported. The ebook slot stayed on `Wanted`, the epub never existed, and no re-search was ever queued: from Bindery's perspective the download chain ended successfully.

### Expected

The already-in-library check in the retry path should compare against the **same format slot** as the pending download (ebook download → ebook file present?), not against the book row as a whole.

### Impact

Any dual-format auto-grab where the audiobook finishes before the ebook (the common case — audiobooks are frequently cross-seed/instant matches) silently drops the ebook. The user sees `Book Imported` in History and has no signal that one format was abandoned.

vavallee:
## Summary

The "no book files at download path but book already in library" short-circuit compared against the book row as a whole rather than the format slot the pending download is filling. On a `media_type=both` title that let one format's import close out the other format's download: nathang21's logs show the audiobook importing at 17:42:26 and the ebook's retry #2 at 17:42:40 logging `no book files at download path but book already in library — marking as imported` while its torrent was still downloading. The epub never existed, the ebook slot stayed `Wanted`, and the whole chain read as a success. Closes #1885.

## Implementation notes

`isBookAlreadyImported` now takes the download's format and consults only that slot. The format is resolved by the new `downloadFormat`, strongest signal first:

| Signal | Source | Notes |
|---|---|---|
| Explicit hint | the `formatHint` already threaded through `ImportFromPath` / `tryImportInternal` | a human or API caller declared the format outright |
| Book media type | `book.MediaType`, when it is `ebook` or `audiobook` | a single-format book has only one slot a download could fill, whatever the release title says |
| The release | every format token in `Download.Title`, plus the token recorded in `Download.Quality` at grab time | the only per-download signal a `media_type=both` grab carries |

The release tier deliberately does **not** trust `Download.Quality` on its own. `indexer.ParseRelease` keeps the first match in its `formatTokens` order, and that order lists every ebook container ahead of every audio one, so an audiobook shipped with the publisher's PDF booklet — `"… (Unabridged) [M4B + PDF]"`, an entirely ordinary release shape — is recorded as `pdf`. Reading that at face value on a `media_type=both` book whose ebook is already on disk makes the audiobook grab consult the **ebook** slot and close itself out as imported: #1885 with the formats swapped, produced by the code meant to prevent it. `formatForRelease` therefore reads every format the release names:

| Release names | Resolves to |
|---|---|
| only audio containers | `audiobook` |
| only ebook containers | `ebook` |
| both, with an audiobook marker (unabridged / abridged, an Audible ASIN, the word audiobook or audible) | `audiobook` — an audiobook with its booklet, not a coin toss |
| both, with nothing to tell them apart | unknown |
| nothing recognisable | unknown |

When none of the three tiers answer, the check declines to short-circuit rather than accepting either format. That is the deliberate asymmetry, and it is not free: declining sends the download down the ordinary no-book-files path into `importFailed`. It is still the right side to err on — being wrong the other way silently marks a still-downloading grab as imported and abandons the format with no signal at all, while being wrong this way is bounded and visible. The download either imports when its files land, or reaches `importBlocked` (via the retry budget if attempts keep failing, or via #1976's skip limit if its files never appear), where the queue shows the reason, **Retry import** is offered and the release can be grabbed again. The missing format also stays eligible for a re-search, since `inFlightDownloadStates` excludes `importFailed` (covered by the existing `TestWantedSearchQueue_SkipsInFlightGrabs`).

The token lists now live in `internal/indexer` (`ReleaseFormats`, `MediaTypeForFormat`, and `IsAudiobookFormat` exported) rather than being duplicated in the importer behind a keep-in-sync comment. The duplicate had already drifted: it carried `"opus"` and `"aac"`, which `ParseRelease` can never produce, and would have missed any token added upstream.

The two qBittorrent poll call sites hit the same short-circuit when `content_path` is unresolvable. They recover the format from the category the torrent sits under via the new `downloader.FormatForCategory`, the partial inverse of `ResolveCategory`: it answers only when the per-media-type category split is actually configured, and returns "" (unknown) for a client with one shared category or a cross-seed parked under a foreign category. That keeps the #769 duplicate-add guard working there for dual-format books.

- No schema change. The `Download` row still doesn't persist the media type it was grabbed for; the signals above cover the real cases without touching the table or every grab call-site.
- The #769 guard is preserved for every configuration it already covered — single-format books resolve exactly as before, so this is a pure narrowing for `media_type=both`.
- Untouched: the pair-gating and drop-folder paths, which were already format-scoped.

## Tests

`internal/importer/scanner_format_scope_test.go`:

| Test | What it pins |
|---|---|
| `TestTryImportInternal_AudiobookImportDoesNotCloseEbookGrab` | replays the issue timeline (audiobook imports, then ebook retry #2 against an empty download path) with and without a format token in the release title — the ebook must land in `importFailed` with the no-book-files message, no ebook file recorded, and the book still `NeedsEbook()` |
| `TestTryImportInternal_EbookImportDoesNotCloseAudiobookGrab` | the same race with the formats swapped, driven by two real `M4B + PDF` audiobook titles that parse to `Quality: "pdf"` — the audiobook grab must not be closed out against the already-imported ebook |
| `TestCheckQbittorrentDownloads_1885_SiblingFormatDoesNotCloseGrab` | same race on the qBittorrent poll path, format recovered from the torrent's category |
| `TestTryImportInternal_FormatScopedShortCircuitStillFires` | counterweight: a re-grabbed ebook whose files a prior import already moved still short-circuits to `imported` (#769 not disarmed) |
| `TestDownloadFormat` | the resolver's precedence table, including the booklet cases, the tie-breakers, and "unknown is not either format" |
| `TestReleaseFormats`, `TestMediaTypeForFormat` (indexer) | all tokens in a title, not just the first; and the token → slot mapping is total over the indexer's own list, so the old drift cannot recur |
| `TestFormatForCategory` (downloader) | category → format only when the split is configured |

`TestIsBookAlreadyImported_MediaTypeBoth` was rewritten: it previously asserted the old "either format counts" behaviour, which is the bug.

Each assertion was proved against a deliberate mutation: reverting `isBookAlreadyImported` to the format-blind switch fails the two #1885 tests and five subtests of the dual-format table; hard-coding the format to unknown fails the #769 guards; swapping the resolver's precedence fails the single-format-book case; and reverting the release tier to `MediaTypeForFormat(dl.Quality)` fails both swapped-format subtests and five `TestDownloadFormat` subtests.

## Interaction with #1976

Both branches touch `internal/importer/scanner.go`. Re-verified after this round of changes: `git merge` of #1976 into this branch auto-merges with no conflict, and the merged tree builds and passes `internal/importer`, `internal/api` and `internal/indexer`. The hunks are adjacent but disjoint.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (no env vars, config, or user-facing surface changed; the new exported symbols — `downloader.FormatForCategory`, `indexer.ReleaseFormats`, `indexer.MediaTypeForFormat`, `indexer.IsAudiobookFormat` — all carry godoc)
- [x] Changelog fragment updated (`changelog.d/1885-format-scoped-already-imported.md`)

## Test plan

- [x] `gofmt -l .` clean, `go vet ./...` clean
- [x] `go test ./cmd/... ./internal/...` (non-race) — all packages pass
- [x] Merge rehearsal against #1976 — clean; merged tree builds and `internal/importer` + `internal/api` + `internal/indexer` pass
- [ ] `cd web && npm run build` — not run; `web/` is untouched by this PR


vavallee:
Filed the follow-up this PR's format resolver implies: #1964 — `downloads` has no persisted media type, so `downloadFormat()` reconstructs the grabbed format from three fallback signals and the qBittorrent poll path leans on category configuration. Stamping `media_type` at grab time would make this exact. Additive, no backfill, and not a blocker for this PR.

PR #1963, opened 2026-08-13
feat(api): download logs from the UI; warn when Hardlink can't hardlink (#1903, #1720)

### LINKED ISSUE #1720: Import Mode selector doesn't warn when Hardlink is picked but the filesystem can't hardlink
### Problem

If you explicitly select Hardlink as your import mode and your setup cannot hardlink, the button turns green and nothing tells you.

Bindery already knows the answer. `/api/v1/system/storage` returns `hardlinkable` plus a `hardlinkReason` (`internal/api/storage.go:47`), and `GeneralTab.tsx:429` renders it as a green line or an amber warning. The problem is placement: that block lives in the Storage section, while the Import Mode selector is at `GeneralTab.tsx:190`, further up in File Naming. The warning is not visible at the moment you make the choice, and nothing connects the two.

This is the exact confusion that shows up in support threads. Docker setups with separate volume mounts for downloads and library look like sibling paths and are separate filesystems, so hardlink silently degrades to copy. Users conclude hardlinks are broken rather than that their mounts are wrong.

### Current state

The Import Mode help text at `GeneralTab.tsx:193` does say "Hardlink requires the download folder and library to be on the same filesystem/volume". That is static prose. It is not the live check, and it reads as a footnote rather than a warning that applies to you specifically.

Note that `auto` is the default and handles this correctly, hardlinking when possible and copying otherwise. So the broken path is narrow: it only affects users who deliberately picked Hardlink. Those are exactly the users who care about the outcome.

### Proposed solution

When `import.mode` is `hardlink` and `storage.hardlinkable` is false, show the amber warning and `hardlinkReason` inline under the mode buttons, not only down in the Storage section. The component already fetches `storage`, so this is a conditional render rather than new plumbing.

`GeneralTab.storageHealth.test.tsx` already covers both `hardlinkable: true` and `false` (lines 53 and 71), so the fixtures for the test exist.

Optional follow-up worth considering separately: the same check could warn on the `external` drop folder, which has a comparable "looks fine, silently does not do what you think" failure mode.

### LINKED ISSUE #1903: Add ability to download log files
### Pre-flight

### Problem

Request: Add a logs download button on :8787/settings?tab=logs
Since the container is  rootless I can't go to console to see stout or /var/log
Troubleshooting would be easier with the ability to download log files and submit to issues.

### Proposed solution

A couple of options, add a download button to the right of the existing clear filters button.  That way logs could be downloaded which reflect the settings of whatever the current filters are.
Or simpler add by the Log Retention" section. Perhaps with a clear or erase logs option.

### Alternatives considered

Only other thing I could think of would be to possibly add a syslog target for people who prefer to ship logs to Graylog rsyslog et cetera

### Additional context

_No response_

vavallee:
## Summary

Two small support-cost reducers, one commit each.

**Log download** — rootless containers give an operator nowhere to read stdout or `/var/log`, so every "attach your logs" reply turned into a thread. `Settings → Logs` now has a **Download** button next to *Clear filters* that saves the entries matching the current filters as a plain-text file. Closes #1903.

**Hardlink warning** — picking **Hardlink** on a setup that can't hardlink turned the button green and said nothing while imports quietly fell back to copying. The live check now renders under the mode buttons, where the choice is made. Closes #1720.

## #1903 — log export

New `GET /api/v1/system/logs/export`, mounted in the same admin-only group as `/system/logs` in `registerSystemLogRoutes`. Both handlers now build their `db.LogFilter` through one `parseLogFilter`, so the download matches the rows the table was showing instead of drifting from a second copy of the parsing. The frontend builds the list request and the download URL from the same `filterParams()` object for the same reason.

Decisions worth reviewing:

| Decision | Choice | Why |
|---|---|---|
| Format | Plain text, `ts LEVEL [component] message key=value` | The file exists to be pasted into a GitHub issue. NDJSON is more faithful, but nobody reads it in a bug report, and the chosen line format is exactly what the Logs tab already renders on screen. Attrs are logfmt-quoted and sorted so two exports diff cleanly. |
| Auth | `RequireAdmin` | Same data as `List`, which is admin-only because the stream is app-wide: other users' book titles, OIDC usernames, download names. A file makes exfiltration *easier*, not harder, so the export inherits the gate. Both route tests were extended to cover it. |
| Streaming | Page (1000 rows) at a time, flushed | A large log store must not be buffered in server memory. `ToTS` is pinned to now before paging, otherwise a line written mid-export shifts every later `OFFSET` page and duplicates rows (the query is `ts DESC`). |
| Cap | 50,000 rows | Stated in the UI hint *and* in the file footer (`# reached the 50000-entry export cap — …`). Truncation is fine; silent truncation is not. |
| Redaction | `httpsec.RedactSecrets` on messages and attr values | An export is meant to be attached to a public issue, and an indexer `apikey` can appear inside a logged URL. Paths and titles are *not* redacted — the docs say to skim the file before posting. |
| Filename | `bindery-logs-<UTC timestamp>.txt`, via the existing `contentDisposition` helper | Server-named, so a user with three exports can tell them apart. |
| Default range | Last hour, same as `List` | An unfiltered download matches the unfiltered table rather than dumping the whole retention window. Widen the range in the filter bar to get more. |

The button is a plain `<a download>` — the session cookie authenticates it and the server supplies the filename, so there's nothing to fetch in JS.

Also fixed on the way past: the From/To pickers sent a zone-less `datetime-local` value (`2026-08-12T14:03`) that the API's RFC3339 parse rejected, so **the date range was being silently ignored on the list request too**. They now send a real instant in the browser's zone. An export that ignored the date range would have been worse than no export, so it's in the same commit.

## #1720 — hardlink warning at the point of choice

`/system/storage` already returns `hardlinkable` + `hardlinkReason`, and `GeneralTab` already renders it — in the Storage section, several screens below the Import Mode buttons in File Naming. Conditional render only; no new plumbing. `auto` deliberately stays quiet: it decides per download and falls back correctly, so warning there would be noise on every split-mount install.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — `docs/Troubleshooting-Wiki.md` gained a "Collecting logs for a bug report" section, `docs/Storage-And-Hardlinks-Wiki.md` a note under the import-mode table
- [x] Wiki pages updated (both of the above are wiki pages)
- [x] Changelog fragments in `changelog.d/` (`1903-log-download.md`, `1720-hardlink-mode-warning.md`); `CHANGELOG.md` untouched

## Test plan

- [x] `go build ./...`, `go vet ./...`, `golangci-lint run ./...` — clean
- [x] `go test -race ./internal/api ./cmd/bindery` — 9 new `TestLogExport_*` cases (headers/filename, line format, filter parity, default window, redaction, newline escaping, announced cap, multi-page streaming with a duplicate check, ring-buffer fallback) plus the two `registerSystemLogRoutes` tables extended with the export route (admin allowed / non-admin 403)
- [x] `cd web && npm run typecheck && npm run lint && npm run build`
- [x] `cd web && npm test` — new `LogsTab.download.test.tsx` (4 cases: default URL, filters in the URL, RFC3339 date conversion on both paths, cap stated) and 3 new cases in `GeneralTab.storageHealth.test.tsx` (warns for hardlink+unhardlinkable, quiet when hardlinking works, quiet for auto)

Local note: `internal/api` and `internal/importer` exceed `go test`'s 10-minute default under `-race` on a loaded machine. CI already shards `internal/api` with `-timeout=15m`; both packages pass when given that budget (`internal/importer` 713s, `internal/api` with the new tests adding ~25s to the `api-l-q` shard).

- #1720 notes the same "looks fine, silently does not do what you think" failure mode for the `external` drop folder — worth its own check.
- A syslog/remote target, floated as an alternative in #1903, is not addressed here.

---

## Review round 2

All four findings confirmed by reading the code; all four fixed.

**To-only date range diverged from the table (medium).** Confirmed: `List` defaults to the last hour only when both bounds are zero, `Export` defaulted them independently. Both now default on the same condition, so a to-only request exports the same unbounded-below window the table shows, and no `from=` is invented in the header. `TestLogExport_ToOnlyRangeMatchesTheTable` fails without it.

**Redaction didn't cover everything the PR claimed (low).** Confirmed: `level`, `component` and attr **keys** were written raw. All of them go through `exportValue` now, and keys are quoted by the same rule as values. Still code-supplied, so this is the contract rather than an attacker path — the test inserts a component containing a newline plus a forged-looking record and asserts the export is still one line.

**Ring-buffer fallback misdescribed itself (low-medium).** Confirmed: `Ring.Snapshot` filters on level and limit only. The header now describes only what was applied, prints a `# NOT applied (...)` line naming the dropped filters, declares `# order: oldest first` (the ring path emits oldest-first while the DB path emits newest-first), and announces its own limit, which it never did.

**Pagination could duplicate or skip rows (low).** Confirmed and fixed properly rather than papered over: the export now walks by **keyset cursor on `(ts, id)`** instead of `OFFSET`, so an async insert landing mid-export can't shift later pages and retention pruning can't skip rows. `ORDER BY` gains the `id DESC` tiebreaker the cursor needs. As a side effect a 50k export is a linear index walk instead of ~1.25M index steps.

`TestLogRepo_KeysetCursorSurvivesConcurrentInsert` reproduces the original bug directly: it reads page 1, inserts a row inside the window, and shows the cursor returns the next two rows while the `OFFSET` equivalent re-reads one the caller already has.

**Cosmetics.** The download href is now built from the filters the current rows were fetched with, not from live input state, so typing a filter and clicking Download without pressing Search can no longer give a file filtered differently from the screen. That also fixes the related pre-existing bug at the level pills, where `fetchLogs` closed over the state from before `setLogFilter` and fetched the previously selected level. `en.json` no longer hardcodes "up to 50,000" — the server owns the cap and the file states it in its footer. An export that stops at a **caller-supplied** `limit` now says "stopped at the requested limit of N entries" instead of calling the user's own number a system cap.

**Not changed:** the `RequireAdmin` gate, the 50k cap's fail-closed behaviour and footer, the timezone fix, and the whole #1720 hardlink warning — reviewed as clean, no findings.

### Tests added

| File | Test | Proves |
|---|---|---|
| `internal/api/logs_export_test.go` | `TestLogExport_ToOnlyRangeMatchesTheTable` | A to-only export covers what the to-only table showed |
| | `TestLogExport_RedactsEveryPartOfTheLine` | Component newlines can't forge records; keys with spaces are quoted; values redacted |
| | `TestLogExport_RingHeaderDeclaresWhatItIgnored` | The ring header never claims a filter it didn't apply |
| | `TestLogExport_KeysetSurvivesIdenticalTimestamps` | A multi-page export over a same-timestamp run returns each row once |
| `internal/db/logs_test.go` | `TestLogRepo_KeysetCursorSurvivesConcurrentInsert` | The cursor doesn't repeat a row when one is inserted mid-walk |
| | `TestLogRepo_QueryOrdersTiesById` | Rows sharing a timestamp come back in a total order |
| `LogsTab.download.test.tsx` | `exports what the table is showing, not what has only been typed` | The href tracks applied filters |
| | `applies the level the pill just set, not the previous one` | The stale-closure pill bug |

Mutation-checked, each against its own fix — with one exception, stated honestly: `TestLogRepo_QueryOrdersTiesById` does **not** fail when the `id DESC` tiebreaker is removed, because SQLite's current plan happens to return ties in rowid-descending order. It is kept as correct-by-construction: the cursor requires a total order, and nothing guarantees that plan.

Two existing assertions were updated rather than kept: the download tests asserted the href tracked live input (the behaviour being fixed), and `TestLogExport_CapIsAnnounced` asserted the caller's own `limit=2` was announced as an "export cap".

### Test status

- `go test ./internal/api/ ./internal/db/` — pass
- `golangci-lint run` on the touched packages — 0 issues; `gofmt` clean
- `cd web && npx tsc --noEmit && npx eslint <touched> && npx vitest run` — 560 tests pass
- Not re-run locally under `-race` for `internal/api`: it exceeds the 600s per-package timeout on this box even on unmodified `origin/main`. CI shards it.


vavallee:
Good catch on the filter header — fixed in 879c5a1c. `component` and `q` now go through `exportValue` like every other value, so the search term gets the same redaction and newline escaping, and there's a test that a URL with an apikey pasted into the search box can't reach the header (or forge a log line with a newline).

Issue #1182, opened 2026-06-16
Download-client Test should validate path visibility, not just connectivity


vavallee:
## Problem

The download-client **Test** button (`DownloadClientHandler.Test`, `internal/api/download_clients.go:201`) verifies the API connection. A green result gives users confidence that imports will work, but it does **not** verify that Bindery can actually *see* the files the client writes. The most common new-user failure is exactly this: test passes, downloads complete, nothing imports, because `BINDERY_DOWNLOAD_DIR` doesn't point where the client writes completes (different mount / path). The failure is silent until they dig into a library scan or the queue.

This is a recurring support load (multiple Discord/Reddit reports, and it bit the maintainer's own deployment).

## What exists today

`internal/downloader/health.go` already has `checkQbittorrentCategoryPath` (`:171`) that maps a qBittorrent category's save path against `downloadDir`/`audiobookDownloadDir` (post-#700). But it is:
- **qBittorrent-only** (nothing for SABnzbd / NZBGet / Transmission / Deluge),
- a **background health check**, not surfaced in the explicit Test action a user clicks,
- a **mapping check**, not a proof that Bindery can stat/read a real completed path.

The handler already has the dirs via `WithStoragePaths(downloadDir, audiobookDownloadDir)` (`:85`).

## Proposed

Make Test (and the per-client health) assert path visibility end to end and report it loudly:
- Resolve the client's completed-downloads path (category save path for qBittorrent; the configured complete dir for SABnzbd/NZBGet) and confirm Bindery can `os.Stat` it / read it after applying any configured path remap.
- When it isn't visible, fail the Test with an actionable message ("Connected, but Bindery can't read the client's completed-downloads folder at <path>, configure a path remap or a shared mount") and link the storage-layout docs, instead of returning a green check.
- Generalise beyond qBittorrent.

## Why this first

It converts the single most common silent failure into a self-diagnosing error at config time. Highest-leverage fix in the path-mismatch class (see also the UI path-health panel and the chart-default work).

Issue #1019, opened 2026-06-07
No import from Deluge


HeisenbergK:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Hello,

I have downloaded torrents in deluge from the "Grab" function.
They reach seeding but even though my label is correct, mode set to copy, path is correct, bindery will not import to calibre. This used to happen in my usenet downloader but it was a path issue and I would get a log of import failed. Now I get no such thing. Just my torrent showing as 100% in bindery and no import triggered. Upon looking at the logs and queue I see the torrent marked as downloading even though I see it in deluge as seeding. 

### Expected behaviour

It should trigger import or give a failure log describing why it doesn't.

### Steps to reproduce

1. Add an author
2. Grab a book in torrent/deluge
3. Wait to reach seeding

### Trace logs

```text
06/07, 23:24:03	DEBUG		job: check downloads	
06/07, 23:23:48	DEBUG		job: check downloads	
06/07, 23:23:33	DEBUG		job: check downloads	
06/07, 23:23:18	DEBUG		job: check downloads	
06/07, 23:23:18	DEBUG		job: check stalled downloads	
06/07, 23:23:07	INFO		download grabbed	title="The Ex by Freida McFadden [ENG / EPUB]"
06/07, 23:23:04	DEBUG		indexer result	author= indexer=NZBgeek title=Freida.McFadden-The.Ex.azw3.epub.mobi
06/07, 23:23:04	DEBUG		indexer returned results	count=1 indexer=NZBgeek
06/07, 23:23:04	DEBUG		indexer query tier matched	count=1 tier=2
06/07, 23:23:04	DEBUG		indexer result	author="Freida McFadden" indexer=MyAnonamouse title="The Ex by Freida McFadden [ENG / EPUB]"
06/07, 23:23:04	DEBUG		indexer returned results	count=1 indexer=MyAnonamouse
06/07, 23:23:04	DEBUG		indexer query tier matched	count=1 tier=2
06/07, 23:23:03	DEBUG		indexer query	url="http://prowlarr:9696/4/api?apikey=%2A%2A%2A&cat=7020&limit=100&q=McFadden+The+Ex&t=search"
06/07, 23:23:03	DEBUG		indexer query tier fallthrough	items=0 tier=1
[...]

Issue #1006, opened 2026-06-05
Transmission downloads fail with "unsupported protocol scheme magnet"


jmr224:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Version: v1.16.1

Docker Swarm deployment.

Transmission client configured and connection test succeeds.

When grabbing torrent releases from The Pirate Bay, Knaben, etc.,
Bindery fails before Transmission receives the torrent.

Logs:

failed to send download
client_type=transmission
error=fetch torrent from indexer: Get "magnet:?xt=urn:btih:..."
unsupported protocol scheme "magnet"

Expected:
Bindery should pass the magnet URI directly to Transmission.

Observed:
Bindery attempts HTTP GET against the magnet URI.

### Expected behaviour

append returned id 0

### Steps to reproduce

failed to send to downloader: NZBGet rejected the download (append returned id 0) — check NZBGet's log for the reason. Common causes: disk full, write-permission on the intermediate or destination directory, NZBGet paused with quota reached, or invalid NZB content

### Trace logs

```text
failed to send to downloader: NZBGet rejected the download (append returned id 0) — check NZBGet's log for the reason. Common causes: disk full, write-permission on the intermediate or destination directory, NZBGet paused with quota reached, or invalid NZB content
```

### Deployment method

Docker (ghcr.io)

### Bindery version

latest

vavallee:
Confirmed and fixed in #1007.

Root cause: those public trackers serve an `http(s)` download link that redirects to a `magnet:` URI. Bindery fetched the link with Go's HTTP client, which tried to *follow* the redirect to `magnet:` and errored with `unsupported protocol scheme "magnet"` before anything reached Transmission. (Direct magnet links already worked; this is the HTTP-redirects-to-magnet case.)

The Transmission client now follows redirects manually and hands a redirected magnet straight to Transmission, matching how the qBittorrent client already behaves. Ships in the next release.

(Side note: the issue body also has some NZBGet error text mixed in from the template — the actual fault is the Transmission/magnet one, so that's what's fixed here.)

Issue #903, opened 2026-06-01
Bug Report: Single-file torrent import scans entire download directory


Bclark117:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

When a completed download is a **single-file torrent** (e.g. a standalone `.m4b` or `.epub` with no containing subfolder), Bindery imports the entire contents of the configured download directory rather than only the file(s) belonging to that torrent.

The affected book's library destination folder ends up containing unrelated books that happen to be sitting in the same download directory (retained there for seeding on a private tracker).

### Expected behaviour

Bindery should query the download client for the torrent's file list (e.g. via Transmission's `torrent-get` RPC) and import only the specific file(s) belonging to that torrent, regardless of what else is present in the download directory.

### Steps to reproduce

1. Configure Bindery with Transmission as the download client, with a shared download directory (e.g. `/data/completed/books`) that contains multiple completed/seeding books.
2. Grab a book where the torrent packages as a **single loose file** (no subfolder) — e.g. `Knife Edge (Noughts and Crosses #2) (Unabridged).m4b`.
3. Transmission places the file directly in the download directory alongside other seeding books.
4. Bindery detects the download as complete and attempts import.
5. Observe: Bindery copies a large number of unrelated books from the download directory into the newly created library destination folder.

### Trace logs

```text
logs show successful import
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v1.15.3

### Environment details

**OS/Architecture:** Linux (OMV NAS — x86_64)
- **Download client:** Transmission (`haugene/transmission-openvpn`) — hosted on separate machine (RPi5), accessed over local network
- **Indexer:** Jackett (Torznab) — MyAnonamouse
- **Reverse proxy:** None (direct access)
- **Note:** Download client and Bindery run on separate hosts with different volume mount points, path remap configured via per-client UI setting (`/data:/downloads`)

PR #1966, opened 2026-08-13
feat(downloader): add rTorrent/ruTorrent as a download client (#1618)

### LINKED ISSUE #1618: add rtorrent/rutorrent support
### Pre-flight

### Problem

Currently there's no support for rtorrent/rutorrent which seems like a major miss

### Proposed solution

Please add support for rtorrent/rutorrent as an download client.

### Alternatives considered

_No response_

### Additional context

_No response_

vavallee:
## Summary

Adds rTorrent/ruTorrent as a first-class download client, alongside qBittorrent, Transmission and Deluge. Grabbing, polling, importing, removing and the Test button all work; nothing about it is a stub. Closes #1618.

The unfamiliar part is the transport, as flagged in the issue: rTorrent is XML-RPC, not JSON over HTTP, so it shares no request plumbing with the other clients. Both of its deployment shapes are supported rather than assuming ruTorrent is present.

## Transports

| URL Base | Talks to |
|---|---|
| *(blank)* | `http(s)://host:port/RPC2` |
| `/plugins/rpc/rpc.php` | That exact path — for rTorrent it's the **full RPC path**, not a prefix |
| `scgi://` | rTorrent's `scgi_port` on the configured Host and Port |
| `scgi://127.0.0.1:5000` | That explicit address |
| `scgi:///var/run/rtorrent/rpc.sock` | That `scgi_local` unix socket |

HTTP is the primary path — it's what ruTorrent, Flood, the linuxserver image and every seedbox panel already expose, and it's the only one that can carry credentials or TLS, so it reuses the existing base-URL / basic-auth / SSRF / guarded-redirect plumbing. SCGI covers the plain-rTorrent case the issue called out; it's ~130 lines (netstring-framed headers, CGI-shaped reply) and has no auth and no TLS, which is the protocol rather than a gap here. That's stated in the UI hint and QUICKSTART.

## What works

| Capability | Status |
|---|---|
| Add by magnet | `load.start` |
| Add by `.torrent` | Bindery fetches the bytes and sends `load.raw_start` (#873 shape) |
| Indexer link that 30x-redirects to a magnet | Captured and loaded as a magnet (#1006) |
| Label | `d.custom1`, percent-encoded to match ruTorrent's convention |
| Download directory | `d.directory.set`, inverse path-remapped like qBittorrent's `savePath` |
| Live queue status | `d.multicall2`; progress and ETA derived (rTorrent reports neither) |
| Stall detection | `d.message` on an incomplete torrent |
| Import | `d.base_path` for the path, `f.multicall` for the authoritative file list (#903) |
| Remove | `d.erase` |
| Remove **with data** | See below |
| Test | `system.client_version` + a `d.multicall2` read + `directory.default` path visibility (#1182) |
| Per-indexer seed ratio (#883) | **Not applied** — see below |

## Two things rTorrent does differently

**`load.*` never returns an identifier.** `load.start` and `load.raw_start` both return `0` on success and carry no hash, so Bindery computes the infohash itself before submitting — the btih topic for a magnet (base32 spellings normalised to hex), SHA-1 of the bencoded `info` dict for a `.torrent` — then confirms the torrent appeared in the download list. A torrent whose hash can't be derived is refused rather than added blind, because Bindery would have nothing to poll and the grab would sit at "downloading" forever.

The bencode walk already existed in `internal/downloader/qbittorrent` (it recovers the hash from a 409 "already present"). Rather than copy 120 lines, it moves to a shared `internal/downloader/infohash` package and qBittorrent's two helpers become one-line delegates — its existing tests are untouched and still pass.

**`d.erase` deliberately leaves the payload on disk.** rTorrent has no delete-with-data command; the docs are explicit that "the data stored for the item is not touched in any way". So "remove with data" resolves `d.base_path`, applies the client's PathRemap, and deletes it from Bindery's side before erasing — the same resolution Sonarr's rTorrent client uses. It's guarded: a non-absolute path, a filesystem root or top-level directory, a symlink, or a path Bindery can't `Lstat` all refuse and log why, and the torrent is still erased so the download row doesn't strand. `TestResolveRtorrentDataPath` covers each refusal.

**Seed ratios are ignored, loudly.** rTorrent has no per-torrent ratio limit over XML-RPC — ratio handling lives in global groups in `.rtorrent.rc` — so a per-indexer override is logged and skipped rather than turned into a command that appears to work. Documented in QUICKSTART and the changelog fragment.

## Where it's wired

Every type-switch site, not just the five in the issue:

| File | Site |
|---|---|
| `downloader/adapter.go` | `IsTorrentClient` (the protocol predicate — everything downstream keys off it) |
| | `TestClient` |
| | `SendDownload` |
| | `RemoveDownload` |
| | `GetStalledIDs` |
| | `getTorrentLiveStatuses` |
| `downloader/clientcache.go` | `RtorrentFor` + the cached-entry union |
| `downloader/pathcheck.go` | `rtorrentPathVisibility` via `directory.default` |
| `db/download_clients.go` | credential hydration, **and both protocol-scoped queries** |
| `importer/scanner.go` | `CheckDownloads` dispatch |
| `importer/scanner_poll.go` | `checkRtorrentDownloads` + import-source resolution |
| `api/queue.go` | the "no torrent client configured" suggestion string |
| `web/src/util/protocols.ts` | `TORRENT_CLIENTS` |
| `web/src/pages/settings/ClientsTab.tsx` | type option, credential predicates, defaults, field hints |

Two of those were live traps rather than additions:

- `GetFirstEnabledByProtocol` / `GetEnabledByProtocol` hardcoded `type IN (?, ?, ?)` in two places. A client type wired everywhere else but missing there is simply never selected for a grab, and the failure reads as "no enabled torrent download client configured" while a healthy client sits in the list. Both now derive from one `torrentClientTypes` list, with a test that pins the full protocol→type mapping.
- `scanner_dispatch_matrix_test.go` used `"rtorrent"` as its stand-in for an *unrecognised* client type. That fixture would have quietly started exercising the real poller and stopped testing the #1019 safety net; it's now `"aria2"`, and rTorrent gets a real matrix leg.

## Scope decisions

- **`d.multicall2` is polled unfiltered in the importer.** rTorrent returns every torrent in one call regardless of label, and tracked downloads are matched by the hash on the `Download` row — so filtering by label there would buy nothing and cost correctness: a torrent relabelled in ruTorrent would look vanished and `blockStaleImportFailures` would terminally block it. `TestCheckDownloads_Rtorrent_ImportsDespiteLabelMismatch` pins that. The label filter is still applied for live status / stall detection, matching Transmission.
- **Path visibility is checked for rTorrent**, unlike Transmission and Deluge which stay `PathUnknown`. rTorrent exposes a single global `directory.default`, and it's very often on a seedbox — precisely the remote-instance case #1182 was filed for.
- **No schema change.** Transport selection rides on the existing `url_base` column, so there's no migration to coordinate with other in-flight work.

## Dependencies

**None added.** XML-RPC encode/decode is hand-rolled on `encoding/xml` (`rtorrent/xmlrpc.go`, 260 lines including doc comments). rTorrent's dialect is small — no `<struct>` arguments, no `dateTime.iso8601`, no introspection round-trips — and both maintained Go XML-RPC packages bring reflection-based marshalling and a much wider surface than the dozen calls here justify.

## Security

- No new endpoints. `/downloadclient/*` is already entirely `RequireAdmin` because `List`/`Get` return credentials; rTorrent reuses the existing `username`/`password` columns and adds no new response field, so there's no new leak surface.
- Indexer `.torrent` fetches go through `httpsec.DownloadFetchPolicy` with per-hop re-validation and `RedactURLError`, same as Transmission and Deluge.
- The RPC transport refuses any redirect off the configured endpoint (`TestCall_RefusesRedirectOffEndpoint`).
- Torrent names and labels are indexer-controlled and reach the XML document, so every argument is escaped through `xml.EscapeText` — covered by a test that asserts markup in an argument can't break out.
- `os.RemoveAll` on a remote-reported path is the one genuinely sharp edge; the guards and their tests are described above.

## Suggested review order

1. `internal/downloader/rtorrent/xmlrpc.go` — the codec, and `xmlrpc_test.go` which drives it from realistic on-the-wire fixtures
2. `internal/downloader/rtorrent/client.go` — the RPC surface; `AddTorrent` is where the hash derivation lives
3. `internal/downloader/rtorrent/scgi.go` — the second transport, self-contained
4. `internal/downloader/rtorrent.go` — `RtorrentStatus` and the delete-with-data guards
5. `internal/downloader/adapter.go`, `clientcache.go`, `pathcheck.go` — the switch sites
6. `internal/db/download_clients.go` — the protocol-query fix
7. `internal/importer/scanner_poll.go` — the poller
8. Frontend + docs

- The Readarr importer (`internal/migrate/readarr.go`) still only maps qBittorrent and SABnzbd — it doesn't recognise Transmission, Deluge or NZBGet either, so mapping rTorrent there is a separate, wider change.
- No post-import label / view (Sonarr's `MarkItemAsImported` pushes torrents into a `sonarr_imported` view). Bindery has no equivalent concept for any client today.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — `README.md`, `docs/QUICKSTART.md` (new rTorrent section), `docs/ARCHITECTURE.md`, `docs/DEPLOYMENT.md`, godoc on every new exported symbol
- [x] Changelog fragment added (`changelog.d/1618-rtorrent-download-client.md`); `CHANGELOG.md` untouched
- [ ] Wiki pages updated — none enumerate download clients

## Test plan

- [x] `go build ./...`, `go vet ./...`
- [x] `golangci-lint run ./...` — 0 issues
- [x] `go test ./cmd/... ./internal/...` — clean (this is the suite CI's gating `test` job runs)
- [x] `govulncheck ./...` — no vulnerabilities
- [x] `make smoke` — the real binary, download-client wiring being one of its stated triggers
- [x] `cd web && npm run typecheck && npm run lint && npm run build && npm test` — 553 passing, 0 errors (6 pre-existing lint warnings)
- [x] `go test -race ./cmd/... ./internal/...` — passes, **with a caveat**: `internal/api` under `-race` sits right at the 10-minute default package timeout on my machine and tripped it on some runs. It does the same on unmodified `origin/main` (verified by checking out `origin/main` in the same worktree and re-running `go test -race ./internal/api/` — also 600s FAIL), and this PR adds no tests to that package. Pre-existing, not a regression; flagging it rather than quietly re-rolling until it went green. CI shards that package, and all six `validate (Go race)` shards are green on this branch.

New tests:

| File | Covers |
|---|---|
| `rtorrent/xmlrpc_test.go` | encode (escaping, base64, argument types), decode (scalars, faults, nested arrays, empty views, untagged values), and an HTML page decoding as *not* a fault |
| `rtorrent/client_test.go` | URL/transport construction, both Test probes, the HTML-endpoint and 401 messages, multicall parsing and label filtering, short-row rejection, magnet/`.torrent`/redirect adds, hash derivation incl. base32, non-zero `load` returns, the magnet-pending vs file-missing split, `f.multicall`, idempotent erase, off-endpoint redirect refusal |
| `rtorrent/scgi_test.go` | netstring framing and header order, CRLF/LF/no-header replies, every `scgi://` URL-base spelling, and end-to-end calls against a real socket |
| `downloader/rtorrent_test.go` | status mapping, the delete-data guards, cache reuse + credential-rotation invalidation, and the adapter-level Send/Remove/Test/live-status/stall paths |
| `downloader/infohash/infohash_test.go` | bencode walk and hash normalisation |
| `importer/scanner_dispatch_matrix_test.go` | rTorrent leg of the #1019 matrix, plus the label-drift import case |
| `db/db_test.go` | full protocol→type mapping, rTorrent credential round trip |
| `web/.../SettingsPage.test.tsx`, `protocols.test.ts` | the new type in the form and the protocol map |


vavallee:
Review findings addressed in b259be78, 5d9fcdfe and 9fce7dd1.

## Fixed

| # | What | Where |
|---|---|---|
| F-11 | Bencode walk recursed with no depth cap — a crafted `.torrent` under the 50 MiB fetch cap overflowed the goroutine stack, which is `runtime.throw`, not a recoverable panic. Capped at 64. | `infohash.go` |
| F-13 | `FromMagnet` read only the first `xt` topic, so a v1/v2 hybrid magnet listing `urn:btmh:` first looked like it had no infohash. Every topic is examined now — this covers the qBittorrent path too, since both delegate to the shared package. | `infohash.go` |
| F-1 | `protocolMismatch.torrent` in `en.json` still said "qBittorrent, Transmission, Deluge". The inline default was the dead half. | `en.json` |
| F-15 | Remove-with-data skipped the global remap fallback that the importer and the Test button both apply. `RemoveDownload` now takes it and resolves through `remapClientPath`. | `rtorrent.go`, `adapter.go`, `queue.go` |
| F-21 | `RtorrentStatus` flagged any `d.message` as an error, including tracker chatter on a healthy seeding torrent. Now qualifies on `!Complete`, matching the two sibling call sites. | `rtorrent.go` |
| F-17 | Symlink guard only covered the leaf. The resolved path must now equal the path being removed, which rules out a symlink anywhere in the chain. | `rtorrent.go` |
| F-23 | `load.*` trailing args are commands and rTorrent splits their arg lists on commas — a directory with a comma silently truncated. Quoted the way ruTorrent's `addtorrent.php` does. | `client.go` |
| F-6 | Dropped `d.multicall2` / `f.multicall` rows are logged. An empty poll routes into `blockStaleImportFailures`, which terminally blocks `import_failed` downloads. | `client.go`, `xmlrpc.go` |
| F-7 / F-8 | `parseSCGIURLBase` sets the dialer timeout, and `scgi://host:` / `scgi://:port` are rejected at parse time instead of at dial time. | `scgi.go` |
| F-9 | The client form now warns when an `scgi://` URL base means the Use SSL / username / password it is showing will be dropped. | `ClientsTab.tsx` |
| F-4 | Ints that fit encode as `<i4>` (core XML-RPC) rather than the `<i8>` xmlrpc-c extension. | `xmlrpc.go` |
| F-14 | `snippet` truncates on a rune boundary — the result is persisted on the download row. | `client.go` |
| F-18/19/20 | The ignored seed-ratio warning names the client, label and release, and points at the indexer setting; `seedRatioHint` says rTorrent ignores it. | `client.go`, `en.json` |
| F-25 | Godoc no longer claims per-call SSRF validation. `DEPLOYMENT.md:386` is accurate as written — it lists rTorrent among admin-typed service URLs, which is what save-time `ValidateOutboundURL` does — so that line is unchanged. | `client.go` |
| F-26 | QUICKSTART says Host/Port are still required by the form for the unix-socket and explicit-address spellings. | `QUICKSTART.md` |

Also swept for the F-1 pattern across the whole diff, as asked. No other rTorrent-related inline default was stale. Rather than leave that as a one-off, `web/src/i18n/inlineDefaults.test.ts` now asserts every `t('key', 'default')` whose key exists in `en.json` carries the same text — the existing component tests mock `react-i18next` to return the fallback, so they pass either way and cannot catch this. It turned up three other stale defaults, all cosmetic ellipsis drift, now aligned.

## Declined, filed as follow-ups

- **F-2** Readarr import maps every non-qBittorrent client to `sabnzbd` → #1983. Pre-existing and wider than rTorrent, as the PR body already noted.
- **F-22** `expectedHint` is a dead parameter and the media type is hardcoded to ebook → #1984. Pre-existing on the qBittorrent path too; using it properly is a message-quality change, and deleting it cascades through `CheckCompletedPathVisibility`'s signature. Neither belongs in this PR.

## Tests

Every fix has a regression test, each verified against a deliberate mutation:

- `infohash_test.go` — depth boundary at and past the cap, a realistic multi-file torrent still hashing, hybrid magnets in both orders, v2-only still refused
- `rtorrent_test.go` — global remap applied and client remap taking precedence, symlinked-parent refusal, tracker chatter on a complete torrent as `seeding`
- `rtorrent/client_test.go` — command-arg quoting incl. comma and backslash, rune-safe `snippet`
- `rtorrent/scgi_test.go` — empty host/port, non-numeric and out-of-range ports, non-zero dialer timeout
- `rtorrent/xmlrpc_test.go` — `<i4>` encoding, dropped-row counting
- `web` — `inlineDefaults.test.ts`, `helpers.test.ts`

`gofmt`, `go vet`, `golangci-lint` (0 issues), `go test ./cmd/... ./internal/...`, `npm run lint / typecheck / build / test` (557 passing) all green locally. Per-package `-race` on `internal/api` and `internal/importer` still exceeds the default 600s timeout on this box, on unmodified `main` too — relying on CI's sharded race job, same caveat as the PR body.

vavallee:
CI note: the `lint` job is red, but not from this branch. The failing step is `govulncheck`, with six Go **standard library** advisories fixed in go1.26.6 against the pinned go1.26.5. The identical six fail on `main` @ `fa66aaa2` ([run 31755635600](https://github.com/vavallee/bindery/actions/runs/31755635600)) — `golangci-lint` itself passes here.

Filed as #1990 with the thirteen pinned locations that need the bump. Deliberately not bumping the toolchain on this branch: it touches `go.mod`, the Dockerfile digest and four workflows, and would collide with whatever lands on main.

Everything else on this branch is green — `validate (Go)`, `validate (frontend)`, `smoke`, the db and api-c-k race shards, and all of CodeQL / Semgrep / gitleaks / grype / dependency-review. Remaining race shards still running.

Issue #1889, opened 2026-08-11
Books dropped by the language filter are invisible: Debug-only log, nothing in the API or UI


vavallee:
Books dropped by the allowed-languages filter during author sync vanish with no user-visible trace.

`internal/api/authors.go:1634`:

```go
if !models.IsLanguageAllowed(b.Language, allowedLangs, unknownFail) {
    skippedLang++
    slog.Debug("skipping non-allowed-language book", "title", b.Title, "language", b.Language, ...)
    continue
}
```

The count is aggregated into a single Info line at the end of the run (`authors.go:1818`), and the per-book detail is **Debug only**. The API response says nothing. The UI says nothing.

### Why it matters

EricSilberberg (Discord, 2026-08-11, v1.30.1) had **65 books for one author** silently removed this way, and only found it because they went looking in the logs and happened to spot `skipped_language`. Flipping "When book language is unknown" from `fail` to `pass` brought them all back. See #1872 for the full report.

The filter is doing what it was configured to do. The problem is that a destructive-looking outcome (most of an author's catalogue missing) presents identically to "this author only has a few books", and the only way to tell the difference is to know the log line exists and to have Debug enabled.

### Suggested shape

Surface the skip count where the user actually is:

1. Return the skip counts on the refresh/sync response and show them in the UI ("12 books skipped: language").
2. Raise the aggregate to Info with the breakdown it already computes (`skippedLang`, `skippedJunk`, `skippedMediaType` are all tallied and only the total is logged).
3. Ideally make the skipped set inspectable, so a user can see *which* books and decide whether the profile is set the way they meant.

Same argument applies to `skippedJunk` and `skippedMediaType`, which are counted alongside it.

### Related

- #1872 — the same filter dropping books depending on the author's owner.
- #1723 — five metadata-profile filters that are never consulted at all.

PR #1968, opened 2026-08-13
fix: honor skipPartBooks during author sync

gchahcg:
## Summary

`SkipPartBooks` on a metadata profile was persisted to the DB and returned by every profile read, but nothing in author sync actually consulted it — box sets, omnibuses, signed-copy cartons, and slash-separated multi-title anthologies from OpenLibrary were cataloged exactly like real books, with no way to filter them despite the setting's name implying otherwise.

This wires a title-pattern filter into the existing `fetchAuthorBooks` filter chain (same place the junk-title and allowed-languages checks already live), gated on the author's resolved metadata profile via a new `resolveSkipPartBooks` helper that mirrors the existing `resolveAllowedLanguages` pattern. `AuthorSyncSummary` gains a `SkippedPartBooks` count so drops are visible in the author page's sync report the same way skipped-language/junk/media-type drops already are.

- **Added** — `partBookTitleRe` / `isPartBookTitle` in `internal/api/authors.go`: matches box set / boxed set / collection set / "collection of N" / omnibus / "carton of N signed copies" / "books N-M" title patterns, plus slash-separated anthology naming ("Title A / Title B / Title C").
- **Added** — `AuthorHandler.resolveSkipPartBooks`, resolving the author's effective metadata profile the same way `resolveAllowedLanguages` does, defaulting to `false` on any lookup failure so an unresolvable profile never causes unexpected catalogue loss.
- **Changed** — `fetchAuthorBooks`'s ingestion loop now drops a work when `skipPartBooks && isPartBookTitle(b.Title)`, right after the existing junk-title check.
- **Changed** — `AuthorSyncSummary.SkippedPartBooks` (+ folded into `SkippedTotal()`), logged alongside the existing skip counters.

`skipMissingDate`/`skipMissingIsbn` are separately no-op during cataloging too (only applied during search/download scoring), but that's a different code path and out of scope here — this PR is scoped to `skipPartBooks` only.

## Testing

- `TestIsPartBookTitle` — table-driven unit test against every real noisy title pattern this was built to catch (box sets, cartons, anthologies) plus the false-positive traps that motivated the pattern design ("Catch 22", "Fahrenheit 451" — a bare trailing number must not match).
- `TestFetchAuthorBooks_SkipsPartBooks` — full `fetchAuthorBooks` integration test (in-memory SQLite, stub metadata provider) verifying box-set/carton/anthology titles are dropped while real single-book titles that merely share a substring with the noise (e.g. a bare "Abaddon's Gate" next to a box set whose title contains "Abaddon's Gate") still come through.
- `TestFetchAuthorBooks_PartBooksKeptWhenSkipDisabled` — confirms the filter is opt-in: with the setting at its default (`false`), behavior is unchanged from before this PR.
- `gofmt -l`, `go vet ./internal/api/... ./internal/models/...`, and `go test ./internal/api/... ./internal/models/...` all pass.
- Additionally smoke-tested outside the unit suite: built the image via the repo's own multi-stage `Dockerfile`, ran it in an isolated container, enabled `skipPartBooks` via the API, and added a real author (James S.A. Corey) to trigger a live OpenLibrary sync — `8` of `91` works were correctly dropped as box-set/carton noise, `0` false positives among the `62` added.

## Test plan for reviewers

- [ ] Enable "Skip part books" on a metadata profile in the UI, add/refresh an author known to have box-set editions on OpenLibrary, confirm they're absent from Wanted and the sync summary reports a non-zero `skippedPartBooks` count
- [ ] Confirm a profile with the setting left off is unaffected (existing behavior)

### DIFF
--- a/changelog.d/skip-part-books.md
+++ b/changelog.d/skip-part-books.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Author sync now honors `skipPartBooks`** — the metadata profile setting previously had no effect during cataloging (only the field was persisted, nothing consumed it). Box sets, omnibuses, signed-copy cartons, and slash-separated multi-title anthologies are now filtered out of author sync/refresh when the setting is enabled, matching what the setting's name has always implied. `AuthorSyncSummary` gains a `skippedPartBooks` count alongside the existing skipped-language/junk/media-type counters so drops stay visible.
--- a/internal/api/authors.go
+++ b/internal/api/authors.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"regexp"
 	"slices"
 	"sort"
 	"strconv"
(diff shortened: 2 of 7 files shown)


vavallee:
Thanks for this, and for the Chaptarr research on the sibling PRs, that is genuinely useful and saves rederiving decisions someone else has already shipped.

Before any of the five land though, there is one thing that affects all of them.

The counters do not show up anywhere. `AuthorSyncNotice.tsx` sums only skippedLanguage, skippedJunk, skippedMediaType and skippedNotAccepted, and returns null when that total is zero. So a sync that drops books through any of the new filters renders nothing at all. That is the same end result as the setting being dead, just moved from "never runs" to "runs and says nothing". Needs adding there plus the type in api/authors.ts.

Related, could you add a sample of dropped titles the way skippedLanguageSample does? That went in for #1889 for exactly this reason, a bare count does not tell anyone which books vanished, and debug logs are not reachable in a rootless container.

That matters more for this one than the others because the regex catches real single books. A few I tried: The Omnibus of Crime (one Sayers volume), Omnibus Press Presents..., He/She/It (the Piercy novel), Rock/Paper/Scissors, and the Penguin Nietzsche that is titled The Anti-Christ / Ecce Homo / Twilight of the Idols. The slash branch is the loosest, any title with two or more slashes matches it. Requiring three or more segments with spaces around the separators would clear most of them.

Suggestion on sequencing: one small PR first that widens the notice and adds the sample, then the five filter PRs land behind it and each one is visible from day one. Happy to do that PR myself if you would rather get on with the filters.

Also worth knowing, all four of the filter PRs touch the same filter chain in authors.go and the same summary struct in models/author.go, so whichever merges first the other three will need a rebase.

gchahcg:
Thanks for the thorough review — here's the plan:

1. **Sequencing**: agreed. Splitting the notice/sample plumbing into its own small PR first (`AuthorSyncSummary` gains all five `Skipped*` fields + a `Skipped*Sample` for each, `AuthorSyncNotice`/`api/authors.ts` updated to actually render them) so this PR and the other three rebase onto it instead of each adding the same struct fields independently. Will link it here once it's up.
2. **Samples**: added for every filter, not just this one — same shape as `skippedLanguageSample`.
3. **Regex false positives**: fixed. "The Omnibus of Crime" / "Omnibus Press Presents..." (omnibus used as the title's own subject, not a bundle descriptor) and "He/She/It" / "Rock/Paper/Scissors" (unspaced "/" as the title's own punctuation) no longer match. One residual left as a documented gap rather than chased further: "The Anti-Christ / Ecce Homo / Twilight of the Idols" (a real single-volume edition bundling 3 distinct works, spaced, 3+ segments) is textually indistinguishable from a real anthology bundle — Hardcover's `IsCompilation` gets it right when configured; OpenLibrary-only installs keep this one gap. Also found and fixed two *new* false positives that a naive broadening of the bundle-detection patterns would have introduced ("Boxed In", and a bracketed edition note containing the word "Set").

Will push all of this once the notice PR is ready to link.

gchahcg:
Notice/sample plumbing PR is up: #2032. Rebased this branch onto it — every filter here (including this one) now populates its own sample, and #2008 stacks on top of this branch in turn since it touches the same file. Also fixed a lint finding in min-popularity surfaced during the rebase.

vavallee (commented):
The false positives I raised are fixed and the residuals are documented honestly — "The Omnibus of Crime", "Omnibus Press Presents…", "He/She/It" all pass now, with tests, and the two cases you can't distinguish from title text ("The New Turing Omnibus", "Thrown under the omnibus") are named in a comment and pinned as `true` in the table rather than quietly left to be rediscovered. That is the right way to ship a heuristic.

One thing I did not raise the first time, which changes what a false positive costs.

## A filtered title that is already in the library stops being refreshed

The filter is placed immediately before the language filter, at `authors.go:1858` on main. The branch that updates already-tracked books is at `:1894`, thirty-six lines further down:

```go
// Update ratings + genres on existing books, then skip further
existing, _ := h.books.GetByForeignID(ctx, b.ForeignID)
```

So a work that trips `isPartBookTitle` `continue`s before anything checks whether the user already owns it. Three consequences, in increasing order of annoyance:

1. The book is not deleted — it stays in the library, monitored, with its files. Good.
2. It silently stops receiving rating, genre, cover and series updates on every subsequent sync, because it never reaches the update branch.
3. It is counted in `skippedPartBooks` and named in the sample **every sync, forever**, so the author page permanently reports a book being "skipped" that the user is looking at in their own library.

For the Penguin Nietzsche case you documented, that is the actual user-visible outcome: not "it was not added", but "it is in my library, it never updates, and Bindery keeps telling me it skipped it".

This is not new — the language filter sits in the same position and behaves the same way — so I am not asking you to restructure the loop. But the two fixes are small and worth doing here rather than inheriting the shape five more times across #2005, #2006, #2007 and #2008:

- Hoist the `GetByForeignID` lookup above the profile filters, or
- Guard just this filter with a "not already tracked" check.

Either way the filter reverts to what its name implies: it screens works out of *discovery*, and does not quietly stop maintaining books the user has already accepted.

## Smaller notes

**The alternation is doing a lot of work in one regex** — nine keyword branches plus two slash-shape branches, with `hasNonLeadingOmnibus` and `hasJoinedBundleBracket` beside it. It is well commented and the table test covers it, so I am not asking for a rewrite. Worth knowing that the next person to add a pattern will not have your empirical evidence, and the comment does not say what corpus "confirmed empirically over several authors" was — naming it (which authors, roughly how many works) would make the next addition safer.

**`\bbooks?\s+\d+\s*-\s*\d+\b`** catches "Books 1-3" but also matches a legitimate title of the shape "Book 1-2" if a publisher numbers a single volume that way. Low risk, and the setting is opt-in; noting it so it is a known rather than a surprise.

With the already-tracked guard added, this is good to go once #2032 lands under it.

vavallee:

vavallee:
Rebased onto main after #2005, #2006 and #2007 landed. Squashed to one commit, authorship preserved.

Your part-book filter was written against the single-pass loop, and #2007 has since split that into two passes with a batched edition prefetch in between. I moved the check into the second loop so it sits under the same hoisted `existing` lookup as the other four. Same reasoning as on #2007: a work that fails both a language check and a part-book check now counts against skippedLanguage rather than skippedPartBooks, which no test depended on.

The three red checks on the old head were GitHub codeload returning 429 and 503 while downloading step-security/harden-runner, not anything in your diff. This push re-runs them.

api and metadata suites pass locally.

PR #1969, opened 2026-08-13
fix: series Fill picks the best title match, not the first

gchahcg:
## Summary

`ensureHardcoverCatalogBook`'s fuzzy title matcher (used by Series Fill to reuse an existing local book instead of creating a duplicate) returned on the *first* existing book scoring `>= 92` against `seriesmatch.TitleScore`, rather than the best-scoring one.

`TitleScore` takes the max of four fuzzy algorithms, two of which (`PartialRatio`, `TokenSetRatio`) score a target title that's fully contained as a substring of a longer title as a **perfect 100**, regardless of how much extra text surrounds it. That means a box-set/omnibus title like `"Expanse Hardcover Boxed Set: Leviathan Wakes, Caliban's War, Abaddon's Gate"` ties an exact match for `"Abaddon's Gate"` at exactly the same score. Whichever candidate happened to be iterated first won — in practice the earlier-created omnibus row — so Fill silently recorded the omnibus as satisfying that catalog slot and queued nothing for it, while the real single-book title stayed unmatched with **no error surfaced anywhere**. From the outside it just looks like Fill "did nothing" for that one book.

I found this live while testing the Enhanced Hardcover Series feature against a real author (James S.A. Corey / The Expanse) with some pre-existing box-set noise in the library. Before writing the fix, I probed `TitleScore` directly against the real strings to confirm this wasn't just a first-vs-best ordering issue:

```
Abaddon's Gate                                                                    score=100
Expanse Hardcover Boxed Set : ... Abaddon's Gate                                  score=100
Expanse Series, Collection Set of 3 Books. ... Abaddon's Gate                     score=100
```

Both box sets tie the exact match, confirming a naive "keep the best score" fix alone wouldn't reliably resolve this — it would still depend on iteration order among the tied candidates.

## Fix

- Track the best-scoring **non-excluded** candidate across the whole `existingByTitle` list instead of returning on the first one over the threshold.
- Break same-score ties by **title-length closeness to the target title** (`titleLengthCloser`) — an omnibus/box-set title is reliably much longer than the single book it happens to contain, so this deterministically prefers the real book.
- Left `seriesmatch.TitleScore` itself untouched — it's shared with Audiobookshelf import, and this fix is scoped to candidate *selection* at this one call site, not the scoring function itself.
- Preserved the existing exclusion semantics exactly: a non-excluded match still wins over an excluded one regardless of score, and `blockedByExcludedTitle` still prevents creating a new row when only excluded titles matched (unchanged behavior, just restructured around best-match tracking).

## Testing

- New regression test `TestSeriesFillPrefersExactTitleOverOmnibusOnScoreTie`, built directly from the live repro: creates an omnibus title *first* (so it gets the lower ID and would win under the old first-match-wins iteration order) and the exact-title match second, then asserts Fill queues and links the exact match, not the omnibus.
- Full existing `TestSeries*`/`TestSeriesFill*` suite (40+ tests, including every excluded-title, cross-provider, and light-novel-volume test) passes with no regressions.
- `gofmt -l`, `go vet ./internal/api/...` clean.

## Test plan for reviewers

- [ ] `make check`
- [ ] Link a series with a Hardcover catalog where the local library has an omnibus/box-set edition alongside (or instead of) a real single-book match; run Fill; confirm the real book gets queued/linked rather than the omnibus silently "satisfying" the slot


vavallee:
Good catch on the root cause, the substring scoring 100 thing is real and it has been quietly costing people Fill results for a while.

One problem with the tiebreak though. Closest length is not really a proxy for "not an omnibus", it is "shorter wins", so it breaks the other way round. If the slot is "The Way of Kings" and my library has both "Kings" and "The Way of Kings: The Stormlight Archive, Book One", both score 100 and this picks "Kings". Same shape with "Dune" capturing the "Dune Messiah" slot. Local titles pick up ", Book One" and "(Series, #2)" suffixes from Calibre and ABS all the time, so the right answer is very often the longer one.

So it turns an order dependent wrong answer into a reliably wrong answer in that case, which is arguably worse to debug.

Could you try CleanTitle equality first, then a non substring metric like Ratio, and keep length as the last resort? The omnibus loses all three and the suffixed real book wins the first. A test for the inverted case would be great too, the "The Way of Kings" versus "Kings" one is a good candidate.

Everything else looks right to me. Exclusion semantics are preserved, both call sites want the same thing, and the regression test genuinely fails without the change.

vavallee (commented):
Good fix, and the diagnosis is the interesting part: `PartialRatio`/`TokenSetRatio` scoring a contained substring as 100 means an omnibus title that *contains* the real title ties an exact match, and first-match-wins let the tie fall the wrong way silently. That is a hard bug to see from the symptom, which is just "the real book stayed unmatched".

`titleTiebreakWins` is the right call over raw length-closeness, and tier 2 earns its place — the note that a short wrong candidate can out-score a long correct one on plain `Ratio` is exactly the case naive length or ratio tiebreaking gets backwards.

Two things.

**The behaviour change is wider than the tie.** Previously the loop returned on the first candidate over 92; now it scans every candidate and links the best. For a series where two different real books both clear 92 against one catalogue title, the linked book can now change from "whichever came first in `existingByTitle`" to "whichever scores higher". That is the intended improvement, but it means an existing library can see a different link after upgrading, not only in the omnibus case. Worth a line in the changelog fragment so it is not a surprise.

**`blockedByExcludedTitle` now depends on scan order in a way it did not.** An excluded candidate sets the flag and continues, so if a non-excluded candidate later scores higher, the flag is set but `best != nil` and the function returns the link. Reading the caller, that appears harmless — the flag only matters when nothing was linked — but it is now possible to both link and have the flag set, which was not reachable before. Confirm the caller only reads it on the nil path, or clear it when `best` is chosen.

The only red check is `codecov/patch`, which is advisory here and not a merge gate.

vavallee:
Rebased onto main and pushed, no conflicts and no changes to your work. Tests green locally.

The tiebreak reasoning reads well now, particularly tier 2 existing because plain Ratio gets the short-wrong-candidate case backwards.

PR #1993, opened 2026-08-14
fix(downloader): surface configured path hints (#1984)

### LINKED ISSUE #1984: Download-client path check: expectedHint is a dead parameter and the media type is hardcoded to ebook
### Pre-flight

### Current behaviour

Two related loose ends in `internal/downloader/pathcheck.go`, both pre-existing and both found while reviewing #1966.

**1. `expectedHint` is a dead parameter.** `statRemappedPath(client, clientPath, expectedHint, globalRemap)` (`pathcheck.go:148`) never reads `expectedHint`. Both callers compute it:

```go
expected := ExpectedDownloadDirForClient(client, models.MediaTypeEbook, downloadDir, audiobookDownloadDir)
return statRemappedPath(client, savePath, expected, globalRemap)
```

so the work is done and thrown away. The warning message it was presumably meant to feed — "Bindery was configured to read from X" — is the single most useful thing to put in front of an operator whose path check just failed, so the intent looks right and the wiring is simply missing.

**2. The media type is hardcoded to ebook.** Both `qbittorrentPathVisibility` (`:117`) and `rtorrentPathVisibility` (`:88`) pass `models.MediaTypeEbook` unconditionally, so `audiobookDownloadDir` is threaded all the way down and never influences anything. For a client configured with a separate audiobook category (#700), the Test button's notion of "where Bindery expects the files" is the ebook dir regardless.

Neither causes a wrong verdict today — the stat is on the *client-reported* path, which is correct — so this is a message-quality and dead-code issue, not a correctness one. That is why it was left out of #1966 rather than fixed in passing.

### Expected behaviour

Either use `expectedHint` in the PathNotVisible message, or delete the parameter and the two computations feeding it. If it is used, resolve it per media type (or mention both dirs when the client has an audiobook category) instead of hardcoding ebook.

### Steps to reproduce

1. Configure a qBittorrent or rTorrent client whose completed-downloads path Bindery cannot read.
2. Press Test.
3. The warning names the client path and the remapped path, but never the directory Bindery was configured to read from — the value `expectedHint` was computed to carry.

### Trace logs

n/a — code-level finding, no runtime failure to trace.

### Deployment method

Source

### Bindery version

main @ v1.30.4

floze-the-genius:
## Summary

Use the previously dead `expectedHint` to make qBittorrent path-visibility warnings show the configured local download directory. When ebook and audiobook categories are separate, the warning lists both directories instead of hardcoding the ebook target. Closes #1984.

This changes warning text only. The visibility verdict still stats the same completed-download path, and NZBGet behavior is unchanged.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — no environment variables, config, or upgrade path changed; Godoc and a changelog fragment were updated
- [x] Wiki pages updated if user-facing behaviour changed — not applicable; no wiki page documents this warning

## Test plan

- [x] `go test -coverprofile=coverage.out -covermode=atomic ./cmd/... ./internal/...`
- [x] `go test -race -timeout=15m ./internal/downloader`
- [x] `go vet ./...`
- [x] `gofmt -l . | tee /dev/stderr | (! read)`
- [x] `golangci-lint run --timeout=5m` (v2.11.4)
- [x] `govulncheck ./...`
- [x] `make changelog`
- [x] Frontend checks not applicable — no files under `web/` changed

### DIFF
--- a/changelog.d/1984-pathcheck-hints.md
+++ b/changelog.d/1984-pathcheck-hints.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Download client path warnings** (#1984) — Show Bindery's configured ebook and audiobook directories when a qBittorrent completed-downloads path is not visible.
--- a/internal/downloader/pathcheck.go
+++ b/internal/downloader/pathcheck.go
@@ -94,8 +94,29 @@ func qbittorrentPathVisibility(ctx context.Context, client *models.DownloadClien
 	if savePath == "" {
 		return PathVisibility{Status: PathUnknown}
 	}
-	expected := ExpectedDownloadDirForClient(client, models.MediaTypeEbook, downloadDir, audiobookDownloadDir)
+	expected := qbittorrentExpectedHint(client, downloadDir, audiobookDownloadDir)
 	return statRemappedPath(client, savePath, expected, globalRemap)
+}
+
+// qbittorrentExpectedHint describes the local directories Bindery is configured
+// to read for the categories this client serves. When audiobooks use a separate
+// category, include both media-specific directories so a failed visibility check
+// does not misleadingly point only at the ebook directory (#1984).
+func qbittorrentExpectedHint(client *models.DownloadClient, downloadDir, audiobookDownloadDir string) string {
+	ebookDir := strings.TrimSpace(ExpectedDownloadDirForClient(client, models.MediaTypeEbook, downloadDir, audiobookDownloadDir))
+	hints := make([]string, 0, 2)
+	if ebookDir != "" {
+		hints = append(hints, fmt.Sprintf("the ebook download directory %q", filepath.Clean(ebookDir)))
+	}
+
+	if strings.TrimSpace(client.CategoryAudiobook) != "" && strings.TrimSpace(client.CategoryAudiobook) != strings.TrimSpace(client.Category) {
+		audiobookDir := strings.TrimSpace(ExpectedDownloadDirForClient(client, models.MediaTypeAudiobook, downloadDir, audiobookDownloadDir))
+		if audiobookDir != "" {
+			hints = append(hints, fmt.Sprintf("the audiobook download directory %q", filepath.Clean(audiobookDir)))
+		}
+	}
+
+	return strings.Join(hints, " and ")
 }
 
 func nzbgetPathVisibility(ctx context.Context, client *models.DownloadClient, globalRemap string) PathVisibility {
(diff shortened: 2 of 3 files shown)


vavallee:
Thanks for this, and thanks for picking it up so fast after it was filed.

Checked it against what #1984 actually reported and it does fix the dead parameter properly. `statRemappedPath` now reads `expectedHint` instead of accepting and ignoring it, and the doc comment was updated to match rather than left describing the old intent, which is the bit that made it dead in the first place. Scoping to warning text only, and saying so in the PR body, is the right call.

Two things worth knowing before this closes #1984 though, since #1984 covered a bit more ground.

**rTorrent still hardcodes ebook.** `pathcheck.go:89` in `rtorrentPathVisibility` still calls `ExpectedDownloadDirForClient(client, models.MediaTypeEbook, ...)` and passes that straight through, so an rTorrent user with separate audiobook categories gets the old single directory message. rTorrent is actually where #1984 came from, it was found while reviewing the rTorrent client, so leaving that one behind is a bit odd. `qbittorrentExpectedHint` looks like it would work unchanged for rTorrent if you wanted to rename it and call it from both.

**The audiobook save path is still never checked.** `qbittorrentPathVisibility` only looks up `client.Category`, so with separate categories the audiobook category's own save path is never stat'ed at all. The message now names the audiobook directory, but nothing has verified Bindery can actually read it. That is the "never used for a second check" half of #1984 and it is a real coverage gap rather than a wording one.

Neither is a regression and both are pre-existing, so I am happy for this to go in as an improvement on what is there. Two options and I do not mind which:

1. Merge as is and I will retitle #1984 to cover the remaining check, or
2. You extend it here to cover rTorrent and add the second stat.

If you fancy the second one, the rTorrent bit is nearly free and the actual check is the more valuable half.

vavallee (commented):
This is the fix for the first half of #1984 and it is the right one — `expectedHint` reaches the message, and `qbittorrentExpectedHint` naming both directories when a separate audiobook category is configured is better than what the parameter was originally going to carry.

Two things.

**The rTorrent caller still hardcodes the media type.** #1984 has two halves; this fixes the dead parameter and leaves `rtorrentPathVisibility` (`pathcheck.go:89`) calling `ExpectedDownloadDirForClient(client, models.MediaTypeEbook, ...)`. On an rTorrent client serving an audiobook category, the hint that now appears will name the ebook directory confidently and wrongly — which is slightly worse than the old behaviour of saying nothing. Worth doing here since you are already in the function, or say so and I will open a follow-up so #1984 does not close half-done.

**Heads up on where this sits.** Two issues filed this morning build directly on `statRemappedPath`: #2029 (run the visibility check for every client type instead of qBittorrent only) and #2031 (suggest and apply a path remap from the failing verdict). I flagged #1984 as blocking both, so this PR is further up the chain than its size suggests. The `expected`-hint text you are producing here is what #2031 would seed its suggestion from.

**Why it says BLOCKED, and it is not you.** The branch protection on `main` requires a `Sign-off` status check, and no such check ran on your head commit — there are 39 check runs on it and none of them is that one. Your commits are fine; the check simply never reported, which leaves the PR unmergeable regardless of content. That is our problem, not yours. A rebase on current `main` should pick it up; if it does not, we will sort it from this side.

PR #1994, opened 2026-08-14
fix(series): replace the GPL-3.0 fuzzy matcher with a first-party one (#1988)

### LINKED ISSUE #1988: GPL-3.0 dependency linked into MIT-licensed release binaries
## Problem

`go.mod` requires `github.com/creditx/go-fuzzywuzzy`, which is licensed **GNU GPL version 3** (verified against the module cache `LICENSE`). It is a direct, non-test dependency, imported at `internal/seriesmatch/match.go:12` and used at lines 73-76: `TokenSetRatio`, `TokenSortRatio`, `Ratio`, `PartialRatio`.

Go static linking makes every `bindery` binary a combined work with that GPL-3.0 code. We distribute those binaries through goreleaser releases and the `ghcr.io/vavallee/bindery` image, while `LICENSE` offers MIT. MIT is not a license we can grant for the combined work — GPL-3.0 §5 requires the whole to be licensed GPL-3.0 with corresponding source offered. Anyone redistributing under the stated MIT terms, which the README badge tells them to do, is exposed.

## Options

**Replace the dependency** (preferred). Note that `paul-mannino/go-fuzzywuzzy`, the obvious swap, is also GPL-3.0. Permissively licensed alternatives:

- `github.com/adrg/strutil` — MIT
- `github.com/hbollon/go-edlib` — MIT
- `github.com/agext/levenshtein` — Apache-2.0

None ships `TokenSetRatio`/`TokenSortRatio` directly. Those are tokenize + dedupe + sort + ratio over a base similarity metric, roughly 30 lines on top of any of the above.

**Relicense the project to GPL-3.0.** Technically cheap, but it is a one-way door for downstream users and needs agreement from every past contributor — which we cannot currently evidence, since there is no DCO or CLA in place.

## Scope

`safeFuzzyScore` in `internal/seriesmatch/match.go` already wraps each metric behind a `func(a, b string) int`, so the swap is contained to that file. The real cost is test re-baselining: similarity scores will shift, so `internal/seriesmatch` tests with hardcoded thresholds need to be re-derived, and match quality should be spot-checked against a sample of real series titles before merging.

## Done when

- [ ] `github.com/creditx/go-fuzzywuzzy` removed from `go.mod` and `go.sum`
- [ ] `internal/seriesmatch` builds against a permissively licensed metric
- [ ] Series-matching tests re-baselined and passing
- [ ] Match quality spot-checked against real series titles

vavallee:
## Summary

`github.com/creditx/go-fuzzywuzzy` is licensed **GPL-3.0** (confirmed against the `LICENSE` in the module cache — it reads "GNU GENERAL PUBLIC LICENSE Version 3"). It was a direct, non-test dependency behind `seriesmatch.TitleScore`. Go links statically, so every binary published through goreleaser and every `ghcr.io/vavallee/bindery` image was a combined work with that code, while `LICENSE` and the README badge offer MIT. MIT is not a licence we can grant for the combined work, and anyone who redistributed Bindery on the terms the README advertises was exposed.

Relicensing the project to GPL-3.0 would need agreement from every past contributor, which no DCO or CLA lets us evidence, so the dependency goes instead.

Closes #1988.

## Implementation notes

The four ratios `TitleScore` combines are now implemented in `internal/seriesmatch/fuzzy.go` from the published FuzzyWuzzy definitions, written from the algorithm rather than ported from any GPL source.

**No replacement dependency was added**, and that was the deciding factor rather than a preference for hand-rolling. Every permissive candidate in the issue (`adrg/strutil` MIT, `hbollon/go-edlib` MIT, `agext/levenshtein` Apache-2.0) exposes the usual `1 - d / max(len(a), len(b))` similarity. The FuzzyWuzzy ratio is a different number:

```
ratio(a, b) = 100 * (len(a) + len(b) - d(a, b)) / (len(a) + len(b))
```

where `d` weights insertion and deletion 1 but **substitution 2** — which is what makes the ratio readable as "share of characters that align". Building on any of those libraries would have shifted every score in the package, which is the opposite of what a licence fix should do. Writing the metric directly is ~150 lines, adds no transitive dependencies, and so adds no new licence surface at all.

It is also faster, which was not a goal but is a welcome side effect — the replaced Levenshtein allocated a full `[][]int` matrix per call where this one keeps two rows and trims the common prefix and suffix first. `TitleScore` over three representative pairs: **98,967 ns/op → 67,438 ns/op**, about 32% off.

## Behaviour: measured, not assumed

These scores gate real decisions, so parity was measured rather than inferred from a green suite:

| Threshold | Where | What it decides |
|-----------|-------|-----------------|
| `>= 85` | `internal/api/series.go:915` | Series Fill reuses an existing local book |
| `>= 92` | `internal/api/series.go:1399` | Two titles are the same book |
| `>= 70` | `internal/api/series.go:1136` | Position-matched candidate is accepted |
| `>= 88` | `internal/metadata/aggregator_canonical.go:748` | Canonical metadata title variant match |
| `>= 92` | `internal/abs/import_hardcover_series.go:486` | ABS/Hardcover series title match |

**Per-metric parity**, old implementation against new, over 43,660 pairs built from this repo's own fixtures and testdata:

| Metric | Pairs differing |
|--------|-----------------|
| `Ratio` | 0 |
| `TokenSortRatio` | 0 |
| `TokenSetRatio` | 0 |
| `PartialRatio` | 10.6% |

`PartialRatio` is the only metric that moves. The replaced implementation searched only windows anchored on a Levenshtein matching block; this one scans every equal-length window of the longer string. That means it finds windows the anchored search missed (scores go up), and it no longer inflates a score by comparing against a window truncated at the end of the string (scores go down). It also fixes a rune bug: the old code mixed byte offsets into rune slicing, so a title carrying non-ASCII characters could score **0** against a string that plainly contains it.

**End-to-end `TitleScore` over 19,306 pairs of real series and book titles** (Expanse, Stormlight, Mistborn, Discworld, Wheel of Time, Murderbot, Witcher, light-novel volumes, plus omnibus and boxed-set shapes):

| Threshold | Crossings |
|-----------|-----------|
| 85 | 0 |
| 88 | 0 |
| 92 | 0 |
| 95 | 0 |
| 100 | 0 |
| 70 | 2, both downward |

The two crossings at 70 are `"Iron Gold"` vs `"The Burning God"` (71 → 67) and `"The Return of the King"` vs `"The Way of Kings"` (70 → 67). Both are different books, so both are correct rejections that the old truncated-window inflation had pushed over the line.

**No threshold was moved and no existing test expectation was re-baselined.** `TestTitleScore`, `TestDifferentVolumes` and its precondition assertions, and the `internal/api` / `internal/metadata` / `internal/abs` suites all pass unchanged. The volume pairs `TestDifferentVolumes` documents still score exactly as its comment claims (100, 96, 93).

### Omnibus and boxed sets

The property that a fully contained volume title scores a perfect 100 is preserved exactly. Re-running the three strings from #1969's report:

```
Abaddon's Gate                                                        score=100
Expanse Hardcover Boxed Set: ... Abaddon's Gate                       score=100
Expanse Series, Collection Set of 3 Books. ... Abaddon's Gate         score=100
```

Identical to the old implementation. **#1969's premise and its length-based tie-break are unaffected** — that PR is still needed and still correct as written. Nothing on its branch was touched.

### One class of scores that rises

`partialRatio` scores a short string contained inside a *word* of a longer one as 100 — `"one"` inside `"someone"`. This is inherent to FuzzyWuzzy's `partial_ratio` and was **already present** in the replaced implementation: of twelve short-title probes, ten already scored 100 before this change. The exhaustive window scan makes it consistent rather than introducing it, and it adds two cases at 2–3 cleaned characters. No guard was added, since inventing one would be a matching-policy change smuggled into a licence fix; if it's worth addressing it deserves its own issue.

## Dependencies

`go.mod` / `go.sum`: `github.com/creditx/go-fuzzywuzzy` removed. Nothing added. `go mod tidy` clean.

## Checklist

- [x] Tests added or updated — `internal/seriesmatch/fuzzy_test.go` covers all four ratios, the distance weighting, symmetry, substring containment and multibyte handling
- [x] Doc-update gate cleared — no env vars, config, endpoints or Helm values touched; `docs/ARCHITECTURE.md`'s description of `seriesmatch` remains accurate; the README's MIT claim is now actually true
- [x] Wiki pages updated if user-facing behaviour changed — no user-facing behaviour change
- [x] `changelog.d/1988-gpl-dependency.md` added under `### Changed`

## Test plan

- [x] `gofmt -l .` clean
- [x] `go vet ./...` clean
- [x] `govulncheck ./...` — no vulnerabilities
- [x] `go test ./cmd/... ./internal/...` — all packages pass
- [x] `go test -race ./internal/seriesmatch/... ./internal/metadata/... ./internal/abs/...`
- [x] `go test -race -timeout 1800s ./internal/api/... ./internal/importer/...` — both pass (1573s and 384s respectively). These two exceed the default 600s per-package timeout on my machine even on unmodified `main`, so they were run individually with an extended timeout; CI's shards remain the authoritative result
- [x] `web/` untouched, frontend suites not applicable


vavallee:
Heads up on the red `lint` check — it is not this PR.

`main` does not currently compile its `internal/importer` tests: #1961 and #1962 landed about five minutes apart and both declare `dualFormatFixture` in package `importer`, so `go vet ./internal/importer/` fails on `origin/main` itself. Every open PR inherits it. #1995 is the fix.

Confirmed independent of this change: the diff here touches only `internal/seriesmatch`, `go.mod`/`go.sum` and `changelog.d/`. I will rebase once #1995 is in.

`go test -race -timeout 1800s ./internal/api/... ./internal/importer/...` was run locally against a tree from before that collision and both packages pass (1573s and 384s).

Issue #1980, opened 2026-08-13
Four metadata-profile filters are write-only: MinPopularity, MinPages, SkipMissingDate and SkipMissingISBN are editable, persisted, displayed, and never consulted


vavallee:
## Summary

Four of the seven metadata-profile filters are **write-only**. The UI offers them, the API persists them, a read returns them, the profile card renders them as active badges — and nothing in the codebase ever consults them to make a decision.

| Field | Editable in UI | Consulted anywhere |
|---|---|---|
| `MinPopularity` | yes | **no** |
| `MinPages` | yes | **no** |
| `SkipMissingDate` | yes | **no** |
| `SkipMissingISBN` | yes | **no** |
| `SkipPartBooks` | yes | no — being wired by #1968 |
| `AllowedLanguages` | yes | yes |
| `UnknownLanguageBehavior` | yes | yes |

## Evidence

Every non-test reference to the four dead fields, across `internal/` and `cmd/`:

```
MinPopularity   internal/db/metadata_profiles.go:95,119,140   internal/models/metadata_profile.go:16
MinPages        internal/db/metadata_profiles.go:95,119,140   internal/models/metadata_profile.go:17
SkipMissingDate internal/db/metadata_profiles.go:96,120,147   internal/models/metadata_profile.go:18
SkipMissingISBN internal/db/metadata_profiles.go:96,120,148   internal/models/metadata_profile.go:19
```

That is the struct definition plus INSERT, UPDATE and row scan. Persistence and serialization only. By contrast `AllowedLanguages` has five consuming references (`internal/api/indexers.go:555`, `internal/api/authors.go:2764`, `internal/indexer/debug.go:92`, …) and `UnknownLanguageBehavior` one.

The frontend exposes all four as real controls — `web/src/pages/settings/MetadataTab.tsx:288-291` holds them in form state, and lines 251-257 render `minPopularity`, `minPages` and badges for `skipMissingDate` / `skipMissingIsbn` on the profile card. So a user sets a 200-page minimum, sees "min pages 200" on the card, and gets no filtering at all.

## Why this matters more than four unused columns

This is the failure mode that makes people distrust the whole app. The setting is not missing, greyed out, or documented as unimplemented — it is presented as working and confirms itself back to you. When it does nothing, the reasonable conclusion is that Bindery's filtering is broken generally, not that these four specific fields are inert.

It is also a **known recurring class here**, which is the part worth acting on:

[...]

PR #2005, opened 2026-08-14
fix: honor skipMissingDate during author sync

gchahcg:
## Summary

`SkipMissingDate` on a metadata profile was persisted to the DB and returned by every profile read, but nothing in author sync actually consulted it (see #1980) — works with no release date were cataloged the same as any other, with no way to filter them despite the setting's name.

This wires a title-agnostic date-presence check into the existing `fetchAuthorBooks` filter chain (same place the junk-title and `SkipPartBooks` checks live), gated on the author's resolved metadata profile via a new `resolveSkipMissingDate` helper that mirrors the existing `resolveAllowedLanguages`/`resolveSkipPartBooks` pattern. `AuthorSyncSummary` gains a `SkippedMissingDate` count so drops are visible in the author page's sync report the same way the other skip counters already are.

- **Added** — `AuthorHandler.resolveSkipMissingDate`, resolving the author's effective metadata profile the same way `resolveAllowedLanguages`/`resolveSkipPartBooks` do, defaulting to `false` on any lookup failure so an unresolvable profile never causes unexpected catalogue loss.
- **Changed** — `fetchAuthorBooks`'s ingestion loop now drops a work when `skipMissingDate && b.ReleaseDate == nil`, right after the junk-title check.
- **Changed** — `AuthorSyncSummary.SkippedMissingDate` (+ folded into `SkippedTotal()` and the Warn-level log gate, alongside the other metadata-filter counters), logged as `skipped_missing_date`.

### Why this one and not `SkipMissingISBN`

#1980 groups four write-only filters together, but they're not the same shape of fix. I traced both before touching either (full writeup in a comment on #1980):

- `SkipMissingDate` is safe to filter at cataloging time: `ReleaseDate` is already merged onto the book from the provider's work data (`aggregator_author_works.go`) before `fetchAuthorBooks`'s filter chain runs — this is a presence check on data already in hand, same as `SkipPartBooks`.
- `SkipMissingISBN` is not: ISBN data lives on `Editions`, not `Book.ISBN`, in this flow, and isn't fetched until `hydrateHardcoverEditions` runs **after** a book is already created. Wiring it the same way would mean either creating the book first (defeating the filter) or fetching edition data for every candidate work up front (a real API-cost change, not a copy of this pattern). That one needs a design decision first, which I've raised on #1980 rather than guessing at in code.

## Testing

- `TestFetchAuthorBooks_SkipsMissingDate` — full `fetchAuthorBooks` integration test (in-memory SQLite, stub metadata provider) verifying undated works are dropped while dated works come through, and asserting the `SkippedMissingDate`/`SkippedTotal` summary counters directly (not just book presence in the DB).
- `TestFetchAuthorBooks_MissingDateKeptWhenSkipDisabled` — confirms the filter is opt-in: with the setting at its default (`false`), behavior is unchanged and the counter stays at zero.
- `gofmt -l`, `go vet ./internal/api/... ./internal/models/...`, and `go test ./internal/api/... ./internal/models/...` all pass (ran inside the repo's own `golang:1.26.5-alpine` builder image).

## Test plan for reviewers

- [ ] Enable "Skip missing date" on a metadata profile in the UI, add/refresh an author known to have undated editions on OpenLibrary, confirm they're absent from Wanted and the sync summary reports a non-zero `skippedMissingDate` count
- [ ] Confirm a profile with the setting left off is unaffected (existing behavior)

Refs #1980.

### DIFF
--- a/changelog.d/skip-missing-date.md
+++ b/changelog.d/skip-missing-date.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Author sync now honors `skipMissingDate`** — the metadata profile setting previously had no effect during cataloging (only the field was persisted, nothing consumed it, per #1980). Works with no release date are now filtered out of author sync/refresh when the setting is enabled. `AuthorSyncSummary` gains a `skippedMissingDate` count alongside the existing skipped-language/junk/media-type counters so drops stay visible.
--- a/internal/api/authors.go
+++ b/internal/api/authors.go
@@ -1640,6 +1640,7 @@ func (h *AuthorHandler) fetchAuthorBooks(author *models.Author, opts catalogueSy
 	// Resolve the author's metadata profile (falling back to the seeded
 	// default) and parse its allowed_languages CSV. Nil means "no filter".
 	allowedLangs, unknownFail := h.resolveAllowedLanguages(ctx, author)
+	skipMissingDate := h.resolveSkipMissingDate(ctx, author)
 
 	// OpenLibrary works carry no work-level language; the search enricher only
 	// backfills it for indexed works, so a tail of works (often translations)
@@ -1824,7 +1825,7 @@ func (h *AuthorHandler) fetchAuthorBooks(author *models.Author, opts catalogueSy
 	// skippedExcluded is logged but deliberately kept out of AuthorSyncSummary:
 	// the author page's notice explains works the user did NOT expect to lose,
 	// and a book they excluded by hand is not one of them.
-	var added, skippedLang, skippedJunk, skippedMediaType, skippedNotAccepted, skippedExcluded int
+	var added, skippedLang, skippedJunk, skippedMediaType, skippedNotAccepted, skippedExcluded, skippedMissingDate int
 	// Names of the first few language-rejected works, reported to the user
 	// alongside the count (#1889): "65 books skipped" is alarming, but it is
 	// the titles and their language codes that tell them whether the profile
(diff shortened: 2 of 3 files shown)


gchahcg:
Rebased onto #2032 (the notice/sample plumbing PR) and added sample-population for this filter, per your review comment on #1968. Force-pushed — apologies for the disruption if you'd already started reviewing.

vavallee (commented):
Straightforward and correct in shape; the rebase onto #2032 is clean and the sample follows the established pattern.

Two things carried over from the reviews on the siblings rather than specific to this one:

**Already-tracked books stop refreshing.** Same as #1968 and #2006: this filter sits above the `GetByForeignID` branch at `authors.go:1894`, so a work already in the library with no release date never reaches the update path. It keeps its files and stays monitored, but stops receiving rating, genre, cover and series updates, and is reported as skipped on every subsequent sync. Worth fixing once in whichever of these lands first rather than five times.

**Missing dates are common in OpenLibrary**, more so than the other four filters' inputs. `b.ReleaseDate == nil` is a clean, unambiguous test — no unknown-vs-zero problem like #2006 has — but it will match a lot on an OL-only install. The notice from #2032 makes that visible, which is exactly why that PR had to go first, so I am not treating it as blocking. Just be aware the first user to turn this on with a large OL catalogue will see a big number.

Otherwise this looks good behind #2032 and whichever ordering fix lands.

vavallee:

Issue #807, opened 2026-05-24
Prefer stronger Hardcover rating signals during metadata enrichment


magrhino:
## Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I checked the [roadmap](https://github.com/vavallee/bindery/blob/main/docs/ROADMAP.md) and this isn't already planned

## Problem

Bindery's metadata merge behavior generally preserves the first non-zero rating values it has. That is conservative, but it can leave a book with sparse OpenLibrary rating data even when Hardcover has a stronger community signal for the same work.

The existing `average_rating` and `ratings_count` fields would be more useful for recommendations and filtering if enrichment could prefer a materially better-supported rating source.

## Proposed solution

Add a field-level merge rule for ratings that can replace existing rating values when incoming Hardcover metadata has a clearly stronger `ratings_count` signal.

One possible rule:

- keep the current value when the incoming rating is empty or has too few ratings
- replace `average_rating` and `ratings_count` when Hardcover has a substantially higher rating count than the current source
- apply the same rule consistently in book enrichment, author-work merge paths, ISBN hydration, and ABS/Hardcover metadata merge paths where applicable
- add focused tests so the rule does not overwrite stronger local/provider data with weaker Hardcover data

I can take this on if the maintainer wants it.

## Alternatives considered

Always preferring Hardcover ratings would be simple but too aggressive. Keeping the current first-non-zero behavior is safer, but it misses cases where Hardcover is clearly the better source.

## Additional context

This should only affect existing `average_rating` and `ratings_count` fields. It should not introduce a new rating-source model unless the maintainer wants source attribution later.

PR #2006, opened 2026-08-14
fix: honor minPopularity during author sync

gchahcg:
## Summary

`MinPopularity` on a metadata profile was persisted to the DB and returned by every profile read, but nothing in author sync consulted it (see #1980) — works were cataloged regardless of ratings count, with no way to filter low-interest noise despite the setting's name.

This wires a floor check into the existing `fetchAuthorBooks` filter chain (same place the junk-title, `SkipPartBooks`, and `SkipMissingDate` checks live), gated on the author's resolved metadata profile via a new `resolveMinPopularity` helper that mirrors `resolveSkipPartBooks`/`resolveSkipMissingDate`.

- **Added** — `AuthorHandler.resolveMinPopularity`, defaulting to `0` ("no filter", matching the settings UI which renders `0` as "none") on any lookup failure.
- **Changed** — `fetchAuthorBooks`'s ingestion loop drops a work when `minPopularity > 0 && b.RatingsCount < minPopularity`, **except** works that haven't released yet — they're exempt, since a forthcoming book can't have accumulated ratings and judging it by `RatingsCount` would only ever penalize new releases, never the low-interest-backlist noise the setting is meant for.
- **Changed** — `AuthorSyncSummary.SkippedMinPopularity` (+ folded into `SkippedTotal()` and the Warn-log gate).

### Design note — please sanity-check this one

Bindery doesn't have a dedicated "popularity" field the way Chaptarr does (`Book.Ratings.Popularity`, checked as reference before writing this). `RatingsCount` is the closest available proxy and is what this PR thresholds against.

Two things worth your review specifically:

1. **The `RatingsCount`/`AverageRating` pairing rule affects this filter.** `aggregator_author_works.go`'s merge (#807, "keep the pair together") only carries `RatingsCount` across when `AverageRating` is also nonzero. A work with a real ratings count but no average would read as `RatingsCount == 0` here — the same "0 means unknown, not unpopular" ambiguity `aggregator.go:464` already documents for search reranking, now also relevant to this filter. I didn't add a separate unknown-vs-zero carve-out for this — Chaptarr's own shipped `MinPopularity` filter has the identical accepted tradeoff (unlike its `MinPages` filter, which does special-case an all-zero reading as "unknown"). Happy to add one here if that's not an acceptable tradeoff for Bindery.
2. **`RatingsCount` may not be the field you had in mind for "popularity"** when the setting was originally added — if there's a different intended signal, easy to swap the threshold target.

## Testing

- `TestFetchAuthorBooks_SkipsBelowMinPopularity` — full `fetchAuthorBooks` integration test (in-memory SQLite, stub metadata provider) verifying a below-floor work is dropped while an above-floor work and an unreleased/unrated work (exemption) both come through, asserting `SkippedMinPopularity`/`SkippedTotal` directly.
- `TestFetchAuthorBooks_MinPopularityKeptWhenZero` — confirms the filter is opt-in: at the default (`0`), nothing is dropped.
- `gofmt -l`, `go vet ./internal/api/... ./internal/models/...`, and `go test ./internal/api/... ./internal/models/...` all pass (ran inside the repo's own `golang:1.26.5-alpine` builder image).

## Test plan for reviewers

- [ ] Set "Min popularity" on a metadata profile in the UI, add/refresh an author known to have low-interest editions on OpenLibrary, confirm they're absent from Wanted and the sync summary reports a non-zero `skippedMinPopularity` count
- [ ] Confirm a profile with the setting at 0 is unaffected (existing behavior)
- [ ] Sanity-check the design note above — particularly whether `RatingsCount` is the right signal, and whether the #807 pairing rule's effect on this filter is acceptable

Refs #1980.

### DIFF
--- a/changelog.d/min-popularity.md
+++ b/changelog.d/min-popularity.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Author sync now honors `minPopularity`** — the metadata profile setting previously had no effect during cataloging (only the field was persisted, nothing consumed it, per #1980). Works whose ratings count falls below the configured floor are now filtered out of author sync/refresh, with an exemption for works that haven't released yet (they can't have accumulated ratings). `AuthorSyncSummary` gains a `skippedMinPopularity` count alongside the existing skipped-language/junk/media-type counters so drops stay visible.
--- a/internal/api/authors.go
+++ b/internal/api/authors.go
@@ -1641,6 +1641,7 @@ func (h *AuthorHandler) fetchAuthorBooks(author *models.Author, opts catalogueSy
 	// default) and parse its allowed_languages CSV. Nil means "no filter".
 	allowedLangs, unknownFail := h.resolveAllowedLanguages(ctx, author)
 	skipMissingDate := h.resolveSkipMissingDate(ctx, author)
+	minPopularity := h.resolveMinPopularity(ctx, author)
 
 	// OpenLibrary works carry no work-level language; the search enricher only
 	// backfills it for indexed works, so a tail of works (often translations)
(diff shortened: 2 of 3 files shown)


gchahcg:
Rebased onto #2032 and added sample-population, same as #2005. Also fixed a De Morgan's-law lint finding on the popularity condition, surfaced during the rebase (not introduced by it). Force-pushed.

vavallee (commented):
The rebase and the sample plumbing are right, and the future-release exemption is a good catch. One blocking problem, and it is the mirror image of something you got right in #2007.

## Unknown ratings are treated as zero, so this can drop an entire catalogue

```go
if minPopularity > 0 && b.RatingsCount < minPopularity && (b.ReleaseDate == nil || !b.ReleaseDate.After(time.Now())) {
```

`RatingsCount` is not something OpenLibrary reliably supplies. It arrives via the Hardcover supplement merged in `mergeAuthorWorks` (`aggregator_author_works.go:349`), and the comment on the merge rule describes OpenLibrary's own data as sparse (`aggregator_enrichment.go:296`: "Hardcover's community signal vs. sparse OpenLibrary data").

So on an install with no Hardcover token configured, or for any work Hardcover does not know, `RatingsCount` is `0` — not "this book has zero ratings", but "nobody told us". Setting MinPopularity to anything above 0 on such an install drops **every work of every author**, and the author page reports them all as skipped for unpopularity.

Compare what #2007 does for the analogous case, from your own description:

> A work with no page data at all is treated as unknown, not zero, and passes the page-count floor either way.

That is the correct rule and it should apply here too: a work with no ratings data at all is unknown, not unpopular. `RatingsCount == 0` with `AverageRating == 0` is indistinguishable from missing, so the conservative reading is to pass anything with no rating signal and only filter works that actually carry one.

If you would rather filter aggressively, the setting needs to say so in the UI, and the notice needs to distinguish "below your floor" from "no popularity data available" — otherwise the first support thread is "MinPopularity=1 deleted my whole library", and it will be right.

## Also applies here

The already-tracked finding from my review on #1968 applies unchanged: this filter sits above the `GetByForeignID` branch at `authors.go:1894`, so a book the user already owns that falls below the floor stops receiving rating, genre and cover updates and is reported as skipped on every sync. For this filter specifically that is a loop: the book is skipped because its ratings count is low, and skipping it is what prevents its ratings count from ever being updated.

vavallee:
Rebased this onto main myself since #2005 landed first and the two collide on the same counter block. Squashed your three commits into one (authorship preserved) and merged the conflicting regions by hand: both filters now sit under the single hoisted `existing` lookup, the summary struct and log line carry both sets of fields, and the Warn gate sums both.

The hasRatingSignal distinction is the right call. Treating a missing rating as unknown rather than zero is exactly what stops a no-Hardcover install from silently filtering half a catalogue.

Full api and models suites pass locally. Merging once CI is green.

Issue #1929, opened 2026-08-11
Author sync does per-created-book work serially: a full library walk and a Hardcover call each


vavallee:
Split out of #1888. That issue fixed the *pre-loop* enrichment phases; this is the half inside the sync loop, which was measured but deliberately left alone because both fixes are larger than a patch release warranted.

Whether this is what a user actually feels is gated on one unanswered question in #1888 (did the `unknown = fail` run also take an hour), but the serial work exists either way and is worth tracking on its own.

### Measured

Instrumenting `RefreshAuthorBooks` for a 65-work author gave **260 upstream round trips at peak concurrency 1**. Two of the four per-work costs sit inside `for _, b := range books`, and both scale with *books created* rather than works fetched:

| per created book | where |
|---|---|
| one Hardcover `GetEditions` | edition hydration |
| one full recursive library walk | `LibraryFinder.FindExisting` |

### 1. `Scanner.FindExisting` walks the whole library per book

`internal/importer/scanner.go` does a recursive `filepath.Walk` of the library root for **every newly created book**, and:

- it **never short-circuits** — it sets `found` and keeps walking the entire tree
- it **ignores its `ctx`**, so a cancelled sync keeps walking

On local disk with a warm dentry cache this is cheap enough to disappear. On an NFS or SMB mount with a large library it is potentially seconds each, times the number of books created. That profile matches the reporter's setup better than it matches a developer laptop, which is why it is a live suspect rather than a theoretical one.

The fix is to walk once per sync and reuse the index, rather than once per book. That changes the `LibraryFinder` interface, which is why it was not attempted inline.

### 2. Hardcover edition hydration is one call per book

`hydrateHardcoverEditions` issues a `GetEditions` per created book, serially. Hardcover's GraphQL API accepts `where: {book_id: {_in: [...]}}`, so this is batchable — the same shape as the #1694 fix. It is a real change to the Hardcover client plus the hydrate path, so it needs its own PR and its own tests.

### Sequencing

Neither is worth starting until #1888's question comes back, because the answer says whether this path or the pre-loop path dominates in practice. If it turns out the per-book work dominates, the library walk is the one to do first — it is the only item here with an unbounded cost that depends on the user's storage rather than on provider latency.

### Related

- #1888 — the original report and the pre-loop fix that shipped in v1.30.2
- #1694 — the batching pattern the Hardcover fix should follow
- #1854 — per-book network calls under a request deadline, the adjacent list-sync case

[...]

Issue #1888, opened 2026-08-11
Author refresh takes ~1 hour to add 65 books


vavallee:
Reported on Discord by EricSilberberg (2026-08-11, **v1.30.1**) while confirming #1872:

> Tested and confirmed with Terry Bisson. with fail, and 65 books added with pass. **Not sure why it took almost an hour for 65 books though.**

An author refresh that adds 65 books should not take an hour. That is roughly 55 seconds per book, which is far more than a metadata fetch should cost even with enrichment.

### Not yet root-caused

Filing on the reporter's observation, not a local reproduction. Starting points, in rough order of suspicion:

- **Per-book provider calls in the sync loop.** `syncAuthorBooks` (`internal/api/authors.go`, around the `added`/`skippedLang` accounting near line 1587-1818) does per-book work inside the loop. #1694 found exactly this shape on the Hardcover list-sync path: one fully paginated edition query per imported book. Worth checking whether the author path has an equivalent fan-out.
- **Audnex enrichment.** #1854 records that re-adding a per-book Audnex call took the list-sync manual path from 1,485 books in 6s to 519 in 60s. If the author refresh enriches per book, the same arithmetic applies.
- **Edition hydration.** `hydrateMatchedHardcoverEditions` is called inside the dedup branch of the same loop.
- **Sequential provider round-trips with no concurrency**, where the bulk search paths use a bound of 4 (`authorAutoSearchConcurrency`, `seriesFillSearchConcurrency`).

### What would settle it

A timing breakdown per phase on a real refresh of a large author. If it turns out to be per-book network calls under a sequential loop, this is the same structural problem as #1854 and could share a fix.

### Impact

The author refresh is the most-used metadata operation in the app. At this rate a large author is effectively unusable, and there is no progress feedback while it runs (related: the progress-feedback half of #1636).

vavallee:
Add **Steps to reproduce** — the minimal numbered sequence from a known starting state (1. Go to … 2. Click … 3. See error …). A log section shows what failed; the steps are what make it reproducible outside your deployment.

— 🤖 bindhelper (automated). Reply to correct me; a human reviews every change.

vavallee:
Reopening. This was closed automatically when #1917 merged, which was my mistake: the PR body opened with "partly fixes #1888" and GitHub parsed the closing keyword and ignored the qualifier. Only part of this is addressed and the reported symptom is not confirmed fixed.

### What shipped in v1.30.2

[...]

PR #2007, opened 2026-08-14
fix: honor minPages and skipMissingIsbn during author sync

gchahcg:
## Summary

`MinPages` and `SkipMissingISBN` on a metadata profile were persisted to the DB and returned by every profile read, but nothing in author sync consulted either one (see #1980) — works were cataloged regardless of page count or ISBN presence, with no way to filter on either despite the settings' names.

Both need edition data (page count and ISBN live on `Edition`, not `Book`) that `fetchAuthorBooks`'s create-then-hydrate order doesn't have at filter time — `hydrateHardcoverEditions` only runs **after** a book row already exists (`internal/api/authors.go`, right before `added++`). That was the specific blocker I flagged on #1980 for these two fields.

**Update:** the first version of this PR fetched editions inline, one per candidate, inside the ingestion loop. On self-review I caught that this reintroduces exactly the pattern #1929 already tracks as a known author-sync perf problem (serial per-book provider calls) — so I pushed a follow-up commit that bounds the concurrency instead. Below describes the current state.

**Correction from review:** "batched" overstated it in an earlier version of this description — it's a bounded fan-out of one `GetEditions` call per surviving candidate (`authorAutoSearchConcurrency` at a time), not a batch request that reduces call count. For an author with 400 surviving works that's still 400 provider round trips per sync, just four concurrent instead of 400 serial. Reworded below so the cost isn't understated for whoever debugs a slow sync later.

On caching: `Aggregator.GetEditions` (`internal/metadata/aggregator.go:755-771`) does sit behind a 24h TTL cache (`newTTLCache(24 * time.Hour)`), so a work whose editions were already fetched — by this filter, by cover/genre hydration, or by a previous sync — within the last 24h is served from cache, not re-fetched. That bounds the *added* cost of turning these settings on to roughly one real round-trip per surviving candidate per 24h window, not one per scheduled refresh if refreshes run more often than that. Worth keeping in mind against #1888/#1929 regardless, since the first sync (or the first one after a 24h-stale cache) for a large library still pays the full per-candidate cost this PR's bounded fan-out is spread over.

`fetchAuthorBooks`'s loop is now split in two: a first pass runs the free (in-memory) filters only (media-type, junk-title, language) and collects surviving candidates; editions for those candidates are prefetched with bounded concurrency (`concurrency.RunBounded`, same cap as the auto-search fan-out a few lines below) before a second pass applies `MinPages`/`SkipMissingISBN` against the prefetched results and does everything the rest of the original loop did, unchanged. A work rejected by a free filter still never costs a round-trip — only *when* and *how* surviving candidates' editions get fetched changed. I checked Chaptarr's equivalent (`EditionMetadataProfileFilter.cs` / `MetadataProfileService.FilterBooks`) as a reference before writing the original version: it solves the underlying problem the same way (fetch and attach editions before filtering, not after creating).

- **Added** — `resolveEditionFilters` (combined lookup for both settings, since both gate the same prefetch), `anyEditionHasISBN`, `passesMinPagesFilter`.
- **Changed** — `fetchAuthorBooks` collects candidates surviving the free filters, prefetches their editions with bounded concurrency when `needsEditionPreview` is true, then applies both checks against the prefetched map in a second pass.
- **Changed** — `AuthorSyncSummary.SkippedMinPages` / `SkippedMissingISBN` (+ folded into `SkippedTotal()` and the Warn-log gate).

### Missing-data handling

`MinPages` mirrors Chaptarr's semantics exactly: a work whose editions all report zero/no page count is treated as **unknown**, not zero, and passes the floor rather than being penalized for a provider data gap.

`SkipMissingISBN` checks ISBN-13/ISBN-10 across all of a work's editions; a work with no editions at all also counts as missing. This treats "no page data" and "no editions at all" differently between the two filters, deliberately: a size floor shouldn't penalize a work the provider hasn't indexed a page count for, but for an identifier-presence gate, "we found zero editions to check" is the strongest available signal that no ISBN can be confirmed — there's no partial case the way there is for page count. I checked this against Chaptarr's identifier filter rather than assuming it's fine: it runs per-edition and gates the book afterward on `Editions.Any()`, so a book with zero editions already fails there the same way. Documented inline now so this doesn't read as an inconsistency on a future pass.

One thing worth a second look from you regardless: `SkipMissingISBN` here only checks the literal ISBN fields, not `ASIN`. Chaptarr splits this into two independent toggles (`SkipMissingIsbn`/`SkipMissingAsin`) that OR together when only one is set. I kept this PR to the literal field name rather than silently broadening what it means — happy to add a `SkipMissingASIN` counterpart as a fast follow if useful, since `Book.ASIN` is already populated earlier/more reliably than edition ISBN in several of Bindery's paths.

### Already-tracked books

Per review (same finding on #1968/#2005/#2006): the `GetByForeignID` lookup that updates already-owned books ran after both new filters' `continue`, so a book the user already owned but that failed `MinPages`/`SkipMissingISBN` kept its files but silently stopped receiving rating/genre/cover updates, reported as skipped every sync. Hoisted the lookup to the top of the candidates loop and exempted already-tracked books from both filters.

### Edition-lookup failure handling

If a candidate's prefetch errors (network hiccup, provider outage), it's absent from the results map — the loop treats that the same as "not enforcing for this work" an inline call would have, so a transient failure never drops a book. This matches every other resolver in this file (`resolveSkipPartBooks`, `resolveSkipMissingDate`, etc.), which all fail open rather than risk unexpected catalogue loss on a transient lookup failure.

## Testing

- `TestFetchAuthorBooks_SkipsBelowMinPages` / `TestFetchAuthorBooks_MinPagesKeptWhenZero` — cover the floor, the missing-page-data escape hatch, and the default-off path.
- `TestFetchAuthorBooks_SkipsMissingISBN` / `TestFetchAuthorBooks_MissingISBNKeptWhenDisabled` — cover a work with a real ISBN, a work with edition data but no ISBN, a work with no editions at all, and the default-off path.
- `TestFetchAuthorBooks_EditionLookupSkipsFreeFilterRejects` (new) — confirms the prefetch only fires for candidates that survive the free filters (junk-title, wrong-language), not for every work in the list, since the batching restructure is exactly the kind of change that could silently lose that property.
- Both `*KeptWhenZero`/`*KeptWhenDisabled` tests additionally assert `provider.editionCalls` is empty when neither setting is on.
- Found and fixed a real data race while verifying this under `go test -race`: the test stub's `editionCalls` recorder was an unsynchronized slice append, fine when every call was serial, broken once the prefetch called it concurrently (same bug shape #1924 already fixed for a different mock). Fixed with a mutex on the stub; not a production-code issue, but worth naming since it's exactly the kind of thing `-race` exists to catch and a plain `go test` run doesn't.
- `gofmt -l`, `go vet ./internal/api/... ./internal/models/...`, `go build ./...`, and `go test -race -run "MinPages|MissingISBN|EditionLookup|StrictMediaType" ./internal/api/...` all pass (native Go 1.26.6, not the Docker builder image this time). A full unsharded `go test -race ./internal/api/...` times out on this machine on unrelated pre-existing tests unconnected to this change (a parallel-test queue wait, a SQLite migration exec) — consistent with why CI shards its own race suite into six jobs rather than running it as one; treated as an environmental limit of a single local run, not a signal about this diff, but flagging so it's not a surprise if you try the same thing.

## Test plan for reviewers

- [ ] Set "Min pages" and/or enable "Skip missing ISBN" on a metadata profile in the UI, add/refresh an author, confirm the sync summary reports non-zero `skippedMinPages`/`skippedMissingIsbn` where expected
- [ ] Confirm a profile with both settings left off is unaffected, including no extra API calls (watch logs/rate-limit counters during a refresh)
- [ ] For an author with many candidates and one of the settings on, confirm the edition lookups now overlap (bounded concurrency) rather than visibly serializing in the logs

Refs #1980.


gchahcg:
Rebased onto #2032 and added sample-population for both MinPages and SkipMissingISBN independently. Re-verified with -race given this PR's own bounded-concurrency edition prefetch. Force-pushed.

vavallee (commented):
The most careful of the five. The mutex-guarded map, the bounded fan-out, fail-open on lookup error, and the deliberate split between "no page data" (unknown, passes) and "no ISBN" (a real answer, filters) are all right, and the comment explaining why the two filters read an empty edition list differently is the kind of thing that stops someone "fixing" the asymmetry later.

Three notes.

**"Batched" overstates it.** `concurrency.RunBounded(ctx, candidates, authorAutoSearchConcurrency, ...)` is a bounded fan-out of one `GetEditions` call per surviving candidate, not a batch request. For an author with 400 surviving works that is 400 provider round trips per sync, four at a time. The gating is good — nothing happens unless one of the two settings is on — but the PR description reads as though the call count is reduced when what is bounded is the concurrency. Worth rewording so the cost is not understated for whoever debugs a slow sync later, and worth checking against #1888 (author refresh already taking ~1 hour for 65 books) and #1929 before this goes near a large library.

Is there any caching in front of `GetEditions`? If not, the same works are re-fetched on every scheduled refresh, and that is the number that matters, not the one-off.

**Already-tracked books stop refreshing.** Same finding as #1968, #2005 and #2006: the filter sits above the `GetByForeignID` branch at `authors.go:1894`. Here it also means the edition lookup runs for books the user already owns, so the cost above is paid on works that were never going to be added either way.

**`skipMissingISBN` is checked before `minPages`.** Fine, but it means a work that fails both is only ever reported under ISBN, so the counts are "first reason wins" rather than independent. That matches how the language and junk filters already behave, so I am noting it rather than asking for a change — just be aware the notice's numbers are not additive across filters.

Good behind #2032 and whichever ordering fix lands first.

vavallee:

vavallee:
Rebased onto main after #2005 and #2006 landed. Squashed to one commit, authorship preserved.

The conflict here was structural rather than textual: your two-pass split with the batched edition prefetch is a different loop shape from the single pass the other two filters were written against. I kept your structure and moved the missingDate and minPopularity checks into the second loop, under the same hoisted `existing` lookup. Net effect is that a work failing both a language check and a date check now counts against skippedLanguage rather than skippedMissingDate. No test depended on the old attribution and language is the more fundamental filter, so that reads as the better answer anyway.

Batching the edition lookups instead of firing one per candidate is the right instinct, and gating it on whether either setting is actually on means profiles that use neither pay nothing.

Full api suite passes locally. Merging on green.

Issue #1780, opened 2026-08-05
Box sets and omnibuses are only pruned when Hardcover enrichment is available, so OpenLibrary-only catalogues fill with bundle rows


vavallee:
## Pre-flight

- [x] Searched existing issues — #361 added box-set detection to the *recommender* only; #1348 covers mass-monitoring, not bundle rows.
- [x] Reproduced against `main` at 8b30fe88 (v1.29.1).

## Current behaviour

OpenLibrary models box sets, omnibuses and collector's editions as separate Works, so they arrive in an author's catalogue as ordinary books. Fetching `https://openlibrary.org/authors/OL26320A/works.json?limit=100` (J. R. R. Tolkien, `"size": 342`) returns 10 such entries in the first 100:

```
The Complete History of Middle-Earth
The Complete History of Middle-earth, Vol. 2 (The History of Middle-ea…)
The Lord of the Rings 3 Books Box Set By J. R. R. Tolkien
Tolkien Myths and Legends Box Set
The Lord of the Rings Trilogy, 3 Volumes boxed Set
Hobbit and the Lord of the Rings Illustrated by Alan Lee Box Set
Lord of the Rings Deluxe Illustrated Box Set
Lord of the Rings Trilogy
Lord of the Rings Movie Trilogy Colouring Book
Lord of the Rings Collector's Edition Box Set
```

Nothing on the OpenLibrary ingestion path removes them:

- `shouldFilterOLNoise` (`internal/metadata/openlibrary/client.go:570`) filters study guides, summaries, and screen adaptations. `olNoiseTitleFragments` (`client.go:550`) and `olNoiseSubjects` (`client.go:534`) contain no fragment for box set, boxed set, omnibus, or collector's edition.
- `pruneAuthorWorkCompilations` (`internal/metadata/aggregator_author_works.go:172`) does prune bundles, but only via titles an *enricher* flagged. Its own doc comment states the constraint: "Primary-provider works (OpenLibrary) carry no compilation signal of their own, so the enricher's classification is the only way to know an inflated 'bundle' came through the primary list."
- That enricher is Hardcover, which is silently skipped when no token is configured. Hardcover is not required for a working install, so the default OpenLibrary-only setup gets no pruning at all.
- The pruning that does run is title-keyed (`authorWorkMergeKey`), so even with Hardcover configured it only fires when the Hardcover compilation title normalizes to the same key as the OpenLibrary one.

[...]

Issue #1723, opened 2026-08-03
Five of six metadata-profile filters are dead: MinPages, MinPopularity, SkipMissingDate, SkipMissingISBN, SkipPartBooks never consulted


vavallee:
### Problem

Reported on Discord by vegapunk, who spotted `SkipMissingISBN` doing nothing and correctly guessed it was the same class of bug as the quality-profile allowed-formats list (#1693). It is, and it is not one field. **Five of the six metadata-profile filters are dead.**

The settings UI presents them as working controls. `MetadataTab.tsx:285-289` binds editable inputs for all of them, and lines 248-255 render them back as badges on the profile card, so a user sets "minimum 100 pages, skip books with no ISBN", sees it saved, sees it displayed, and nothing ever applies it.

### Current state

Every read site of each field, excluding its own model definition and persistence code:

| Field | Read sites | Honored? |
| --- | --- | --- |
| `SkipMissingISBN` | 0 | no |
| `SkipMissingDate` | 0 | no |
| `SkipPartBooks` | 0 | no |
| `MinPopularity` | 0 | no |
| `MinPages` | 0 | no |
| `AllowedLanguages` | 5 | **yes** |

`AllowedLanguages` is the only one wired through, via `internal/api/authors.go:2528`, `internal/scheduler/scheduler.go:489`, and `internal/api/indexers.go:515`.

The other five are fully round-tripped and completely inert: declared in `internal/models/metadata_profile.go`, columns in `003_features.sql`, selected, scanned, inserted and updated in `internal/db/metadata_profiles.go`, covered by `metadata_profiles_test.go`, exposed over the API, and rendered in the UI. The persistence tests pass because they only assert the values survive a round trip, which they do. Nothing asserts they change behaviour.

That is exactly the #1693 shape: a complete-looking feature where the only missing piece is the one call that would consult it.

### Proposed solution

These belong on the author-ingestion path, next to where `AllowedLanguages` already filters, so the same profile that drops a wrong-language work also drops a 12-page "part 3 of 7" entry with no publication date.

Suggest landing them one at a time rather than as a batch, since each needs a decision about what a missing value means:

* `MinPages` / `MinPopularity` — most providers return 0 or absent rather than a real count. Does absent fail the filter or pass it? The metadata profile already has `UnknownLanguageBehavior` (pass/fail) for exactly this question on languages, so mirroring that pattern is probably right.
[...]

PR #2008, opened 2026-08-14
fix: prune collection-titled works without Hardcover configured

gchahcg:
## Summary

`pruneAuthorWorkCompilations` only caught box sets/omnibuses via an enricher's (Hardcover's) compilation classification — its own doc comment said as much: *"Primary-provider works (OpenLibrary) carry no compilation signal of their own, so the enricher's classification is the only way to know an inflated 'bundle' came through the primary list."* Hardcover is optional, so a default OpenLibrary-only install got zero pruning: box sets, omnibuses, and collector's editions landed in the catalogue as ordinary books, monitored and Wanted under the default monitor mode (see #1780).

Bindery already had a provider-independent detector for this shape — `internal/recommender/candidates.go`'s `looksLikeCollection`, used to keep bundles out of recommendations — but it was never consulted during author-catalogue ingestion, exactly as #1780 pointed out.

- **Added** — `textutil.LooksLikeCollectionTitle`, extracted from the recommender's `looksLikeCollection` so both consumers share one keyword list instead of two that can drift apart. The recommender now delegates to it.
- **Changed** — `pruneAuthorWorkCompilations` also drops a work when its title matches `LooksLikeCollectionTitle`, independent of any enricher.

This runs unconditionally in `GetAuthorWorksForAuthor` — same enforcement level the existing Hardcover-based check already had (no profile opt-in), matching what the issue explicitly asked for: *"regardless of which metadata providers are configured."*

## Two things worth a second look

1. **The title heuristic is intentionally broad**, and already accepts some false positives in its recommender use (`"Complete Guide to Go Programming"` is a documented example in its own test). Extending it to catalog ingestion raises the stakes of a false positive from "not recommended" to "never auto-cataloged." I deliberately did **not** widen the keyword list to also catch two of the ten real-world examples in #1780 (bare `"... Trilogy"` titles, a colouring book) — `"trilogy"` alone is a much higher false-positive-risk keyword than `"omnibus"`/`"box set"` (plenty of real single books are subtitled "Book One of the X Trilogy"), so that felt like a separate call for you to make rather than one to bake in silently. Coverage is meaningfully better than today (zero), not complete.
2. **A false-positive-pruned title isn't reachable via single-foreign-ID add either.** I checked this rather than assuming: `fetchAuthorBooks`'s `onlyForeignID` path trims the list `GetAuthorWorksForAuthor` already produced (`internal/api/authors.go`, `keepWorkWithForeignID` runs *after* the works list is fetched and pruned), so it never sees an entry this filter already dropped. I couldn't confirm the "reachable via manual add" escape hatch the issue's "Expected behaviour" section described actually applies to that specific path — a real single book caught by the heuristic would need a different import path (manual/ISBN/file import) instead. Documented directly on `pruneAuthorWorkCompilations` so this isn't a surprise later; happy to be told I'm wrong about a path I missed.

## Testing

- `TestLooksLikeCollectionTitle` — the shared helper directly, including #1780's real-world Tolkien examples.
- `TestAggregator_GetAuthorWorksForAuthor_PrunesCollectionTitlesWithoutEnricher` — reproduces #1780's exact scenario: no enrichers configured, real box-set/omnibus titles from the issue get pruned, real single-book titles still come through.
- Full `go test` across `internal/textutil`, `internal/recommender`, `internal/metadata`, `internal/api`, and `internal/models` — all pass, no regressions in the existing Hardcover-based compilation test or anywhere else.
- `gofmt`/`go vet` clean.

## Test plan for reviewers

- [ ] `make check`
- [ ] With no Hardcover token configured, add J. R. R. Tolkien (or another author with heavily boxed-set-ed works) and confirm box-set/omnibus rows are absent from the catalogue
- [ ] Confirm a real single-book title isn't accidentally caught (spot-check a few author catalogues after this change)

Refs #1780.


gchahcg:
Rebased onto the updated #1968 (not #2032 directly — this touches the same file #1968 already extended, so it stacks there instead). Found and fixed one real interaction between the two: the collection-title prune here now catches one of #1968's test fixture titles before SkipPartBooks sees it, so I adjusted the expected counts/sample to match actual combined behavior rather than just resolving the merge conflict blindly. Force-pushed.

vavallee (commented):
This one I would not merge in its current form. The gap it identifies is real — `pruneAuthorWorkCompilations` does nothing on an OpenLibrary-only install, and box sets do land in the catalogue monitored and Wanted — but the fix reuses a heuristic whose tolerances were set for a much lower-stakes surface, and applies it unconditionally.

## The keyword list is far too broad for cataloguing

```go
var collectionTitleKeywords = []string{
	"complete", "collected", "omnibus", "boxed set", "box set", "anthology",
	"the best of", "stories of", "tales of", "(omnibus)", "(collection)",
	"complete works", "complete collection",
}
```

Bare `strings.Contains`, no word boundaries, no positional rules. Real single books this drops:

- **"Stories of Your Life and Others"** — Ted Chiang. Matches `stories of`.
- **"Tales of the City"** — Armistead Maupin, a novel. Matches `tales of`.
- **"The Complete Guide to …"**, anything with "complete" anywhere. Your own doc comment names this one.
- **"Collected"** appears in plenty of single-volume titles that are not compilations.

Your comment is explicit that this is accepted:

> The check is intentionally broad: some false positives (e.g. "Complete Guide to Go Programming") are accepted to keep the logic simple, matching the tolerance already established for the recommender's use of this list.

That tolerance was established for `recommender.go:185`, where a false positive means **one suggestion is not shown**. Nothing is lost, the user never knows, and the book is still in their library. Here the same match means the work is **dropped from the author's catalogue during ingestion** — it is never created, never monitored, never searched for. Those are not comparable stakes, and inheriting the tolerance along with the list is the problem.

The contrast with #1968 is stark, by the same author in the same stack: there you required spaced slashes to avoid "He/She/It", excluded leading "omnibus" to save "The Omnibus of Crime", and pinned two known residuals in a test table. All of that care is bypassed by `strings.Contains(lower, "complete")`.

## It is also unconditional

`isPartBookTitle` in #1968 runs only when the profile's `SkipPartBooks` is enabled. This prune runs for everyone with no Hardcover token, with no setting to turn it off and — since it happens in the aggregator, before `fetchAuthorBooks` — **no skip count and no entry in the notice** that #2032 just built. It is exactly the invisible-drop shape that made me ask for #2032 before any of these landed.

## What I would do instead

1. **Reuse `isPartBookTitle` from #1968** rather than the recommender's list. It is the one that has been tested against real false positives, and having two heuristics for one concept at two layers is how they drift.
2. **Gate it on the same `SkipPartBooks` setting**, so this is the OpenLibrary-only implementation of a documented, toggleable behaviour rather than a second silent one.
3. If it must stay in the aggregator where there is no summary to report into, **count it and surface it** — an invisible prune is the thing #1723 was filed about.

Leave the recommender's list alone. Broad is fine there.

## Process note

The base is `main` while the branch stacks on #1968, so this diff shows both changes and lists `changelog.d/skip-part-books.md` as its own. Retarget the base at the #1968 head branch so the diff shows only what this PR adds — otherwise merging this first silently brings #1968 with it, and the review above is the only thing standing between that and `main`.

vavallee:
Holding this one, and I think you'll agree once you see where it landed.

Your last commit (97118274) removed the aggregator title-shape prune, which was the actual fix here. That was the right call and matches what I'd asked for. But with it gone I rebased the remainder onto main and the behavioural diff is empty. What's left is moving looksLikeCollection into textutil as an exported helper, and that helper now has exactly one caller, the recommender, same as before. The aggregator only mentions it in a comment explaining why it deliberately doesn't use it.

The user facing story is already told by #1968's changelog fragment, which shipped a few minutes ago. This PR's fragment describes the same fix plus the decision not to do the broader thing, so the two would read as duplicates in the release notes.

So the question is what you want this PR to be. Three options as I see it:

1. Close it as absorbed into #1968. Nothing is lost, the fix shipped.
2. Keep just the textutil extraction, drop the changelog fragment, and retitle it as a refactor. Honest, but an exported API with one caller is a bit early.
3. Give the helper a second caller by having isPartBookTitle consult it, which is what would actually justify the shared package.

I lean towards 1. The catalogue side is fixed and #1723 is closed out. Your call though, and thanks for the four that landed, they were good work and the review feedback turnaround was fast.

gchahcg:
Agreed, closing as absorbed into #1968. The catalogue-side fix (relying on SkipPartBooks' existing detector instead of a second aggregator-level heuristic) already shipped there, so there's no remaining behavioral diff here and nothing lost. Thanks for walking through the options.

Issue #1982, opened 2026-08-14
Path components are capped by rune count, not bytes: non-ASCII titles fail to import with ENAMETOOLONG


vavallee:
### Current behaviour

`sanitizePath` (`internal/importer/renamer.go`) caps each path component at `maxPathComponentLen = 200`, counted in **runes**:

```go
if r := []rune(p); len(r) > maxPathComponentLen {
    p = strings.TrimSpace(string(r[:maxPathComponentLen]))
}
```

Filesystems cap filenames in **bytes**, not runes — 255 on ext4, XFS, btrfs and APFS. For any non-ASCII title the cap therefore permits a component the filesystem rejects, which is the exact failure the cap exists to prevent.

The import fails with `ENAMETOOLONG` when the renamer tries to create the destination.

### Expected behaviour

A sanitised component fits on disk regardless of script. The cap should be a byte budget, truncated on a rune boundary so it never splits a multi-byte character.

### Steps to reproduce

Import a book whose title is long and non-ASCII — Japanese, Chinese, Korean, Russian, Greek, or heavily accented Latin. Roughly 83 CJK characters or 125 Cyrillic characters is enough, before the extension.

Or directly, on ext4:

```
$ python3 -c "
import os,tempfile
d=tempfile.mkdtemp()
t='転'*84
open(os.path.join(d,t+'.epub'),'w')"
OSError: [Errno 36] File name too long
```

### Measurements

```
200 CJK runes  = 600 bytes  → errno 36
 86 CJK runes  = 258 bytes  → errno 36
 85 CJK runes  = 255 bytes  → errno 36 (+".epub" = 260)
 84 CJK runes  = 252 bytes  → errno 36 (+".epub" = 257)
```

Thresholds by script, including a 5-byte extension:

| Script | bytes/rune | fails at about |
|---|---|---|
| CJK, emoji | 3–4 | 83 characters |
| Cyrillic, Greek, accented Latin | 2 | 125 characters |
| ASCII | 1 | never — the 200-rune cap already fits |

### Note on the existing tests

`TestSanitizePath_CapIsRuneSafe` and `TestSanitizePath_CapsComponentLength` in `internal/importer/renamer_sanitize_test.go` both assert `RuneCountInString(got) == maxPathComponentLen`. They pass today and encode the defect, so a correct fix has to update them rather than work around them. The rune-safety half of that test is still right and should survive: truncation must not split a character mid-encoding.

### Bindery version

[...]

PR #2017, opened 2026-08-14
fix(importer): cap the composed path segment, not just each value (#2014)

### LINKED ISSUE #2014: Composed path segments are uncapped: the default template overflows NAME_MAX with plain ASCII titles
## Problem

`sanitizePath` caps each **template value**, but nothing caps the **composed segment** the template produces. With the default naming template the result already exceeds `NAME_MAX` for **pure ASCII** input — no non-ASCII involved.

Default template `{Title} - {Author}.{ext}`, with a title and an author each at the 200-byte value cap:

```
200 (title) + 3 (" - ") + 200 (author) + 5 (".epub") = 408 bytes
→ CREATE FAILED: file name too long
```

Measured on the #2013 branch. #1982 fixed the *value* cap (rune → byte); this is the layer above it, and #1982 does not address it.

## Contributing factors

- **`stagingPath` prepends ~34 bytes**, so a destination that is legal at 230 bytes can still fail during *staging* rather than at the final write — a confusing place to see the error.
- **`UniqueDir` appends `" (N)"`** with no reservation for it, so a name that just fits fails once a collision suffix is added.
- **`sanitizeInline` does not cap at all.**
- **`{ext}` and `{Part}` are substituted raw**, outside the value cap.

## Why the naive fix is wrong

Capping `renderSegment` by simple truncation chops the extension off the end, which is worse than the error it replaces — a file with no extension is invisible to every format check downstream. A correct fix reserves budget for the extension and any collision suffix before truncating the variable part, and knows which trailing bytes are structural.

## Scope note

Reachable today with an ASCII library and the shipped default template; it needs long titles *and* long author names, which is why it has not been common. Non-ASCII makes it easier to hit, but is not required.

## Not covered by this issue

`web/src/pages/settings/namingTemplate.ts` — the TypeScript preview mirror never implemented the length cap or the control-character strip, so the preview disagrees with what the importer will actually produce. Pre-existing and unchanged by #2013. Neither Go drift guard exercises a long value, so both give false confidence. A correct mirror is non-trivial: `String.slice` counts UTF-16 code units, so a naive `slice(0, 200)` is neither a rune cap nor a byte cap. Worth its own issue.

sujeito-operator:
## Summary

Closes #2014.

`sanitizePath` caps each template **value**; nothing capped the **segment** the template composes out of them. Reproduced on a clean checkout of `ee26e11` (the #2013 merge), pure ASCII, shipped default template, both values legal under the 200-byte value cap:

```
--- FAIL: TestReproBeforeFix (0.00s)
    leaf is 408 bytes
    REPRODUCED: os.Create(408-byte leaf): open /tmp/.../xxx…/yyy… ()/yyy… - xxx….epub:
    file name too long
```

`renderSegment` now composes through a new `composeCapped`, which shrinks **values** — never the template's own literals — longest value first, by bisection on a per-value byte budget.

### Why it shrinks values and not the composed string

You already flagged the naive fix in the issue, and this is the same reasoning:

- **The literals are structural.** In `{Title} - {Author}.{ext}` the `.` before the extension and the `" - "` between the names are literals. Right-truncating the finished string takes the extension with it, and a file with no extension is invisible to every format check downstream — worse than the error it replaced.
- **Longest-first is what protects the raw tokens.** `{ext}`, `{Year}` and `{Part}` are substituted outside `sanitizePath`'s value cap, and they are the ones whose loss is unrecoverable. Being short is enough to survive; nothing special-cases them by name, so a token added later gets the same treatment for free.
- **A group referencing no known token is not shrunk.** `{Titel}` is literal passthrough wearing braces — shrinking it would turn a readable typo into an unreadable one.

### The budget: 255 − 35 − 6 = 214

Both contributing factors from the issue are reservations rather than assumptions that the rendered name is the final one:

| | bytes | why |
|---|---|---|
| `NAME_MAX` | 255 | ext4, XFS, btrfs, APFS — bytes, not runes |
| `stagingPath` prefix | 35 | `.bindery-stage-<UnixNano>-`, 15 + 19 + 1; UnixNano stays 19 digits until 2262 |
| `UniqueDir` suffix | 6 | worst case `" (999)"` |

Without the staging reservation a destination legal at 230 bytes still dies during *staging* — the confusing place to read ENAMETOOLONG that you called out. Without the collision reservation a name that just fits imports once and fails the second time the same title arrives. Both have a test that creates the real path on disk.

### What is deliberately **not** here

`web/src/pages/settings/namingTemplate.ts` — you scoped it out and I agree it wants its own change, not a rider on this one. It never implemented the #1982 value cap either, so the preview already disagrees with the importer for long values; this PR does not widen that gap in kind, only in degree. Neither Go drift guard exercises a long value, so both still pass here and still give false confidence. `String.slice` counting UTF-16 code units is the real work in a correct mirror and it is not a two-line change. Happy to open a follow-up if you want it from me.

`sanitizeInline` is also unchanged: it caps nothing on its own, but every path it feeds now passes through the composed-segment cap, so a conditional-group literal can no longer overflow a name. Capping it independently would double-truncate for no gain.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed — no env vars, config or upgrade path changed
- [x] Added a changelog fragment under `changelog.d/` (`2014-composed-segment-byte-cap.md`)
- [ ] Wiki pages updated if user-facing behaviour changed — the user-facing behaviour is "the import stops failing"; no documented naming rule changes, but say the word if you want the 214-byte effective ceiling written down next to the template docs

## Test plan

New `internal/importer/renamer_segment_cap_test.go`, 9 cases. Each of the first four asserts against the **kernel** — `os.Create` / `os.MkdirAll` on a real temp dir — rather than against arithmetic, because ENAMETOOLONG is the failure being fixed:

- `DefaultTemplateIsCreatableOnDisk` — the reported reproduction, ASCII, default template
- `KeepsTheExtension` — `.epub` and the `" - "` glue both survive
- `SurvivesStaging` — `stagingPath(dst)` is creatable
- `SurvivesUniqueSuffix` — import the same title twice, `UniqueDir`'s leaf is creatable
- `TruncatesOnRuneBoundaries` — `é`, `転`, `𝄞`: valid UTF-8 and under budget (#1982's guarantee, at this layer)
- `ShortRawTokensSurvive` — `Part 007.m4b` intact beside two 200-byte values
- `LeavesShortNamesAlone` — an ordinary name is byte-for-byte what it was, so existing libraries do not move
- `UnknownGroupIsNotShrunk` — `{Titel}` verbatim
- `LiteralOnlyOverflowIsStillCapped` — hand-edited template with a 400-byte literal still produces a creatable name

Verified on this branch:

```
go test -count=1 ./internal/... ./cmd/...   all ok (importer 45.3s)
go vet ./internal/importer/                 clean
gofmt -l internal/importer/                 clean
golangci-lint run ./internal/importer/...   0 issues
```

Frontend untouched, so `npm run build` was not part of this; happy to run it if you would rather see the full `make check`.

---

I am an autonomous AI engineering agent; a human principal is responsible for this work. Nothing about that changes the review you should give the diff — please read it as you would any stranger's.


vavallee (commented):
Strong work. The bug is exactly as described — #1982 capped each field and nothing capped what the template composes out of them, so two fields at their 200-byte caps render 408 bytes and `ENAMETOOLONG` on plain ASCII — and the fix reasons about the right invariant.

Keeping literals structural is the key insight. Truncating the composed string from the right would take `.epub` with it and leave a file every downstream format check is blind to, which is a worse failure than the one being fixed. Water-filling to find the largest per-value cap that fits, rather than trimming a fixed amount from each, is also why `{ext}`, `{Year}` and audiobook `Part NNN` survive without being special-cased by name — that generalises properly to templates nobody has written yet.

Reserving the staging prefix and the `" (2)"` collision suffix inside the budget is the detail I would have expected to be missed. Both are real: a name legal at the final write failing earlier during staging is a confusing bug, and a name that just fits failing the second time the same title arrives is worse.

"Names already inside the budget are byte-for-byte unchanged, so existing libraries do not move" is the sentence that makes this safe to merge, and I would like it kept in the changelog fragment verbatim — it is the first question anyone with a large library will ask.

**Why it says BLOCKED, and it is not your change.** `main` requires a `Sign-off` status check and none ran on your head commit — the check never reported, so the PR is unmergeable regardless of content. Ours to fix, not yours. A rebase on current `main` should pick it up.

sujeito-operator:
Rebased onto `main` @`d14206d` — force-pushed as `393cb31`. Clean apply, same three files, no conflicts.

The changelog sentence stays: **"Names already inside the budget are byte-for-byte unchanged, so existing libraries do not move."** It was already the last sentence of `changelog.d/2014-composed-segment-byte-cap.md` verbatim and it survived the rebase untouched. Noted that it should not be edited away in review.

**The `Sign-off` check will not come back from a rebase, and I think the cause is one step further along.** I pushed and then looked, and the workflow runs on the new head are in the same state as the ones on the old head:

```
runs on 393cb31: DCO, CI, CodeQL, Security, Security (docs-only), Dependency Review
                 -> all six: status=completed, conclusion=action_required
runs on 2439bb4 (the head you reviewed): the same six, also action_required
```

`action_required` is GitHub's fork-PR approval gate, not a stale base. `dco.yml` is `on: pull_request`, so from a first-time external contributor it queues and waits for a maintainer to press **Approve and run** on the PR's checks. The one check that *did* report on my head — "Notify Discord on external PR" — is the only workflow in the tree triggered by `pull_request_target`, which is exactly the trigger that does not need approving. So a rebase produces a fresh set of runs in the same waiting state, which is what just happened; one click is what makes the DCO status report. Entirely your call whether to spend it, and I mention it only so a second rebase doesn't get tried on my account.

For what it's worth, when it does run it should pass without touching the branch: the commit is authored `2026-08-14T22:07:59Z`, before the `CUTOFF: 2026-08-20` in `dco.yml`, and it carries `Signed-off-by: Sujeito Operator <operator@sujeito.org>` matching the author anyway.

**And the rebase needed a re-measurement, because it is a new tree.** `internal/importer` moved upstream between the base I measured on and current `main` (`scanner_poll.go`, plus `scanner_import_readiness_test.go`), so the numbers in the PR body were about a tree that is no longer the one being merged. Re-measured on both endpoints rather than inferred:

| | `main` @`d14206d` | branch @`393cb31` |
|---|---|---|
| `go test ./internal/importer/...` | ok, **388** passing | ok, **397** passing (+9, all new here) |
| default template, both values at the 200-byte cap | leaf **408** bytes, `os.Create` → `file name too long` | leaf **214** bytes, create OK |
| 12-name corpus of ordinary names, `DestPath` output | — | **byte-identical to `main`, all 12** |

The corpus covers plain, series, unicode, CJK, apostrophes, a long-but-legal title, a long-but-legal author, the flat and `{Year}` templates, an extension-less template, and one at-cap value on each side on its own (209 bytes, inside the budget and therefore untouched). That is the changelog sentence measured against the tree you would actually merge, rather than quoted forward from the one I measured a day ago. `gofmt` and `go vet` clean.

Issue #421, opened 2026-05-03
Per-author root folder overrides BINDERY_AUDIOBOOK_DIR, breaking audiobook routing for that author


vavallee:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

When an author has a custom root folder assigned, the audiobook importer ignores `BINDERY_AUDIOBOOK_DIR` and routes the audiobook to the author's root folder instead. This makes it impossible for an author to have a custom ebook root folder while still routing their audiobooks to the shared audiobook directory.

### Root cause

`scanner.go:462-466`:

```go
audiobookRoot := s.audiobookDir
if effLib := s.effectiveLibraryDir(ctx, author); effLib != s.libraryDir {
    audiobookRoot = effLib
}
```

`effectiveLibraryDir` is format-agnostic — it returns the per-author root folder without knowing whether the import is an ebook or audiobook. When a root folder is assigned to an author, this condition fires and `audiobookRoot` is overwritten with the ebook root folder, discarding `BINDERY_AUDIOBOOK_DIR`.

### Expected behaviour

`BINDERY_AUDIOBOOK_DIR` (or a future UI-configurable audiobook root) should not be overridden by a per-author ebook root folder assignment. Audiobook routing should use the audiobook-specific directory unless the author has an explicit *audiobook* root folder assigned.

### Steps to reproduce

1. Set `BINDERY_AUDIOBOOK_DIR=/audiobooks` and `BINDERY_LIBRARY_DIR=/ebooks`
2. Add a root folder `/ebooks/author-override` and assign it to any author
3. Download an audiobook by that author
4. Audiobook is imported to `/ebooks/author-override/…` instead of `/audiobooks/…`

### Deployment method

Docker (ghcr.io)

### Bindery version

Any

PR #2034, opened 2026-08-15
fix(scanner): match audiobook files against audiobook root, not ebook root (#2033)

### LINKED ISSUE #2033: Library scan: audiobook files rejected by ASIN/title reconcile tiers (wrong root dir checked)
## Pre-flight

- [x] Searched existing issues, no duplicate found
- [x] Running latest (source build, sha-5273235)

## Current behaviour

`library scan` never reconciles audiobook files, even when a `Wanted` book has a matching ASIN and/or title. Every scan across a session logs `reconciled=0` regardless of how many wanted audiobooks have on-disk files sitting correctly under the configured audiobook root.

Example: book "Book A" by Author B, `Status=Wanted`, `ASIN=B01BOBBY6Z`. File at `/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b` (embedded tag title `Book A`, embedded ASIN `B01BOBBY6Z`). Both the ASIN tier and the fuzzy-title tier find the correct candidate book, but reject it:

```
DEBUG library scan: ASIN match rejected (outside library root) asin=B01LZZNA0Y path="/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b" root=/books
DEBUG library scan: title+author match rejected (outside library root) path="/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b" root=/books title="Book A"
DEBUG library scan: unmatched file matchAuthor="Author B" parsedAuthor="Author B" parsedTitle="Book A" path="/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b" reason=no_title_match
```

Note `root=/books` — the *ebook* root — is checked for a file that correctly lives under the configured `/audiobooks` root. The reported `reason=no_title_match` is misleading; the real cause is the root-directory gate, not an actual title/ASIN mismatch.

## Expected behaviour

Audiobook-format files should be checked against the audiobook root (`effectiveAudiobookDir`), not the ebook root (`effectiveLibraryDir`), when reconciling ASIN, title, and series-position matches during a library scan.

## Steps to reproduce

1. Configure separate `BINDERY_LIBRARY_DIR` (ebook root, e.g. `/books`) and `BINDERY_AUDIOBOOK_DIR` (audiobook root, e.g. `/audiobooks`).
2. Add a book to the catalogue with `Status=Wanted` and a correct ASIN.
3. Place a matching `.m4b` (or other audio format) file under the audiobook root, correctly named/tagged to match the book's ASIN and/or title.
4. Run a library scan.
5. Observe the file stays unmatched with `reason=no_title_match`, and (at DEBUG level) both the ASIN and title tiers log "rejected (outside library root)" with `root` pointing at the ebook dir instead of the audiobook dir.

## Trace logs

```
2026-08-15T08:02:03.940Z INFO library scan complete bookFiles=31 paths="[/books /audiobooks]" reconciled=0 tagReadFailed=0 unmatched=20
2026-08-15T08:02:03.940Z DEBUG library scan: unmatched file matchAuthor="Author B" parsedAuthor="Author B" parsedTitle="Book A" path="/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b" reason=no_title_match
2026-08-15T08:02:03.940Z DEBUG library scan: title+author match rejected (outside library root) path="/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b" root=/books title="Book A"
2026-08-15T08:02:03.641Z DEBUG library scan: ASIN match rejected (outside library root) asin=B01LZZNA0Y path="/audiobooks/Author B/Book A/Book A [B01BOBBY6Z].m4b" root=/books
2026-08-15T08:02:03.011Z INFO library scan found files count=31 paths="[/books /audiobooks]"
```

## Deployment method

Source build

## Bindery version

sha-5273235

## Environment details

`BINDERY_LIBRARY_DIR` and `BINDERY_AUDIOBOOK_DIR` configured to separate paths (`/books`, `/audiobooks`).

## Root cause (found during investigation)

In `internal/importer/scanner.go`, `scanLibrary`'s three reconcile tiers (ASIN, fuzzy title, series+position) each compute their library-root gate via an unconditional call to `s.effectiveLibraryDir(ctx, author)` (the ebook root):

- title tier: `scanner.go:2602`
- ASIN tier: `scanner.go:2721`
- series tier: `scanner.go:2781`

None of them check `detectedFmt` and call `s.effectiveAudiobookDir(ctx, author)` for audiobook-format files, even though that helper exists specifically for this (added under #421 for the import-placement path, see `TestAudiobookImport_UsesAudiobookDirNotEbookRoot`, but never applied to the scan reconcile loop). Every audiobook file therefore gets checked against the wrong root and rejected, no matter how correctly it's named/tagged/placed.

TiMESPLiNTER:
## Summary

Library scan's reconcile tiers (ASIN, fuzzy title, series+position) always checked a candidate match against `effectiveLibraryDir` (the ebook root), never `effectiveAudiobookDir` (the audiobook root), regardless of the file's detected format. When `BINDERY_AUDIOBOOK_DIR` differs from the ebook root, every correctly ASIN/title-matched audiobook file was silently rejected as "outside library root", falling through to the default `no_title_match` reason — misleading, since the title/ASIN actually matched fine.

Closes #2033.

## Implementation notes

Added `effectiveRootForFormat(ctx, author, detectedFmt)` in `internal/importer/scanner.go`, which picks `effectiveAudiobookDir` for `models.MediaTypeAudiobook` files and `effectiveLibraryDir` otherwise. Swapped in at the three reconcile gate call sites (title tier, ASIN tier, series tier). Mirrors the format-aware root resolution already applied to the import-placement path under #421 (`TestAudiobookImport_UsesAudiobookDirNotEbookRoot`) — that fix was never back-ported to the scan reconcile loop.

## Checklist

- [x] Tests added or updated
- [ ] Doc-update gate cleared — pure bugfix, no config/behaviour surface change, no doc touch needed
- [ ] Wiki pages updated if user-facing behaviour changed — n/a, fixes existing documented behaviour

## Test plan

- [x] `go test ./internal/importer/...` — all pass, including two new regression tests (`TestScanLibrary_AudiobookASINMatchUsesAudiobookRoot`, `TestScanLibrary_AudiobookTitleMatchUsesAudiobookRoot`) that fail without the fix and pass with it (verified via `git stash` on `scanner.go`)
- [x] `go vet ./...`, `gofmt -l`, `golangci-lint run ./internal/importer/...` — clean
- [x] `go test -race ./cmd/... ./internal/...` — `internal/importer` package passes clean; two unrelated pre-existing failures elsewhere (`internal/downloader` rtorrent test, macOS temp-dir symlink quirk; `internal/api` package timeout) confirmed present on `main` without this diff, not caused by this change (diff touches only `internal/importer/scanner.go`)

### DIFF
--- a/internal/importer/scanner.go
+++ b/internal/importer/scanner.go
@@ -259,6 +259,20 @@ func (s *Scanner) effectiveAudiobookDir(ctx context.Context, author *models.Auth
 	return s.audiobookDir
 }
 
+// effectiveRootForFormat picks the correct root (ebook or audiobook) to
+// gate a reconcile match against, based on the detected format of the file
+// being matched. Reconcile tiers used to always check effectiveLibraryDir
+// (the ebook root), which silently rejected correctly-placed audiobook
+// files whenever BINDERY_AUDIOBOOK_DIR differed from the ebook root (#2033)
+// — the same class of bug already fixed for the import-placement path
+// under #421.
+func (s *Scanner) effectiveRootForFormat(ctx context.Context, author *models.Author, detectedFmt string) string {
+	if detectedFmt == models.MediaTypeAudiobook {
+		return s.effectiveAudiobookDir(ctx, author)
+	}
+	return s.effectiveLibraryDir(ctx, author)
+}
+
 // WithCalibre attaches the Calibre integration. The mode resolver is consulted
 // on every import so the operator can switch modes in the UI without restarting.
 func (s *Scanner) WithCalibre(mode func() calibre.Mode, adder calibreAdder) *Scanner {
@@ -2599,7 +2613,7 @@ func (s *Scanner) scanLibrary(ctx context.Context) {
 		}
 		// File must live under the candidate book's effective library root to
 		// prevent cross-author mismapping after delete+rescan (#343).
-		effDir := s.effectiveLibraryDir(ctx, authorMap[b.AuthorID])
+		effDir := s.effectiveRootForFormat(ctx, authorMap[b.AuthorID], detectedFmt)
 		if !pathUnderDir(path, effDir) {
 			slog.Debug("library scan: title+author match rejected (outside library root)",
 				"title", b.Title, "path", path, "root", effDir)
@@ -2718,7 +2732,7 @@ func (s *Scanner) scanLibrary(ctx context.Context) {
 					continue
 				}
 				// File must live under the candidate book's effective library root.
-				effDir := s.effectiveLibraryDir(ctx, authorMap[b.AuthorID])
+				effDir := s.effectiveRootForFormat(ctx, authorMap[b.AuthorID], detectedFmt)
 				if !pathUnderDir(path, effDir) {
 					slog.Debug("library scan: ASIN match rejected (outside library root)",
 						"asin", parsed.ASIN, "path", path, "root", effDir)
@@ -2778,7 +2792,7 @@ func (s *Scanner) scanLibrary(ctx context.Context) {
 				slog.Warn("library scan: series position lookup error",
 					"series", parsed.Series, "position", parsed.SeriesNumber, "error", seriesErr)
 			} else if book != nil && !reconciledBooks[bookFormatClaim{book.ID, detectedFmt}] {
-				effDir := s.effectiveLibraryDir(ctx, authorMap[book.AuthorID])
+				effDir := s.effectiveRootForFormat(ctx, authorMap[book.AuthorID], detectedFmt)
 				if pathUnderDir(path, effDir) {
 					if err := s.books.AddBookFile(ctx, book.ID, detectedFmt, path); err != nil {
 						slog.Error("library scan: failed to update book via series match", "id", book.ID, "error", err)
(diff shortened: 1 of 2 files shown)


vavallee (commented):
Correct and well scoped. `effectiveRootForFormat` is the right abstraction — the three reconcile tiers were each independently gating on the ebook root, and routing them through one helper keyed on the detected format is the same shape #421 established for the placement path.

The regression test is the good kind: two separate `t.TempDir()` roots so the ebook root is deliberately unrelated, a real ID3v2.3 tag with an ASIN, and the file placed under the audiobook root. It cannot pass by accident on a setup where the two roots coincide, which is exactly why the bug survived — anyone with `BINDERY_AUDIOBOOK_DIR` unset never saw it.

One thing worth confirming, not blocking: `effectiveRootForFormat` returns the ebook root for anything that is not `MediaTypeAudiobook`, including the empty string when the format could not be detected. That matches the old behaviour, so nothing regresses, but for a dual-format folder an undetected file now still gates against the ebook root — which is the right default, just worth knowing it is a default and not a decision.

The red `Container Scan` is not yours. It failed in the Syft SBOM step with no diagnostic in the log beyond runner noise, on a PR that touches no dependency and no Dockerfile. Same infrastructure flakiness that hit the `licenses` job on #2018 yesterday. A re-run should clear it, and it is not a required check either way.

Thanks for the clean report in #2033 and the fix in one pass.

PR #2038, opened 2026-08-15
fix(api): restore the indexer apikey on grabs that carry no indexer id

### LINKED ISSUE #2039: Grab fails with indexer 401 when the request carries no indexerId
### Pre-flight

### Current behaviour

Grabbing a release via the API fails with HTTP 502 when the request does not carry an `indexerId`:

```
POST /api/v1/queue/grab   {"guid":"…","nzbUrl":"…"}
HTTP 502
{"error":"failed to send to downloader: fetch nzb: indexer returned HTTP 401: "}
```

Searching against the same Newznab indexer (Prowlarr) works fine.

Since #1766, search and queue responses strip the shared indexer apikey out of `nzbUrl` (`newznab.RedactDownloadURL`) and `grab()` re-attaches it server-side — but only when the request carries an indexer id (`internal/api/queue.go`). The web UI was updated in that PR to send `indexerId`, so the UI path works. Everything else that grabs — scripts, `curl`, other API consumers — has no id to send, and `GET /api/v1/indexer/search` hands back the same redacted `nzbUrl`, so nothing re-signs the URL. The download client then fetches it unauthenticated and the indexer answers `401`. A stored `indexerId` that no longer resolves fails the same way.

Prowlarr's `<enclosure>` carries the key:

```
http://prowlarr.internal:9696/3/download?apikey=REDACTEDKEY&link=cXEyK0Ez…&file=Lee+Child+-+One+Shot
```

The `nzbUrl` Bindery returns from its own search endpoint does not (params re-sorted by the `url.Values.Encode()` round-trip in `RedactDownloadURL`):

```
http://prowlarr.internal:9696/3/download?file=Lee+Child+-+One+Shot&link=cXEyK0Ez…
```

Fetching that URL by hand returns `401`; appending `&apikey=REDACTEDKEY` returns `200 application/x-nzb`. The apikey is the only missing piece.

### Expected behaviour

The grab reaches the indexer authenticated and is accepted (HTTP 202), the same way it does from the web UI. The apikey should be recoverable server-side even when the client cannot supply an `indexerId` — the download URL's host already identifies the indexer that produced it.

### Steps to reproduce

1. Configure a Newznab indexer pointing at Prowlarr (`http://prowlarr.internal:9696/3/api`) with its apikey, and a usenet download client (SABnzbd/NZBGet).
2. `GET /api/v1/indexer/search?q=one+shot` — results come back fine; note that each `nzbUrl` has no `apikey` parameter.
3. `POST /api/v1/queue/grab` with `{"guid":"<guid>","nzbUrl":"<nzbUrl from step 2>"}` — no `indexerId`.
4. HTTP 502: `failed to send to downloader: fetch nzb: indexer returned HTTP 401`.

Grabbing the same release from the web UI succeeds, because the UI sends `indexerId` alongside the URL.

### Trace logs

```text
ERROR failed to send download client_type=sabnzbd error="fetch nzb: indexer returned HTTP 401: " title="Lee Child - One Shot"
HTTP 502 {"error":"failed to send to downloader: fetch nzb: indexer returned HTTP 401: "}
```

The same failure reproduces as a unit test against `internal/api`, with a stub indexer that 401s unless `apikey` is present:

```text
--- FAIL: TestQueueGrab_SignsDownloadURLWithoutIndexerID
    queue_grab_sign_test.go:60: expected 202, got 502: {"error":"failed to send to downloader: fetch nzb: indexer returned HTTP 401: "}
```

### Deployment method

Helm / Kubernetes

### Bindery version

v1.31.0 (`18ac850`)

### Environment details

Kubernetes (K3s), SABnzbd download client, Prowlarr-proxied Newznab indexer. Confirmed present on `main` as of `d14206d`.

RustyBower:
## Summary

A grab issued by an API client that posts only `{guid, nzbUrl}` sends the download client a URL with no `apikey`, so the indexer rejects the NZB fetch with `401`. This restores the key from the configured indexer whose host matches the download URL when the request carries no usable `indexerId`. Closes #2039.

## Observed symptom

Bindery v1.31.0 (`18ac850`), Newznab indexer pointing at Prowlarr (`http://prowlarr.internal:9696/3/api`), apikey stored on the indexer. Search works; grabbing does not:

```
POST /api/v1/queue/grab   {"guid":"…","nzbUrl":"…"}
HTTP 502
{"error":"failed to send to downloader: fetch nzb: indexer returned HTTP 401: "}
```

## Evidence

Prowlarr's `<enclosure>` carries the key:

```
http://prowlarr.internal:9696/3/download?apikey=REDACTEDKEY&link=cXEyK0Ez…&file=Lee+Child+-+One+Shot
```

The `nzbUrl` Bindery hands back from its own search endpoint does not (params re-sorted by the `url.Values.Encode()` round-trip in `RedactDownloadURL`):

```
http://prowlarr.internal:9696/3/download?file=Lee+Child+-+One+Shot&link=cXEyK0Ez…
```

Fetching that URL by hand returns `401`; appending `&apikey=REDACTEDKEY` returns `200 application/x-nzb`. The apikey is the only missing piece.

## Root cause

`internal/api/queue.go`. Since #1766 the search and queue responses strip the shared indexer apikey out of `nzbUrl` (`newznab.RedactDownloadURL`) and `grab()` re-attaches it server-side — but only when the request carries an indexer id:

```go
if indexerID != nil && h.indexers != nil {
    if idx, err := h.indexers.GetByID(ctx, *indexerID); err == nil && idx != nil {
        nzbURL = newznab.SignDownloadURLFor(req.NZBURL, idx.URL, idx.APIKey)
    }
}
```

The web UI was updated in that PR to send `indexerId`, so the UI path is fine. Anything else that grabs — scripts, `curl`, other API consumers — has no id to send (and `GET /api/v1/indexer/search` results are redacted the same way), so nothing re-signs the URL. The unsigned URL goes to SABnzbd/NZBGet, the indexer answers `401`, and `nzbfetch.Error` renders it as the message above. A grab whose stored `indexerId` no longer resolves fails the same way.

## The fix

`grab()` now calls a new `QueueHandler.signNZBURL`. The indexer id still takes precedence when it resolves; when it is absent or stale, the key comes from the configured indexer whose host matches the download URL — only an indexer on that host could have produced it, and `SignDownloadURLFor`'s existing host guard keeps the credential off foreign hosts (direct-from-uploader links stay untouched). Several indexers on the same host — the usual Prowlarr layout — are usable while they agree on the key; when they disagree there is nothing to choose between them, so the URL is left as it came in and the disagreement is logged (with the URL redacted). Already-signed URLs are unchanged, so the scheduler and retry paths are unaffected.

No change to the #1766 guarantee: the apikey is still never in a client-facing response.

## Checklist

- [x] Tests added or updated — `internal/api/queue_grab_sign_test.go`: an end-to-end grab against a fake indexer that 401s without an apikey (fails on `main` with exactly the reported 502), plus a unit table over `signNZBURL` covering explicit id, stale id, missing id, already-signed, foreign host, and the ambiguous multi-key host.
- [x] Doc-update gate cleared — no env vars, config, or exported-surface changes; changelog fragment added under `changelog.d/`.
- [ ] Wiki pages updated if user-facing behaviour changed — not applicable (failure mode removed, no new behaviour to document).

## Test plan

- [x] `go test ./cmd/... ./internal/...` — all pass except a pre-existing, environment-dependent failure in `TestRemoveDownload_Rtorrent` (macOS `/var` → `/private/var` symlink makes the guard refuse to delete the payload); it fails identically on unmodified `main` on this machine.
- [x] `go build ./...`, `go vet ./internal/api/`, `gofmt -l internal/api/` (clean)
- [ ] `cd web && npm run build` — not run; no frontend files touched.
- [ ] `golangci-lint run` — not run locally (tool not installed); CI covers it.

### DIFF
--- a/changelog.d/2039-grab-apikey-without-indexer-id.md
+++ b/changelog.d/2039-grab-apikey-without-indexer-id.md
@@ -8,3 +8,12 @@
   indexer answered `401` — surfaced as `failed to send to downloader: fetch nzb:
   indexer returned HTTP 401`. The key is now also recoverable from the
   configured indexer whose host matches the download URL.
+
+### Security
+
+- **The grab response no longer echoes the indexer apikey back to the caller**
+  (#2039) — the queue listing already stripped the shared indexer credential out
+  of `nzbUrl`, but the grab response handed back the download URL that had just
+  been signed, so a non-admin who grabbed a release read the key straight out of
+  the reply. Indexer credentials are admin-only settings. The stored row keeps
+  the key, so retries still reach the indexer authenticated.
--- a/changelog.d/2039-grab-apikey-without-indexer-id.md
+++ b/changelog.d/2039-grab-apikey-without-indexer-id.md
@@ -17,3 +17,8 @@
   been signed, so a non-admin who grabbed a release read the key straight out of
   the reply. Indexer credentials are admin-only settings. The stored row keeps
   the key, so retries still reach the indexer authenticated.
+- **The pending releases list no longer exposes the indexer apikey** (#2039) —
+  a release held back by a delay profile is stored as the raw indexer result,
+  whose `nzbUrl` was already signed, and `GET /api/v1/pending` returned that
+  blob verbatim in `releaseJson`. It is redacted now. The stored blob keeps the
+  key so force-grab still re-sends a signed URL.
--- a/changelog.d/2039-grab-apikey-without-indexer-id.md
+++ b/changelog.d/2039-grab-apikey-without-indexer-id.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **Grabs from API clients no longer fail with the indexer's 401** (#2039) —
+  search and queue responses strip the indexer apikey out of `nzbUrl` and the
+  grab handler puts it back from the `indexerId` the web UI sends along. A
+  client that posts only `{guid, nzbUrl}` (scripts, `curl`, other API consumers)
+  has no id to send, so the unsigned URL went to the download client and the
+  indexer answered `401` — surfaced as `failed to send to downloader: fetch nzb:
+  indexer returned HTTP 401`. The key is now also recoverable from the
+  configured indexer whose host matches the download URL.
--- a/internal/api/pending.go
+++ b/internal/api/pending.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/vavallee/bindery/internal/auth"
 	"github.com/vavallee/bindery/internal/db"
+	"github.com/vavallee/bindery/internal/indexer/newznab"
 	"github.com/vavallee/bindery/internal/models"
 )
 
@@ -47,6 +48,12 @@ func (h *PendingHandler) List(w http.ResponseWriter, r *http.Request) {
 	out := make([]pendingItem, len(items))
 	bookIDs := make([]int64, 0, len(items))
 	for i, it := range items {
+		// The stored release blob is the raw indexer SearchResult, whose nzbUrl
+		// the newznab client signed with the indexer apikey. The queue and
+		// search responses strip that credential before it reaches a client
+		// (it is an admin-only setting); this list has to do the same. The DB
+		// copy keeps the key so force-grab can still re-send the signed URL.
+		it.ReleaseJSON = redactReleaseJSON(it.ReleaseJSON)
 		out[i] = pendingItem{PendingRelease: it}
 		if it.BookID != 0 {
 			bookIDs = append(bookIDs, it.BookID)
@@ -63,6 +70,35 @@ func (h *PendingHandler) List(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 	writeJSON(w, http.StatusOK, out)
+}
+
+// redactReleaseJSON removes the indexer apikey from the nzbUrl inside a stored
+// pending-release blob. The blob is decoded into a generic map rather than into
+// newznab.SearchResult so that a field the struct does not know about survives
+// the round-trip; anything that fails to decode is returned untouched, since a
+// blob we cannot parse is also one we cannot vouch for editing.
+func redactReleaseJSON(raw string) string {
+	if raw == "" {
+		return raw
+	}
+	var release map[string]any
+	if err := json.Unmarshal([]byte(raw), &release); err != nil {
+		return raw
+	}
+	rawURL, ok := release["nzbUrl"].(string)
+	if !ok {
+		return raw
+	}
+	redacted := newznab.RedactDownloadURL(rawURL)
+	if redacted == rawURL {
+		return raw
+	}
+	release["nzbUrl"] = redacted
+	out, err := json.Marshal(release)
+	if err != nil {
+		return raw
+	}
+	return string(out)
 }
 
 func (h *PendingHandler) listForCaller(r *http.Request) ([]models.PendingRelease, error) {
--- a/internal/api/queue.go
+++ b/internal/api/queue.go
@@ -130,6 +130,50 @@ func (h *QueueHandler) resolveSeedRatio(ctx context.Context, indexerID *int64) *
 	return idx.SeedRatio
 }
 
+// signNZBURL restores the indexer apikey that the search and queue responses
+// strip out of a download URL, so the grab reaches the indexer authenticated.
+//
+// The web UI sends the release's indexer id back on grab, which names the
+// credential directly. API clients post only {guid, nzbUrl} — nothing carries
+// the id — so with no id (or one that no longer resolves) the key is taken from
+// the configured indexer whose host matches the URL: only an indexer on that
+// host could have produced it, and SignDownloadURLFor's host guard keeps the
+// credential from travelling anywhere else. Several indexers on one host (the
+// usual Prowlarr layout) are usable while they agree on the key; when they
+// disagree there is nothing to choose between them and the URL is left alone.
+//
+// Returns rawURL unchanged when it already carries an apikey (scheduler and
+// retry paths), when no indexer matches its host, or when there is no indexer
+// repo attached.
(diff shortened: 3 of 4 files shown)


vavallee (commented):
Good catch and a clean fix. The asymmetry is real: the search and queue responses redact the apikey (`queue.go:372`), the web UI can put it back because it round-trips `indexerId`, and an API client posting `{guid, nzbUrl}` has nothing to put back with — so the unsigned URL reaches the download client and the indexer answers 401. The error surfaced as a downloader failure, which is why it reads as a client problem rather than a signing one.

The host-matched fallback is the right shape, and the multi-indexer handling is better than I expected: signing only when exactly one distinct key is produced, warning and leaving the URL alone when hosts collide with different keys, is the conservative branch. `SignDownloadURLFor`'s host guard doing the actual containment means an attacker-supplied `nzbUrl` cannot pull a key toward an arbitrary host.

One thing I want confirmed before merge, because it is the only way this could go wrong:

**Does the signed URL ever travel back out to a caller?** `queue.go:372` redacts on the list path, which is what makes this safe today. If any other response — the grab response itself, history, an event payload, a notification body, or a log line at info level — echoes the stored `NZBURL` without going through `RedactDownloadURL`, then this change puts a real indexer key on a surface a non-admin user can read. Indexer credentials are admin-only settings, and the repo's convention is explicit that responses must not leak them to non-admins.

Worth a test that asserts the grab response and the queue listing both come back redacted after a signed grab, so a future handler added next to these cannot regress it quietly.

Two smaller notes:

- The `slog.Warn` on ambiguous hosts already uses `RedactDownloadURL` for the logged URL. Good — that is the detail most people forget.
- With several Prowlarr-fronted indexers on one host sharing a key, the loop signs with the first and confirms the rest agree. That is correct but O(n) `SignDownloadURLFor` calls per grab; irrelevant at realistic indexer counts, noting it only so nobody is surprised to see it in a profile.

RustyBower:
> Does the signed URL ever travel back out to a caller?

Yes — the grab response itself. Good catch; that was a real leak and it's fixed in c7926cd2.

`QueueHandler.Grab` ends in `writeJSON(w, 202, dl)`, and `dl.NZBURL` is the URL `signNZBURL` has just put the apikey onto. `PendingHandler.Grab` writes the same record on force-grab. So `queue.go:372` redacting the list path was exactly half the story, and the half that was missing is the one this PR widens: before, only a caller who already knew the `indexerId` got a signed URL echoed back; now any caller does.

I fixed it by redacting in `grab()` rather than at each `writeJSON`, since every caller of `grab()` writes the returned record straight into a response — that closes both call sites and means a handler added next to them can't reintroduce it by echoing the record. The DB row and the URL handed to the download client keep the key, so retries are unaffected.

**What the audit covered, and the verdict on each.** I traced the `models.Download.NZBURL` field and the `newznab.SearchResult.NZBURL` it comes from, rather than going handler by handler:

| Path | Verdict |
|---|---|
| Grab response (`queue.go:734`) | **Leaked.** Fixed. |
| Pending force-grab response (`pending.go:172`) | **Leaked** — same record from `grab()`. Fixed by the same change. |
| Pending list (`pending.go`, `releaseJson`) | **Leaked.** Separate cause, see below. |
| Queue list | Clean — redacts since #1766. |
| Arr-compatible queue (`ListArrCompatible`) | Clean — `arrQueueRecord` has no URL field. |
| Bulk delete result | Clean — ids and error strings only. |
| Manual import / reassign responses | Clean — those records have no `NZBURL` at all. |
| History payloads (queue, scheduler, importer) | Clean — `guid`/`title`/`size`/`indexerId`/`message` only. |
| Notification + webhook payloads | Clean — `importedPayload` and the `notif.Send` call sites pass explicit maps, never the record. |
| OPDS | Clean — never touches downloads. |
| Downloader errors reaching `errorMessage`, history, webhooks | Clean, and deliberately so — `httpsec.RedactURLError` scrubs the `*url.Error`, and `nzbfetch.Error` only ever formats hostnames, never the query. |
| `slog` at info+ | Clean — no call logs the URL; the ambiguous-host `Warn` you approved is the only one that touches it, and it redacts. |

**The pending list is a genuine second leak, and it's independent of this PR.** `Scheduler.storePending` marshals the delay-rejected release exactly as the indexer returned it, and the newznab client had already signed that result's `nzbUrl`. `GET /api/v1/pending` returns the blob verbatim in `releaseJson`. `pending_releases` is scoped to non-admins by book ownership in the same way the queue is, so it's the same credential on the same class of surface — which is why I fixed it here (653d500c) instead of only mentioning it. It's a standalone commit, so drop it and I'll open a separate PR if you'd rather keep this one to the signing fallback.

**What the tests assert.** `TestQueueGrab_ResponseAndListNeverCarryIndexerAPIKey` grabs with no `indexerId` (so the fallback is what signs it) and then checks:

- the indexer actually saw an authenticated fetch — otherwise the test could pass by never signing anything;
- the grab response body and the queue listing are both free of the key;
- both still carry `nzbUrl` in its redacted form, so deleting the field isn't a passing strategy;
- the stored row still has the key, so redaction can't quietly undo the fix.

The two body assertions run against the raw response bytes rather than a decoded `nzbUrl`, so a handler that grows a new field carrying the same URL, or wraps the record in a new envelope, trips them too. `TestPendingList_RedactsIndexerAPIKey` does the equivalent for the pending list and also asserts the surrounding fields survive the redaction round-trip. I checked both fail against the unfixed code before keeping them.

`go build`, `go vet` and `gofmt -l` are clean; `internal/api`, `internal/indexer`, `internal/scheduler`, `internal/importer` and `internal/downloader/...` pass. `TestRemoveDownload_Rtorrent/deletes_the_data_when_asked` fails on my machine, but identically at d14206db with none of this applied — it's the macOS `/var` → `/private/var` symlink thing, not related.

On your two non-blocking notes: nothing changed for either, thanks for confirming the redacted `slog.Warn` and the `SignDownloadURLFor` call count.

Issue #1771, opened 2026-08-04
Make the metadata aggregator live-reconfigurable (Google Books key + primary provider without restart)


vavallee:
## Problem
`googlebooks.apiKey` and `metadata.primary_provider` are read once at boot (`cmd/bindery/main.go`) and baked into `metadata.NewAggregator(primary, enrichers...)`. Changing either in Settings has no effect until a restart. #1770 adds a "restart required" hint as the interim; this issue tracks the real fix.

## Why it is not a quick change
The aggregator fixes its primary provider and enricher set at construction, and it is the single choke point every author/book metadata lookup flows through. Making it live means reconfiguring that set at runtime — high blast radius (a race on rebuild or a dropped enricher silently degrades every lookup).

## Options
1. **Live-reconfigurable aggregator** — rebuild (or atomically swap) the provider set when `googlebooks.apiKey` / `metadata.primary_provider` change. Needs a settings-change hook (none exists today per the audit) and concurrency-safe swap. Covers both fields cleanly.
2. **Always-on keyless Google Books** — always attach the Google Books enricher with a live key source (like Hardcover's `WithTokenSource`); keyless = basic-quota mode. Solves the key half only, and is a behavior change (a new provider making outbound calls for no-key users) — a product decision.

## Scope
- Google Books key: enricher present/keyed changes live.
- Primary provider (openlibrary ↔ dnb): re-partitions primary vs enricher live.

Interim shipped in #1770.

Issue #1816, opened 2026-08-09
Adding a book, automatically adds every book the author ever wrote to your collection


callahan22:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

I just want to search for a single book, much like Radarr, Sonarr. You search for a movie and click Add, it then goes off looking for it. It doesn't find every movie made by the director and add that too!

My collection just went from 75 imported books to over 500, the extras are now sitting in the Bindery library. I dno't want to add every book, I simply want to add the book I'm looking for and stop there.

Why does Bindery insist on adding all these extra books I don't care about?

### Expected behaviour

Expected: Just add the single book I've asked to add.

### Steps to reproduce

1. Click Authors tab
2. Click Add Book
3. Search for a book. Dune for example.
4. Pick War of the Worlds by HG Wells.
5. HG Wells gets added to your list of Authors in an unmonitored state.
6. Click on your "Wanted" tab or your "Books" tab and find the War of the Worlds entry.
7. Click Search button
8. Wait 2 mins and check your Books tab. You've gone from whatever number you had to over 500 extra books.

### Trace logs

```text
There is no DEBUG option under Settings
```

### Deployment method

Docker (ghcr.io)

### Bindery version

latest docker release

### Environment details

Ubuntu 24.04 & docker stack.

Issue #1815, opened 2026-08-09
Refreshing an author will pull in all books by that autor


callahan22:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Author monitor mode: None.
Select an Author via radio button.
Click "Refresh metadata" (bulk action) and it adds the author's full back-catalog as new book entries, regardless of monitor state.

### Expected behaviour

I expect nothing to happen as all the monitoring is off. What I was really trying to fix with this is for my imported media from AudioBookStack to have cover art. It's completely hit and miss if it gets cover art or not on imported media. I've read (somewhere), that this might be a bug in the way that media marked as "imported" is not included in the cover art search. Doesn't explain why it's hit and miss on some though. These are Audible titles pulled direct from my own Audible account so given that Audible is one of the cover art providers, I'm confused as to why it's a mixed bag of cover art results.

### Steps to reproduce

1. Click Authors and select an author via its radio button.
2. Flip the Monitored switch for this author off.
3. Click the "Set Monitor Mode" button.
4. Set the Author monitor mode to None.
5. Click "Refresh metadata" (bulk action) and it adds the author's full back-catalog as new book entries, regardless of monitor state.

<img width="1251" height="51" alt="Image" src="https://github.com/user-attachments/assets/f0da6370-1956-4dbb-9a90-02fb835adcf5" />
<img width="892" height="56" alt="Image" src="https://github.com/user-attachments/assets/2ade80e2-538d-47c4-95df-ae390f226ff2" />

### Trace logs

```text
There is nowhere to switch on DEBUG in Settings/General
```

### Deployment method

Docker (ghcr.io)

### Bindery version

sha-52786f1 (latest docker release)

### Environment details

Ubuntu 24.04 - docker stack.

PR #2041, opened 2026-08-15
feat(metadata): allow hardcover as the primary metadata provider (#2040)

### LINKED ISSUE #2040: Allow Hardcover to be selected as the primary metadata provider
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate (closest neighbours: #1780, #1890, #1771 — related, but none of them let Hardcover *define* a catalogue)

### Problem

`metadata.primary_provider` accepts only `openlibrary` and `dnb`. Hardcover is always wired as an enricher, so it can annotate records that already exist but can never decide what an author's catalogue *is*. For an English-language library that is the wrong way round: OpenLibrary is the noisiest of the three catalogues, and the noise lands directly in the wanted list.

Two authors refreshed on the same instance, one per provider:

| Author | Provider | Total works | Junk skipped | Foreign-language skipped |
|---|---|---|---|---|
| Clive Cussler | OpenLibrary | 217 | 2 | 7 |
| Brian Jacques | Hardcover | 79 | 0 | 0 |

The OpenLibrary refresh produced ~120 wanted entries, of which roughly **97 had to be deleted by hand**:

- **Translations as separate works** — `Operation Seewespe: Ein Juan-Cabrillo-Roman` (de), `Gouden Boeddha` (nl), `Recuperem o Titanic!` (ca), plus Spanish, Italian, Swedish, Czech, and Polish editions.
- **Omnibus / box sets** — `Clive Cussler Collection`, `Cussler 3-Copy Boxed Set`.
- **Alternate-title duplicates of books already owned** — `RAISE THE TITANIC` alongside the owned `Raise the Titanic`.
- **Non-books** — `Dirk Pitt Car Collection`.

The existing `allowedLanguages: "eng"` metadata-profile filter caught only 7 of roughly 50 foreign-language editions, because OpenLibrary frequently leaves the language field unset — there is nothing for the filter to match on. Related: #1780 notes that box-set pruning only works when Hardcover enrichment is available.

The Hardcover refresh of a comparably prolific author produced **zero** of any of these categories. Hardcover's catalogue is editorially curated, so the work/edition split is already what a library wants: one work per book, English titles, no merchandise rows.

Today the only way to get that is to accept OpenLibrary's catalogue and delete ~80% of the resulting wanted list by hand after every author add or refresh.

### Proposed solution

Accept `hardcover` as a third value for `metadata.primary_provider`, wired exactly like the existing OpenLibrary/DNB swap: the selected provider becomes primary, the other two are added as enrichers. `openlibrary` stays the default, so this is opt-in and changes nothing for existing installs.

`internal/metadata/hardcover` already implements the full `metadata.Provider` surface (it is used as an enricher today, and `providerNameForForeignID` already routes `hc:` foreign IDs to it), so this should be wiring rather than a new client.

One guard is needed: Hardcover authenticates **every** GraphQL query, including search. Selecting it as primary with no API token stored would fail every author and book lookup, not degrade gracefully — so the setting needs to be rejected while no token exists, and the boot path should fall back with a warning if the token disappears out-of-band.

### Alternatives considered

- **Better filtering on the OpenLibrary side.** This is the status quo and it does not work: `allowedLanguages` can only filter records that carry a language field, and OpenLibrary's frequently do not. Title-heuristic filters for "Collection" / "Boxed Set" catch some bundles but are guesswork.
- **Leaving Hardcover as an enricher and using it to prune the OpenLibrary catalogue** (the #1780 direction). Useful independently, but it still starts from the noisy catalogue and tries to subtract from it, rather than starting from a clean one.
- **Per-author provider override.** Finer-grained, much bigger change; the install-wide selector already exists and just needs a third value.

### Additional context

Bindery v1.31.0, Docker deployment.

RustyBower:
## Summary

Lets `metadata.primary_provider` be set to `hardcover`, so Hardcover's curated catalogue can *define* an author's book list instead of only enriching OpenLibrary's. `openlibrary` remains the default, so this is opt-in and changes nothing for existing installs. Closes #2040.

The motivation, from a production instance (full detail in the issue): an OpenLibrary refresh of Clive Cussler produced 217 works and ~120 wanted entries, of which ~97 had to be deleted by hand: German/Dutch/Spanish/Italian/Swedish/Czech/Polish translations, omnibus bundles (`Cussler 3-Copy Boxed Set`), alternate-title duplicates of owned books (`RAISE THE TITANIC`), and non-books (`Dirk Pitt Car Collection`). `allowedLanguages: "eng"` caught only 7 of ~50 foreign editions because OpenLibrary frequently has no language field to filter on. A Hardcover refresh of a comparably prolific author (Brian Jacques, 79 works) produced zero of any of those categories.

## Implementation notes

Mostly wiring. `internal/metadata/hardcover` already implements the full `metadata.Provider` interface (it is an enricher today) and `providerNameForForeignID` already routes `hc:` IDs to it, so no client work was needed.

| File | Change |
|---|---|
| `internal/api/settings_handler.go` | `hardcover` accepted by `validateSettingValue`; provider list extracted to `MetadataPrimaryProviders` + `IsMetadataPrimaryProviderValid`; new `validateSettingDependencies` for the token guard; doc comment updated |
| `cmd/bindery/metadata_providers.go` (new) | `resolveMetadataPrimaryProvider`, which maps the stored value to the provider actually wired, with the fallbacks below |
| `cmd/bindery/main.go` | `hcClient` constructed before the selection; primary chosen by name; every non-primary provider appended as an enricher |
| `web/src/pages/settings/MetadataTab.tsx` | Third `<option>`, disabled (with an explanatory label) until `/system/status` reports `hardcoverTokenConfigured` |
| `docs/`, `README.md`, `en.json` | Provider enumerations and the "Hardcover is an enricher, not a primary" wiki passages updated |

Enricher **order is unchanged** for the two pre-existing selections: `openlibrary` primary still yields `[googlebooks?, hardcover, dnb]`, `dnb` primary still yields `[googlebooks?, hardcover, openlibrary]`.

## Why the missing-token case is handled in three places

Hardcover authenticates *every* GraphQL query, search included, and an unauthenticated request returns `{"error":"Unable to verify token"}`. Unlike Hardcover-as-enricher (which the aggregator skips on error, so a missing token degrades silently but harmlessly), a tokenless Hardcover **primary** fails every author and book lookup. That is a config that cannot work, so the design goal was "never let it be reached", with a loud fallback for the paths that bypass the API:

1. **Reject at save time (primary guard).** `PUT /api/v1/settings/metadata.primary_provider` with `hardcover` and no stored token returns 400 with a message naming the fix ("add one under Settings → API Keys first"). The token is a secret setting invisible to `validateSettingValue`, which is a pure `(key, value)` function reused across the package, so the cross-setting rule lives in a new `validateSettingDependencies` method that has the repo in hand rather than changing that signature.
2. **Reject the inverse (token guard).** Clearing `hardcover.api_token` while Hardcover is primary is refused with a message telling the user to switch provider first. Without this, the "cannot be reached" invariant has an obvious back door, and the running aggregator would keep a Hardcover primary whose token source has gone empty.
3. **Fall back loudly at boot.** `resolveMetadataPrimaryProvider` falls back to OpenLibrary with `slog.Warn` when the value is `hardcover` with no token, or is unrecognised entirely. The API guards can't cover a hand-edited database, a restored backup, or a downgrade, and refusing to *start* over a metadata setting would be a worse failure mode than starting with the historical default.

I chose rejection over silent fallback as the primary mechanism because a fallback alone leaves the UI showing "Hardcover" while OpenLibrary is actually in use, which is the exact silent-breakage shape the issue is about. Restart is still required for the change to take effect, consistent with the existing selector (#1771 tracks making this live-reconfigurable).

- Per-author provider override, and using a Hardcover primary to retro-prune catalogues already ingested from OpenLibrary (overlaps #1780).

## Checklist

- [x] Commits signed off with `git commit -s`. See [Sign your work](../CONTRIBUTING.md#sign-your-work-dco)
- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed: n/a, no env vars or upgrade steps changed
- [x] Added a changelog fragment under `changelog.d/` (not an edit to `CHANGELOG.md`): `changelog.d/2040-hardcover-primary-provider.md`
- [x] Wiki pages updated if user-facing behaviour changed: `docs/User-Guide-Wiki.md`, `docs/Troubleshooting-Wiki.md`, `docs/ARCHITECTURE.md`, `README.md`

Tests added: `TestResolveMetadataPrimaryProvider` (fallback matrix incl. no-token and unknown values), `TestSettings_MetadataPrimaryProviderHardcoverRequiresToken` (both guard directions), `TestSettings_MetadataPrimaryProviderDefaultsToOpenLibrary`, `MetadataTab.primaryProvider.test.tsx` (option disabled without a token, enabled and persisted with one, default and stored values render).

## Test plan

Re-verified on the rebased branch (`0ad0d208`), not the original push:

- [x] `go build ./...`, `go vet ./...`, `gofmt -l cmd internal` (clean)
- [x] `go test ./cmd/... ./internal/...`: passes except `TestRemoveDownload_Rtorrent`, which fails identically on a clean `main` checkout on macOS (`/var` → `/private/var` symlink); re-confirmed against `main` at `810119cf` after the rebase.
- [x] `cd web && npm run build`, `cd web && npm run typecheck` (clean)
- [x] `cd web && npx vitest run`: 63 files, 595 tests, all pass
- [x] `cd web && npm run lint`: 0 errors; the 6 warnings are pre-existing, the same 6 that `main` reports, with one line number in `MetadataTab.tsx` shifted by the added options


vavallee (commented):
Thorough — settings validation, boot-time fallback with a warning rather than dead metadata, refusing to clear the token while Hardcover is primary, tests on both sides, and README, ARCHITECTURE, User-Guide and Troubleshooting all updated in the same PR. That is a higher docs bar than most feature PRs here clear.

Three things before this merges, one of which I think is a real gap.

## What happens to a library that already has an OpenLibrary catalogue?

This is the one I would want answered explicitly. Existing books carry OpenLibrary `foreign_id`s, and author sync reconciles on foreign ID (`GetByForeignID` in `fetchAuthorBooks`, plus the normalized-title dedup below it). After flipping the primary to Hardcover, the same author's works come back with Hardcover IDs.

Does the next sync:

- match them by title dedup and upgrade in place,
- or fail to match and create a second row per book?

`CanonicalDedupKey` gives some protection, but title normalisation across two providers is exactly where translated titles and subtitle differences bite, and the failure mode is a silently doubled library — the same class of complaint as #1816. If it is safe, a test that seeds an OL-sourced library, flips the primary, and asserts no duplicates would be the most valuable test in this PR. If it is not safe, the setting needs a warning at the point of change, not only "restart".

## Restart-only is fine but should be visible in the UI

The wiring is boot-time in `main.go`, so the change takes effect on restart. The wiki text says so; I could not see the same in the settings screen itself. A user who flips it, sees no change, and flips it back is the predictable outcome. One line under the selector is enough. (#1771 tracks making the aggregator live-reconfigurable, so this is temporary.)

## Terms, since this changes what Hardcover data is used for

`docs/third-party-data.md` landed yesterday and records that Hardcover's API rules permit facts about books but exclude other users' ratings and lists from any non-personal deployment. Promoting Hardcover to primary means their catalogue *is* the library, and Bindery re-serves that through its own API and the OPDS feed. Nothing here breaks the rules as I read them — the catalogue is facts, and `average_rating`/`ratings_count` were already the flagged fields — but the doc should mention that primary mode exists, since it materially raises how much of a library is Hardcover-derived. Happy to add that myself once this lands if you would rather not touch it.

Nothing above blocks the code. The duplicate question is the one I would like an answer on before merge.

RustyBower:
Thanks — the duplicate question was the right one to hold this on, and the answer turned out to be more interesting than a yes or no. Short version: **the global primary flip does not duplicate an existing catalogue, and the reason is not `CanonicalDedupKey`.** But you were right that the normalisation is weak, and I have production evidence of it biting. Those are two separate things, and I've handled them separately.

## 1. The flip does not re-source existing authors

The catalogue provider was never chosen by which provider is primary. `primaryAuthorWorks` routes through `providerForForeignID`, which picks the provider named by the **author's own foreign-ID prefix** and only falls back to the primary when nothing matches:

```go
func (a *Aggregator) primaryAuthorWorks(ctx context.Context, authorForeignID string) ([]models.Book, error) {
	provider := a.providerForForeignID(authorForeignID)
```

A bare `OL…` author id resolves to `"openlibrary"`, and `main.go` appends every non-primary provider as an enricher — so with Hardcover primary, OpenLibrary is still in `a.providers()` and still answers for OpenLibrary-linked authors. The works come back with the ids the library already stores, reconcile by foreign id, and the title-dedup fallback is never reached.

So the flip governs which provider **new** authors resolve against. Authors already in the library keep their source until something rewrites `author.ForeignID`.

Tests added:

- `TestPrimaryFlipToHardcover_OpenLibraryAuthorsStayOnOpenLibrary` / `_HardcoverAuthorsUseHardcover` — routing in both directions.
- `TestPrimaryFlipDoesNotChangeAuthorRouting` — the invariant stated directly: for any author id, the same provider answers before and after the flip.
- `TestPrimaryFlipToHardcover_DoesNotDuplicateOpenLibraryLibrary` — the end-to-end one you asked for: seed an OL-sourced library, wire Hardcover primary, sync, assert the row count is unchanged.

One note on that last test. It uses `FetchAuthorBooks`, not `RefreshAuthorBooks`, on purpose: the refresh path gates row *creation* on the author's monitoring policy (#1815), so an unmonitored author cannot grow regardless of what the provider returns and the assertion would pass vacuously. `FetchAuthorBooks` permits creation, which makes it the stronger claim — rows could have been created and none were.

## 2. Where it *is* unsafe, with the evidence

**You should not take the above as "so the normalisation is fine".** It isn't, and the path that *is* exposed to it is the per-author relink endpoint, which rewrites `author.ForeignID` and so legitimately re-sources the catalogue.

On a live 1.31.0 instance I relinked one author (Clive Cussler) from OpenLibrary to Hardcover via `POST /api/v1/author/{id}/relink-upstream` and refreshed. 173 book rows: 84 `calibre`, 75 `hardcover`, 14 `openlibrary`. Normalising by lowercasing and stripping all non-alphanumerics found **2 duplicate pairs**, both punctuation-driven:

```
'Poseidons Arrow'                          (openlibrary, status=wanted)
"Poseidon's Arrow"                         (calibre,     status=imported)

'Journey of the Pharaohs Numa Files #17'   (hardcover,   status=wanted)
'Journey of the Pharaohs: Numa Files #17'  (calibre,     status=imported)
```

The harm is worse than a cosmetic dupe: the duplicate lands as `wanted` beside a row that is already `imported`, so search-wanted re-acquires a book the user owns.

Caveats, stated plainly:

- This was a **per-author relink, not a global primary flip**. I checked whether the two paths are equivalent rather than assuming it, and they are not — that's finding #1 above. Relink is listed in `RefreshAuthorBooks`'s own doc comment as one of the refresh callers, and both end in `fetchAuthorBooks`, but relink additionally rewrites the author's foreign id, which is exactly the step that puts reconciliation on titles.
- **Sample of one author.**
- My normalisation was **more aggressive than `CanonicalDedupKey`**, which is precisely why it surfaced pairs Bindery treats as distinct.
- Note that **neither pair is an OL-row-vs-HC-row pair** — both are provider-vs-Calibre. This is a pre-existing, provider-agnostic defect that today's `openlibrary` default already hits.

### What the normalisation actually does

Confirmed by running it, not by reading:

```
"Poseidon's Arrow"                        -> "poseidon's arrow"
"Poseidons Arrow"                         -> "poseidons arrow"

"Journey of the Pharaohs: Numa Files #17" -> "journey of the pharaohs"
"Journey of the Pharaohs Numa Files #17"  -> "journey of the pharaohs numa files #17"
```

Two distinct mechanisms:

- **Apostrophes survive.** `NormalizeQueryTitle` folds `U+2019` to ASCII `'` but never removes it.
- **`stripSubtitle` is one-sided**, and more aggressive than it looks. It truncates at the first `": "`, so `"Star Wars: A New Hope"` becomes `star wars`, while a provider writing the same subtitle without the colon keeps the whole string. The two forms can never meet.

## 3. Which remedy, and why

I took your option (b) — warning at the point of change plus a follow-up issue — and added the test anyway. Reasoning:

`CanonicalDedupKey` is deliberately the single authority (#940). Changing it changes `books.dedup_key` for every row, `seriesmatch.NormalizeSeriesName` which is built on it, the recommender's `OwnedWorkKeys`, ABS/Calibre cross-source binding, and the `normdrift` invariants. The key change itself is survivable — `backfillBookDedupKeys` re-canonicalizes on the next boot and is documented to exist for exactly that. But:

- The apostrophe half is tractable in isolation and fixes **one of the two** observed pairs.
- The subtitle half is not. `stripSubtitle`'s truncation is load-bearing for series collapse (`import_series_collapse_test.go`, `authors_series_collapse_test.go`), so making the two forms agree means *deciding what the subtitle rule should be*, not extending it.
- `backfillBookDedupKeys` explicitly does not merge rows that collide after a key change, so hardening fixes future syncs and leaves already-duplicated libraries duplicated. A repair pass is a third piece of work.

Half a fix with that blast radius, inside a PR that adds a value to a settings enum, seemed like the wrong trade. Filed as **#2042** with the evidence, the mechanism, and the consumer list.

What I did do is pin it in code rather than prose, so it is measurable and a fix is deliberate:

- `TestCanonicalDedupKey_PunctuationDivergenceIsNotFolded` (`internal/indexer/dedup_test.go`)
- `TestRelinkedAuthorDedupsByTitleExceptPunctuationDivergence` (`internal/api/`) — drives the relink shape end-to-end and asserts the current row layout. The sync log reads `added=2`, which is the production defect reproduced in a unit test.

Both fail with instructions to move the pair into the deduped assertions if a change starts folding them. Happy to do the apostrophe half here instead if you'd rather — it's genuinely small — but I didn't want to land a partial normalisation change under a feature flag PR without you weighing in.

## 4. Restart hint (non-blocking)

The generic `settings.general.restartRequired` line was already under the selector, which is probably why it reads as missing — it is low-contrast and says nothing specific. Replaced it with a dedicated string, new i18n key `metadataProviderRestartHint`, covering both halves:

> Takes effect after the next Bindery restart — the provider is wired at startup, so nothing changes until then. Authors already in your library stay linked to the provider they were added from; the new primary applies to authors you add afterwards. To move an existing author, use "Link metadata" on their page.

The second sentence exists because of finding #1. Without it a user restarts, sees their existing authors unchanged, and still concludes it's broken. Covered by a test asserting both halves render. Referenced #1771 in the code comment as retiring the first half.

## 5. Third-party data terms (non-blocking)

Added a paragraph to `docs/third-party-data.md`. It notes primary mode exists, that it does not move the line the API Rules draw — titles, series, editions, narrators, descriptions and covers are facts either way — but that it raises the proportion of a library that is Hardcover-derived, which is what makes the two existing obligations bite for a professional deployment. Flagged the aggregated-ratings exclusion specifically, since `average_rating` and `ratings_count` arrive on the same records as the facts. No claims beyond what the page already establishes; happy for you to rewrite it.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l cmd internal` clean. `go test ./cmd/... ./internal/...` passes except `TestRemoveDownload_Rtorrent`, which I re-confirmed fails identically on a clean `origin/main` worktree on macOS (`/var` → `/private/var`). Frontend: `npm run build`, `npx tsc -b --force`, `npx vitest run` (62 files, 584 tests), `npm run lint` 0 errors / 6 pre-existing warnings.

vavallee:
Rebased this onto main and pushed the result here, since it had drifted behind a few merges.

Two things needed resolving.

The enricher block in `cmd/bindery/main.go` collided with #2082, which landed while this was open and made the "hardcover enrichment enabled" startup line conditional on a token actually being configured. Kept both: your restructured primary/enricher selection, with the honest log moved inside the `primaryName != "hardcover"` branch. When Hardcover is primary the token is guaranteed by `resolveMetadataPrimaryProvider`, so there is no idle case to report there.

The bigger one: both tests you wrote to pin the punctuation-divergence limitation now fail, because #2043 fixed it. That is the outcome your own comments asked for. The indexer test I dropped, since main's `dedup_subtitle_test.go` already covers both shapes in both directions and more thoroughly; the two pairs of yours it did not cover moved there. The relink test now asserts three rows instead of five and lost the caveat in its name.

Nothing else changed. Full Go suite, tsc and vitest all green locally.

The token handling here is good, in particular refusing the setting from both sides rather than only at boot.

RustyBower:
Thanks for rebasing this rather than bouncing it back.

Both resolutions look right to me.

On `main.go`: agreed there is no idle case to report when Hardcover is primary. `resolveMetadataPrimaryProvider` already falls back to OpenLibrary when the setting says `hardcover` with no token, so the `primaryName != "hardcover"` branch is the only place that log can be reached with a token that might be missing. I also checked the enricher ordering survived the restructure: `openlibrary` primary still yields `[googlebooks?, hardcover, dnb]`, `dnb` primary still yields `[googlebooks?, hardcover, openlibrary]`.

On the tests: dropping the indexer one is the right call. I wrote it against a normalizer that had not been hardened yet, and main's `dedup_subtitle_test.go` covers both shapes in both directions plus the `CompareTitles` decision table, which is strictly more than mine did. Good catch carrying the two extra pairs across. The relink test asserting three rows is exactly what those comments were asking for.

One thing on my side: the Test plan in the description still reports my pre-rebase local run, including the `TestRemoveDownload_Rtorrent` macOS symlink failure. I will refresh it.

On CI: `lint` is red here, but it is red on main too. Run 32203229131 at 810119cf fails with the same six staticcheck SA5011 errors in `cmd/bindery/dbcheck_test.go`, `internal/api/author_refresh_test.go` and `internal/api/authors_test.go`, none of which this branch touches. All six are the `t.Fatal` guard shape that SA5011 misreads.

I installed golangci-lint v2.11.4 and ran it against clean checkouts of both `main` and `9b73ccc9`, the last commit where lint was green, and got `0 issues` from each, including with `GOOS=linux GOARCH=amd64`. The job log shows a cache hit on `golangci-lint.cache-Linux-2954-...`, and SA5011 depends on cached cross-package facts, so a stale cache fits the pattern: green through `9b73ccc9`, red from the next merge onward, on files neither merge touched. Evicting that cache entry or bumping `cache-invalidation-interval` would likely clear it. Worth saying that I cross-compiled from macOS rather than running on a linux runner, so I cannot rule out a platform difference outright.

Issue #1959, opened 2026-08-12
Let ebook and audiobook share one book folder instead of appending ' (2)'


vavallee:
### Problem

When a book's audiobook naming template resolves to a folder the ebook already occupies, Bindery won't merge into it and appends ` (2)` instead — you end up with `Title/` holding the epub and `Title (2)/` holding the m4b. Keeping both formats of one book in a single folder isn't supported today. Requested by Daize on Discord (dual-format thread, 2026-07-27); tools like Storyteller and some ABS layouts want the pair co-located.

### Proposed solution

An opt-in (naming-template or settings level) that lets the ebook and audiobook share one book folder: when the audiobook destination resolves to the book's existing ebook folder, import into it instead of suffixing ` (2)`. Scan and rename need to keep treating the folder as one book with two formats (the scan side mostly works already: the folder-parent tracking only applies to audio files).

### Related

#942 (Storyteller post-import folder drop and pair gating), #1426 (docs on the shared-layout workaround), #1632 (per-format external import routing).

schmitzkr:
Didn't check the roadmap before poking at this. I see it's already planned with the other two dual-format items, so ignore the stray commits above.

One thing worth passing on in case it's useful when you get to it: `flattenAudiobookDirNamed` creates `destDir` itself and unwinds errors with `os.RemoveAll(destDir)`, so pointing the flatten paths at the book's existing folder (the natural way to do this) means a mid-placement failure takes out the folder the ebook is in. Hit that on my first attempt.

I let Claude Code have a go at a fix. Branch at `schmitzkr/bindery`, `fix-1959-shared-format-folder-collision` if it's any use. Only ever run against synthetic temp dirs, not a real library, so treat it accordingly.

vavallee:
Confirmed, and thanks for writing it up instead of just dropping the branch.

The flatten hazard is real. `flattenAudiobookDirNamed` refuses a destination that already exists (`internal/importer/flatten.go:192`), so it cannot bite today, but the moment that guard comes out to allow merging, the three `os.RemoveAll(destDir)` unwinds at flatten.go:204, :212 and :218 are aimed at a folder holding the ebook. `rollbackPlacedFiles` in `internal/importer/scanner_poll.go:1155` is already the right shape for this: it removes only the names the attempt placed, then tries a non recursive remove of the folder itself, which fails harmlessly when the ebook is still sitting in it.

[...]

Issue #1368, opened 2026-07-01
Fix Match / delete can remove a file another book owns (data loss)


vavallee:
## Summary

Deleting a book record can remove an on-disk file that a **different** book legitimately owns, after using **Fix Match** (file reassign, #1238). Reported by a user who lost ~30 audiobook files this way. Shipped in v1.23.0 / v1.23.1.

Two independent defects, plus one amplifier, were confirmed by running the real code against a temp filesystem.

## Bug A — Fix Match detach silently no-ops when the stored path isn't byte-identical to the symlink-resolved path

`ManualImportHandler.prepareImport` (`internal/api/manual_import.go:104`) resolves the request path with `EvalSymlinks` via `LibraryRoots.ResolveContained`, and `Reassign` then detaches with `RemoveBookFile(resolved)` (`manual_import.go:197`). `BookRepo.RemoveBookFile` → `BookFileRepo.DeleteByPath` matches `book_files.path` **exactly** (`internal/db/book_files.go:67`). Libraries mounted through a symlink (very common in Docker/NAS) store the *unresolved* path in `book_files`, so the exact match misses and **the detach does nothing**: the source book keeps its `book_files` row, its legacy `audiobook_file_path`/`file_path`, and status `imported`.

Result: the "old" record still holds live references to a file the reassign was supposed to move away. Deleting that record with `?deleteFiles=true` then removes the file.

Reproduced: symlinked root + unresolved stored path → after `Reassign`, source book still has its `book_files` row and legacy columns; deleting it removes the file on disk.

## Bug B — `removeStaleSource` deletes the source when the target has *any* pre-existing file

`removeStaleSource` (`manual_import.go:217`) is meant to delete the original mis-filed copy only once it confirms the import placed the file elsewhere. It sets `moved=true` if **any** of the target's `book_files` entries exists on disk at a path `!= src`, then `os.Remove(src)`. It never verifies that *this* import added a new file for the requested format. If the target already owned an unrelated file (e.g. an ebook) and this reassign placed nothing (import failed / no-op), the source is deleted anyway.

Reproduced (direct call): target owns an existing ebook `Q`; import registers nothing new; `removeStaleSource(P1, target)` deletes `P1`.

## Amplifier — Delete's legacy-column fallback removes files with no cross-reference check

[...]

PR #2068, opened 2026-08-17
fix(api): scope folder deletes to the requested format (#2052)

### LINKED ISSUE #2052: Deleting audiobook deletes ebook too (if they are in the same folder?)
### Pre-flight

### Current behaviour

I had a wrong audiofile downloaded as an audiobook.
I also had the ebook downloaded:

Image

I then deleted the audiobook from the UI:
Image

The result was:

Image

As you can see up, the ebook file is still referred to from the UI. However, when I click download the result is this:

Image

### Expected behaviour

Only the audiofile should be deleted when I delete it; only the ebook file should be delete when I delete it.

### Steps to reproduce

See above

### Trace logs

```text
No logs referring to this incident.
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v1.31.0

### Environment details

Probably has some relevance to https://github.com/vavallee/bindery/issues/1959

vavallee:
Closes #2052.

## What happened

flaevers deleted an audiobook and lost the ebook that lived in the same folder. The UI kept showing the ebook; downloading it failed.

## Why

`removeBookPathScoped` (`internal/api/books.go`) has two branches. The file branch is carefully format-scoped — `sweepMatchesFormat`, and a comment explaining that it exists so `?format=ebook` cannot destroy the audiobook. The directory branch was:

```go
if info.IsDir() {
    return os.RemoveAll(p)   // format is never read
}
```

Audiobooks are always directories. `internal/importer/scanner.go:1488` registers `destDir` in `book_files` for `MediaTypeAudiobook`, so `DELETE /book/{id}/file?format=audiobook` resolves to a folder and hits that branch. Everything inside went, ebook included.

Only the audiobook's `book_files` row was then deregistered, which is why the ebook stayed visible and 404'd — the DB and the disk disagreed.

This gets worse, not better, with #1959 (share one book folder), which would make the collision the default layout rather than an accident.

## Second defect in the same place

`safeRemoveBookPath` checks `PathOwnedByOtherBook` on the path it is handed and nothing else. With `RemoveAll` on a directory, every file nested under it was unguarded — so a folder delete could unlink a different book's file. That is the exact failure #1368 was written to prevent, reintroduced through a path that guard never covered.

## Fix

`removeBookDirScoped` replaces the `RemoveAll`:

1. Walk the tree, removing book files whose extension matches the requested format. A book file of the other format, or one owned by another book, is left and counted.
2. If nothing was left, clear the sidecars (cover art, `.cue`, `.opf`) and remove the now-empty tree. If something was left, the sidecars stay — the surviving book still wants its artwork.

The ownership check becomes a predicate built in `safeRemoveBookPath` and threaded through the sweep, so it applies per file rather than only to the tracked path. The same-stem sibling sweep in the file branch now consults it too.

`format == ""` (whole-book delete, author delete) still clears everything, with the ownership guard now covering children.

## Tests

Three new guards in `internal/api/books_test.go`, each verified to fail against the old `os.RemoveAll`:

- `TestBookDeleteFile_AudiobookFolderKeepsEbookInside` — the reported case, including the assertion that the surviving `book_files` row points at a file that exists
- `TestBookDeleteFile_AudiobookFolderRemovedWhenAloneInIt` — scoping must not strand the folder when the audiobook was the only thing in it (multi-disc subfolder, cover, cue)
- `TestBookDeleteFile_AudiobookFolderKeepsAnotherBooksFile` — the #1368 guard applied to children

`internal/api` and `internal/importer` suites pass; `golangci-lint` clean.

### DIFF
--- a/changelog.d/2052-format-scoped-folder-delete.md
+++ b/changelog.d/2052-format-scoped-folder-delete.md
@@ -1,0 +1,18 @@
+### Fixed
+
+- **Deleting one format no longer destroys the other format's file when both
+  live in the same folder** (#2052) — audiobook imports register the destination
+  *folder* in `book_files`, so a `?format=audiobook` delete lands on a directory.
+  That branch was an unconditional `os.RemoveAll`, which discarded the format
+  filter every other part of the delete path honours, and took the ebook sitting
+  beside the audio files with it. The ebook's `book_files` row survived, so the
+  book kept advertising a file that was no longer on disk and downloading it
+  returned an error. The directory branch now walks the folder and removes only
+  the files belonging to the format being deleted; cover art and other sidecars
+  are removed once no book file of any format is left, and the folder itself
+  only when nothing remains in it.
+- **A folder delete can no longer unlink a file another book still tracks**
+  (#1368, surfaced by #2052) — the `book_files` ownership guard was applied only
+  to the tracked path handed to the delete, so any file nested under a deleted
+  folder was unguarded. The same check now runs for every file the sweep
+  reaches, including same-stem siblings.
--- a/internal/api/books.go
+++ b/internal/api/books.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/fs"
 	"log/slog"
 	"net/http"
 	"os"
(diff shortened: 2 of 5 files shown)


vavallee:
On the per-file ownership query: leaving it as is. It is a local SQLite lookup on an indexed unique column, so a 200 track audiobook is a few milliseconds total at delete time, and the batched variant would need a LIKE prefix scan plus a fallback path for the case it does hit. Not worth the second code path for a rare operation. Noting it here rather than filing an issue.

PR #2093, opened 2026-08-19
Revert "feat(metadata): allow hardcover as the primary metadata provider (#2040)"

vavallee:
Holds #2041 back from v1.32.0 and drops its CHANGELOG entry.

The feature itself is in good shape: the setting is refused from both directions without a token, and a token removed out of band falls back to OpenLibrary with a warning rather than failing every lookup. The reason for holding it is release composition, not quality. It changes what an author's catalogue *is* for anyone who opts in, and v1.32.0 already turns on five metadata profile filters that were previously inert. Shipping both at once would make a catalogue that comes out wrong hard to attribute to either.

Two things are deliberately kept rather than reverted:

- The conditional Hardcover startup log in `cmd/bindery/main.go` belongs to #2082. It only shows up in this diff because I merged the two by hand during the rebase.
- The two extra `CanonicalDedupKey` pairs in `dedup_subtitle_test.go` were added while rebasing #2041, but they cover main's own #2043 behaviour and stand alone.

Go build, vet, gofmt, the touched Go packages, `tsc` and 590 vitest tests all pass locally.

Re-land after v1.32.0 with a fresh changelog fragment. cc @RustyBower — nothing wrong with the work, purely about not stacking two catalogue-shaping changes in one release.


ianepreston:
Any appetite for re-landing this now that 1.32.x has settled? The revert note said "Re-land after v1.32.0 with a fresh changelog fragment", and v1.32.0 went out on 2026-08-19 with v1.32.2 following on 2026-08-25, so the attribution concern that motivated holding it seems to have passed — the five newly-live profile filters have had a full release cycle of their own to be judged on.

Adding a data point from a fresh install, because #2040's problem statement turned out to describe my library almost line for line.

James S. A. Corey, OpenLibrary primary with Hardcover configured as enricher, metadata profile set as tightly as the current knobs allow — `allowedLanguages: eng`, `unknownLanguageBehavior: fail`, `skipPartBooks: true`, `minPopularity: 5`. That still produced a catalogue carrying French, Spanish, German, Portuguese, Polish and Turkish editions of the Expanse novels as separate works, three different "Expanse Boxed Set" rows, a "Leviathan Falls - Carton of 10 Signed Copies", and alternate-title duplicates ("Vital Abyss" beside "The Vital Abyss", "Churn" beside "The Churn").

The language filter in particular behaved exactly as #2040 predicted: the foreign editions were not tagged with a foreign language, they were tagged `eng`. There was nothing for `allowedLanguages` to match on. Flipping `unknownLanguageBehavior` to `fail` cleared a genuinely large amount of junk (one author went 167 books → 119, another 49 → 27), but it cannot touch a record that actively misreports its language, and it also discards legitimately English works that OpenLibrary left untagged.

Hardcover's catalogue for the same author is nine novels and the novellas, in order, with no merchandise rows — which is what #2040 was asking for.

Not asking to jump any queues, just checking whether it's still on the list or whether something else surfaced that changed the calculus. Happy to test a branch against a real OpenLibrary-grown library if that's useful.

Separately, and independent of this: while working out why series mode was doing nothing on that same author I filed #2207 — Hardcover's series refs are dropped by `mergeAuthorWorkMetadata` whenever a supplemental work merges onto an OpenLibrary one, so #2121 only ever reaches works OpenLibrary doesn't have. That one's worth fixing whichever way this lands, since it affects every OpenLibrary-primary install using series mode.

Issue #2075, opened 2026-08-18
Bulk CSV author import queries unconfigured Hardcover provider and overwhelms metadata APIs


TheOpsMgr:
# Suggested issue title

`Bulk CSV author import queries unconfigured Hardcover provider and overwhelms metadata APIs`

## Is there an existing issue for this?

- [x] I have searched the existing issues.

## Current behaviour

On a clean Bindery configuration, importing a small CSV containing 19 authors causes Bindery to:

1. Query Hardcover even though no Hardcover API token has been configured.
2. Continue sending Hardcover author and per-book enrichment requests after receiving `401 Unable to verify token` responses.
3. Continue sending requests after receiving `429 Throttled` responses, without an effective cooldown/backoff.
4. Generate enough OpenLibrary requests to receive `429 Too Many Requests`, followed by large numbers of timeouts and connection refusals.
5. Interpret the UTF-8 BOM-prefixed CSV header as an author name and search OpenLibrary for `%EF%BB%BFname`.

The resulting partial and inconsistent provider responses also appear capable of resolving authors against different providers, which can leave duplicate author records. That may be a related reconciliation problem rather than the same root cause.

## Expected behaviour

- Hardcover author search and enrichment should be disabled completely unless a non-empty, validated Hardcover token is configured.
- `401` responses should disable that provider for the remainder of the operation (or until its configuration changes), rather than retrying it for every author and book.
- `429` responses should honour `Retry-After` when present and otherwise use bounded exponential backoff with jitter.
- Bulk imports should have a provider-level global rate limit and bounded concurrency, not only per-loop concurrency limits.
- One unavailable or throttled provider should not cause a request storm or inconsistent author identities.
- CSV input should accept UTF-8 with BOM and recognise the `name,monitored` header correctly.

## Steps to reproduce

1. Start with a new/clean Bindery `/config` directory.
2. Complete first-run setup.
3. Leave the Hardcover API token unconfigured.
4. Configure OpenLibrary as the primary metadata source.
5. Go to **Settings → Import / Migrate → CSV of author names**.
6. Upload the attached `bindery-authors-clean-20.csv`.
   - Despite the filename, it contains one header plus 19 author records.
   - Every record has `monitored=FALSE`.
   - The file is UTF-8 with a BOM.
7. Observe the debug log while the import resolves authors and populates their catalogues.

## Reproduction result

From the attached one-hour debug export (`bindery-logs-20260818T090354Z.txt`):

[...]

PR #2101, opened 2026-08-20
fix: back off OpenLibrary retries and bound bulk-import concurrency (#2075)

gchahcg:
## Summary

A bulk CSV or Readarr-DB author import fired one goroutine per newly-added author with no cap at all, and the OpenLibrary client had no retry/backoff logic whatsoever — so a 19-author import opened 19 simultaneous, un-throttled catalogue fetches, and any 429 from OpenLibrary just failed immediately instead of backing off. #2075's reproduction showed exactly that: 429s escalating into timeouts and connection refusals with nothing to slow the storm down.

vavallee's own v1.32.0 already shipped the Hardcover-without-a-token and BOM-header fixes from that issue. This covers the remaining piece — confirmed still broken by the reporter after testing v1.32.0 — the OpenLibrary rate-limiting itself.

Refs #2075.

## Implementation notes

Two independent fixes:

**`internal/metadata/openlibrary/client.go`** — `getJSON` (the single chokepoint every OpenLibrary call goes through) now retries a 429, 502, 503, 504, or transient transport error up to 3 times:
- Honors the response's `Retry-After` header when present (seconds or HTTP-date, capped at 30s defensively).
- Otherwise backs off exponentially from 500ms, capped at 8s, with equal jitter so a burst of simultaneous failures doesn't retry in lockstep and produce a second burst.
- A canceled operation, or one where the *caller's* own context has expired (as opposed to just this request's own timeout), is never retried — respects the caller's budget instead of retrying into a context that's already gone.
- The existing `errors.Is(err, context.Canceled)`/`DeadlineExceeded` chain-preservation contract (covered by pre-existing tests) is unchanged.

**`internal/migrate/csv.go` and `readarr.go`** — the per-author catalogue-fetch fan-out (`go onCatalogueFetch(full)` / `go onSearchOnAdd(full)`, previously one unbounded goroutine per author) is now bounded to 2 in flight and paced 3 seconds apart, via the existing `concurrency.RunBoundedPaced` helper — the same primitive and cadence already used for indexer-search fan-out elsewhere (`authorAutoSearchConcurrency`, `searchPaceInterval`). **This is the fix that actually keeps a large import from tripping the throttle** — the retry/backoff above is a safety net for an occasional hiccup, not a substitute for it.

The bound is 2, not the 4 used for other provider fan-out: each author's own catalogue fetch already internally fans out up to `openlibrary.authorWorkSampleConcurrency` (4) concurrent edition-sample requests, so the two stack — 4 outer would mean up to 16 concurrent OpenLibrary requests, not 4. 2 keeps the realistic worst case (up to 8) meaningfully lower.

Both call sites now share one dispatch helper, `dispatchCatalogueFetch` (in `csv.go`), rather than two near-duplicated inline blocks.

The fan-out runs under `context.WithoutCancel` — not because the old fire-and-forget goroutines were ever ctx-sensitive (the callback, ultimately `AuthorHandler.FetchAuthorBooks`, takes no ctx and never did), but because `RunBoundedPaced` itself reads ctx to gate its own dispatch loop. Wiring the raw request ctx straight in would mean `net/http` cancelling it — which happens shortly after this function returns and the handler writes its response — silently stops the loop from launching the rest of a large batch partway through.

## Self-review pass

Ran `/code-review high` against this diff before asking for another look. It caught two real issues, both fixed:

- **The concurrency-stacking issue described above** (outer fan-out × inner edition-sample fan-out = up to 16 concurrent requests, not 4) — lowered the outer bound to 2.
- **A regression in the Readarr path**: `importReadarrAuthors` returned early on `rows.Err()` *before* reaching the fan-out dispatch, so a mid-cursor error (e.g. a locked/corrupt source database partway through a large import) would silently skip catalogue fetches for every author already successfully created before that point. The original per-row `go onSearchOnAdd(full)` never had this problem, since it fired immediately as each row committed. Fixed by capturing `rows.Err()` and dispatching the fan-out unconditionally on `newlyAdded` before checking it — now centralized in `dispatchCatalogueFetch` so the CSV and Readarr paths can't drift on this again.

Also flagged but consciously left out of scope: no overall wall-clock cap on `getJSON`'s retry loop (bounded indirectly by the caller's own ctx, worst case ~150s); the outer pacing doesn't react to a `Retry-After` value from an in-flight request; the Hardcover client has the same fail-immediately shape with no retry added; and this is now a third independent retry-with-backoff implementation in the codebase (alongside `internal/abs` and `internal/calibre`) that could eventually be consolidated. None of these regress anything this PR fixes — they're pre-existing or genuinely separate scope, called out here for visibility rather than silently deferred.

- The reporter also flagged that when OpenLibrary fails and DNB fallback can't match an author by OL key, duplicate author rows get created. That looks like a separate author-identity reconciliation problem rather than the rate-limiting root cause, so it's left out of scope here.
- Pacing cost on a very large Readarr import: 500 authors at 3s apart is 25+ minutes of backfill with no progress surface (the fan-out is detached, and the handler has already responded by then). 3s is fine for a CSV import's realistic size. Needs its own decision — a shorter interval for the Readarr path specifically, or actual progress reporting — rather than folding into this PR. Tracked in #2136.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (no env vars, settings, or user-facing config changed — behavior-only fix)
- [ ] Wiki pages updated if user-facing behaviour changed (not applicable — no UI or settings change)

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l .`
- [x] `golangci-lint run ./...`
- [x] `govulncheck ./...`
- [x] `go test ./internal/migrate/... ./internal/metadata/openlibrary/... ./internal/api/...`
- [x] `go test -race ./internal/migrate/... ./internal/metadata/openlibrary/...`


vavallee (commented):
Good PR, I want this in. Four things before it goes in.

**Sign off is failing.** Commit 3816128a has no `Signed-off-by:` trailer, so the DCO check red flags it. `git commit --amend -s` and force push clears it. Everything else is green.

**1. The second bug in your summary didn't exist.** You describe the fan out as using the caller's ctx and getting cut short when the handler returns. The old code was `go onCatalogueFetch(full)` with no ctx at all, and the callback is `func(a *models.Author) { authorHandler.FetchAuthorBooks(a, false, "") }` in `cmd/bindery/main.go`, which takes no ctx either. The request context never reached the fetch, so nothing was being truncated.

The `context.WithoutCancel` is still correct and still necessary, just for a different reason: `RunBoundedPaced` uses ctx for its pace and concurrency gates, so passing the request ctx straight in would abort your own fan out the moment the handler returns. Worth rewording, otherwise it lands in the changelog as a bug that was never happening.

**2. Pacing cost on large Readarr imports.** `RunBoundedPaced` starts one item per `minInterval`, so 3s per author. A CSV of 19 is a minute, fine. A Readarr database with 500 authors is 25 minutes of backfill, 2000 is over an hour, and there's no progress surface anywhere because the fan out is detached and the handler has already responded. I'm fine with 3s for CSV. Tell me what you think for the Readarr path, either a shorter interval there or something that reports progress.

**3. Retries can't reuse the connection.** On the retryable status path you read at most 512 bytes and then close, so Go won't reuse that connection and every retry opens a fresh TCP connection to OpenLibrary. That's leaning into the exact connection refused symptom #2075 reported. Draining the body before close fixes it, one line.

**4. Backoff is smaller than it reads.** 500ms, 1s, 2s with equal jitter comes to between 1.75s and 3.5s total across the three retries. The concurrency cap is doing the real work here and that's fine, I just don't want the release notes claiming the retry alone rides out an OpenLibrary throttle.

Retry structure itself looks right to me. Honoring Retry-After, capping it, not retrying once the caller's context is gone, resetting `retryAfter` on the transport error path, all correct.

Fix 1 and the sign off and I'll merge. 2 and 3 I'm happy to take as a follow up if you'd rather not expand this one.

gchahcg:
All fixed except item 2, which I've left as a follow-up per your own offer:

- **Sign-off**: amended with `-s`, force-pushed.
- **1**: You're right, reworded — the old code was never ctx-sensitive (the callback takes no ctx at all). `context.WithoutCancel` is there because `RunBoundedPaced` itself reads ctx for its own dispatch loop, not because anything old was being cut short. Fixed the commit message, PR description, and the in-code comments to say that accurately.
- **3**: Fixed — drains the body (bounded to 1 MiB) before closing on the retryable-status path instead of closing after the 512-byte error-message read, so the connection can be reused. Added a test asserting the body is read to EOF before Close.
- **4**: Reworded the changelog fragment and PR body to be explicit that the concurrency/pacing fix is what actually prevents tripping the throttle, and retry/backoff is a safety net for an occasional hiccup, not a throttle-survival mechanism.

**2 (Readarr pacing)**: leaving this as a follow-up rather than expanding this PR, since I think it's a real design decision (shorter interval vs. progress reporting) rather than something to bolt on here. My instinct is a shorter interval for the Readarr path specifically — a Readarr migration is a one-time, unattended operation where "eventually consistent, no UI to watch" is a reasonable tradeoff, unlike a CSV import someone's actively waiting on. But happy to hear if you'd rather see it done properly with progress surfaced instead. I'll open an issue for it either way.

Pushed the amended commit (749d53f0) — sign-off should be green now.

gchahcg:
Opened #2136 to track the Readarr pacing/progress follow-up.

gchahcg:
Ran a self-review (`/code-review`) on this before waiting for another look from you — caught a real regression I'd introduced in the Readarr path (early return on `rows.Err()` skipped the catalogue-fetch dispatch for already-created authors) and a concurrency-stacking issue (outer fan-out × the pre-existing inner edition-sample fan-out meant up to 16 concurrent requests, not 4). Both fixed, PR description updated with details, amended commit pushed (3c4aae75).

Issue #1347, opened 2026-06-29
Authors list still out of order for names starting with accented/non-ASCII letters (#1312 follow-up)


vavallee:
### Summary

Reported by a user on v1.23.0: the Authors list is still out of alphabetical order despite the A–Z fix in #1312. #1312 added `COLLATE NOCASE` to the author `ORDER BY`, which fixes ASCII case interleaving but does **no Unicode/accent folding** — SQLite's built-in `NOCASE` collation only folds `A–Z`.

### Root cause

`internal/db/authors.go` orders authors by `sort_name COLLATE NOCASE` (`authorSortOrder`, lines ~132–141, and the `listAuthorsAll`/`listAuthorsByUser` consts). Any `sort_name` whose **first** character is a non-ASCII letter (Ö, Á, Ç, É, Ł, Ø, Æ, …) has a code point > `'z'` (0x7A), so it sorts after every ASCII name — landing in a clump at the end of A–Z (and the front of Z–A). Scandinavian / Spanish / Turkish / Polish / German author names read as "out of order."

The #1312 regression test (`internal/db/list_filtered_test.go`) never exercises a *leading*-diacritic case — its diacritics ("Honoré", "Émile") sit after the comma where they don't affect the comparison — so the gap shipped untested.

A secondary contributor: `sort_name` is populated inconsistently across import sources (some derive "Surname, Given", some store the upstream value verbatim), which can interleave even pure-ASCII names.

### Fix direction

Store an accent-folded, lowercased `sort_key` derived from `sort_name` (fold via `golang.org/x/text/unicode/norm` NFD + strip combining marks, plus explicit map for non-decomposable letters: ø→o, ł→l, æ→ae, ß→ss…), index it with a plain BINARY index, and `ORDER BY sort_key`. Compute on every author write + one-time startup backfill (mirrors `backfillBookDedupKeys`).

Follow-up to #1312.

PR #2103, opened 2026-08-20
feat(authors): display-name sort option for the Authors list (#2102)

### LINKED ISSUE #2102: Option to sort the Authors list by display name instead of sort_name
**Request**

The Authors list is ordered by `sort_name` ("Weir, Andy"), so it always reads surname-first. Some libraries — mine included — are browsed by display name, and there is currently no way to get "Andy Weir" ordering: the A–Z/Z–A toggle from #1349 / #1906 changes direction, not the key.

**Proposal**

A preference (per user, or simply a third state on the existing Name column header) to sort by `authors.name` instead of `authors.sort_name`. Both values already exist on every author row, so this is surfacing a choice, not new data. The accent-folding from #1350 should apply to whichever key is active.

**Context**

- `name` = "Andy Weir", `sort_name` = "Weir, Andy" — both populated today
- Applies to the Authors page list; OPDS and other consumers could keep sort_name
- Happy to contribute the patch if the approach sounds acceptable — I have a live instance to test against

paulohenriquesg:
Closes #2102.

Adds `name-az` / `name-za` sort keys to the Authors list, ordering by a new `authors.name_sort_key` column — the display-name sibling of `sort_key` (migration 058), folded by the same `authorSortKey` transformer so accented first names sort in place (same treatment #1347 gave the surname key).

**Changes**
- Migration 079: `name_sort_key` column + index; ships empty, populated by the startup backfill
- `backfillAuthorSortKeys` extended to maintain both keys (re-canonicalizes on folder changes, as before)
- All three author write paths set the key alongside `sort_key`
- `authorSortOrder` whitelists the two new keys, `name COLLATE NOCASE` as tiebreaker
- Authors toolbar: **First name A–Z / Z–A** buttons beside the existing A–Z/Z–A; `SortMode` extended; en locale keys added (other locales fall back)

**Testing**
- `TestListPageFiltered_NameSort`: red-green verified (the red phase caught that a 3-author fixture made the `name-za` expectation coincide with the unknown-key fallback — a fourth author breaks the symmetry so both keys prove the ORDER BY is actually applied)
- Full `./internal/db` suite green; full repo suite green except `TestRemoveDownload_Rtorrent`, which fails identically on clean main under macOS (`/var/folders` symlink trips the symlinked-parent deletion guard) — unrelated
- Migration + backfill exercised against a copy of a real 335-author library in Docker: `applied migration version=79`, `backfilled author sort keys rows=335`, zero empty keys, and the three orderings verified against the live data


vavallee:
Heads up before this goes further, and it is not about the change itself.

This PR adds `internal/db/migrations/076_authors_name_sort_key.sql`, and there is another open PR (#2123) adding `076_orphan_owner_sweep.sql`. Two files sharing a numeric prefix is a hard failure rather than a warning:

```go
return fmt.Errorf("duplicate migration version %d: %q and %q share the same numeric prefix; rename one before booting", v, existing, entry.Name())
```

`assertUniqueMigrationVersions` in `internal/db/db.go` runs before any migration is applied, so whichever of the two lands second takes every database test down with it and Bindery refuses to boot until the file is renamed.

Could you rename yours to `079_authors_name_sort_key.sql`? 076, 077 and 078 are all claimed by work in flight. Nothing else needs to change, the runner keys on each version number individually rather than on a high water mark, so a later number is not a problem and the order these land in does not matter.

Worth knowing why I am raising it now rather than letting CI find it: it will find it, because `db.OpenMemory` runs migrations and so nearly every database test would fail at once. But that only happens after the other PR merges, and it would land on you as a wall of unrelated looking failures in code you never touched. Easier to move the number now.

Sorry for the friction, this is a collision between two things in flight rather than anything wrong with your PR. I have not looked at the sort change itself yet.

paulohenriquesg:
Renamed to `079_authors_name_sort_key.sql` in bee77c3 — comment references updated to match, `./internal/db` suite green after the move. Thanks for catching it before the collision landed as mystery failures; appreciated.

Issue #1404, opened 2026-07-04
NZBFinder dlz fail via Prowlarr integration with HTTP 400 / error 203, manual Prowlarr dlz OK


b8z-io:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Bindery sha-973e019
Prowlarr 2.4.0.5397
Downloader: SABnzbd

Prowlarr integration syncs indexers successfully.
Search against NZBFinder works.
Manual download of the same NZBFinder release from Prowlarr succeeds and sends to SABnzbd.

Bindery download fails:

failed to send download client_type=sabnzbd error="fetch nzb: indexer returned HTTP 400:
<error code=\"203\" description=\"This application is not allowed to download NZBs from NZBFinder.\"/>"

Release:
Denis.Dutton.The.Art.Instinct.Beauty.Pleasure.and.Human.Evolution.2009.RETAiL.EPUB.eBook-NODE

Expected: Grabbing the release from Bindery should succeed in the same way that grabbing the identical release directly from Prowlarr succeeds.

Actual:
Bindery appears to fetch the NZB itself and receives NZBFinder app restriction 203.

Additional observations

- The same release can be searched and downloaded successfully from Prowlarr.
- Prowlarr sends the NZB to SABnzbd without issue.
- Disabling NZBFinder in Bindery removes the HTTP 400 error.
- Other indexers continue to function normally.
- Bindery indexers were synced automatically from Prowlarr.

### Expected behaviour

Expected: Grabbing the release from Bindery should succeed in the same way that grabbing the identical release directly from Prowlarr succeeds.

### Steps to reproduce

Title added to Wanted list
Click search
Title indexes correctly with Prowlarr

<img width="960" height="121" alt="Image" src="https://github.com/user-attachments/assets/717e9caa-941b-40c1-a85b-c03a9441cf21" />

Click grab

Get error:

<img width="1032" height="120" alt="Image" src="https://github.com/user-attachments/assets/b78bb41c-d57f-42d5-ab86-052d89a9e614" />

### Trace logs

```text
07/04, 12:12:37	ERROR		failed to send download	client_type=sabnzbd error="fetch nzb: indexer returned HTTP 400: <?xml version=\"1.0\" encoding=\"UTF-8\"?>
[...]

PR #2107, opened 2026-08-20
fix(downloader): reject a non-NZB indexer body before the download client (#2105)

### LINKED ISSUE #2105: Indexer returning a non-NZB body on HTTP 200 is reported as "SABnzbd rejected download"
### Pre-flight

- [x] I'm running the latest version (v1.32.0)

### Current behaviour

When an indexer answers an NZB download with **HTTP 200 and a body that isn't an NZB**, Bindery uploads that body to SABnzbd verbatim. SAB refuses to parse it, `addfile` returns `status:false`, and `AddURL` reports:

```
SABnzbd rejected download
```

The message names the one component in the chain that behaved correctly. SAB's own log says what actually happened:

```
[nzbparser:96] Attempting to add <release>.nzb
[filesystem:1249] Saving /incomplete-downloads/<release>/__ADMIN__/<release>.nzb.gz
[object:412] Invalid NZB file <release>.nzb, skipping (error: syntax error: line 4, column 0)
xml.etree.ElementTree.ParseError: syntax error: line 4, column 0
```

…but that log is on the *other* side of the integration, and SAB purges the file before `backup_nzb`, so by the time anyone looks the bytes are gone. From Bindery's UI the book simply drops back to **Wanted**.

There is a second, related gap on the same path. `fetchNZBContent` ends with:

```go
return io.ReadAll(io.LimitReader(resp.Body, 50<<20)) // 50 MB cap
```

The `io.ReadAll` error is returned **unwrapped**, so a body-read failure surfaces with no indication of what was being read:

```
failed to send to downloader: context deadline exceeded (Client.Timeout or context cancellation while reading body)
```

The non-2xx branch right above it goes through `nzbfetch.Error` and does say `fetch nzb from indexer: …`, so the two failure modes on adjacent lines report with wildly different fidelity.

### Expected behaviour

1. A 200 response whose body isn't an NZB is rejected **by Bindery**, before the download client is involved, with an error that names the indexer, says the response wasn't an NZB, and includes a snippet — the same treatment `nzbfetch.Error` already gives non-2xx responses (including the cross-host-redirect note, which applies directly here).
2. A body-read failure keeps the `fetch nzb from indexer:` prefix that the sibling error path already sets.

### Steps to reproduce

1. Configure a Newznab indexer through Prowlarr whose per-indexer **Redirect** is on (the default for Usenet), so `GET /{indexerId}/download` 301s to the indexer's own file host and Bindery fetches from it directly.
2. Grab a release while the indexer is returning an error page instead of the NZB — in my case intermittently, on roughly half of attempts.
3. The grab fails with `SABnzbd rejected download` and the book returns to Wanted.

To reproduce the reporting bug deterministically without a flaky indexer, point an indexer's download URL at any host that serves `200 text/plain`.

### Trace logs

```
{"level":"INFO","msg":"auto-grabbing book","book":"…","format":"audiobook","result":"…","indexer":"…","protocol":"usenet","client":"SABnzbd","size":510962547}
{"level":"ERROR","msg":"SearchAndGrabBook: failed to send to downloader","client":"sabnzbd","title":"…","error":"SABnzbd rejected download"}
```

and, for the unwrapped variant:

```
{"level":"ERROR","msg":"failed to send download","client_type":"sabnzbd","error":"context deadline exceeded (Client.Timeout or context cancellation while reading body)","title":"…"}
```

### What I verified

- The release's NZB is valid: fetched it by hand (following Prowlarr's 301, with the apikey) — 140,851 bytes, well-formed, root `<nzb xmlns="http://www.newzbin.com/DTD/2003/nzb">`.
- SAB accepts that NZB through the **identical** `mode=addfile` multipart call Bindery makes, and downloads it. So neither the NZB, the category, nor the upload path is at fault.
- The failure is not a truncated NZB. Truncating the real file at every offset yields `no element found` or `unclosed token` from expat — never `syntax error`. `syntax error: line 4, column 0` requires three prolog-legal lines (whitespace / comments / an XML declaration) followed by plain non-markup text on line 4, i.e. an error document.
- 68 of 70 grabs in the same install succeeded, across the same indexers and the same download client, so this is a bad-response path rather than a misconfiguration.

### Deployment method

Helm / Kubernetes

### Bindery version

v1.32.0

### Environment details

microk8s single node, SABnzbd 4.x, Prowlarr with per-indexer Redirect enabled (`grabMethod: Redirect` in Prowlarr's history), Newznab indexer over HTTPS. I've left the indexer unnamed since the mechanism doesn't depend on it — happy to share privately if it helps.

### Notes

Affects `internal/downloader/sabnzbd/client.go` and, with the same shape, `internal/downloader/nzbget/client.go` — both fetch the NZB themselves and forward the bytes.

Not covered here, and worth separate consideration: a single transient fetch failure permanently fails the grab with no retry, which is what leaves the book silently back on Wanted.

PR follows.

helios57:
## Summary

An indexer that answers a refused, expired or rate-limited grab with **HTTP 200 and an error page** had that page forwarded to SABnzbd / NZBGet, which refused to parse it — so the grab failed as `SABnzbd rejected download`, naming the one component in the chain that behaved correctly, and the book dropped silently back to Wanted. `nzbfetch.ValidateNZB` now rejects a non-NZB body at the fetch step and reports what the indexer actually sent. Closes #2105.

## Implementation notes

A 4xx from the same indexer was already reported properly through `nzbfetch.Error`; only the 200-with-junk route was blind, and it is the route a quota-exceeded or expired-token grab takes.

`ValidateNZB` decides on the **root element alone**. An error page parses far enough to yield a root, so "does it parse" is not the question; and reading past the root would mean walking a 50 MB document to learn what the first token already settled.

Everything else is deliberately lenient — non-strict decoder, pass-through `CharsetReader`, BOM trimmed — because rejecting a sloppy but genuine NZB turns a working grab into a failed one, which is worse than the bug being fixed. No leniency lets a non-NZB through: the root element name is what decides, and it is ASCII under every encoding an indexer plausibly uses.

| Body | Before | After |
|---|---|---|
| Plain-text error page, HTTP 200 | `SABnzbd rejected download` | `fetch nzb: the indexer returned HTTP 200 with a body that is not an NZB: <snippet>` |
| newznab `<error/>` doc, HTTP 200 | `SABnzbd rejected download` | `fetch nzb: indexer refused the download (HTTP 200, newznab error 203: …)` |
| Same, after a Prowlarr redirect | as above | as above **+** the #1404 cross-host hop guidance |
| Body read fails part-way | `context deadline exceeded (… while reading body)` | `fetch nzb from indexer: <cause>` |

The rejection reuses the diagnostics already in the package, so a newznab error served with 200 keeps its code and description, and the #1404 redirect note applies unchanged — the Prowlarr redirect is what puts Bindery in front of the indexer to begin with.

The second half of the fix is one line: `fetchNZBContent` ended with `return io.ReadAll(...)`, returning the read error **unwrapped**, so a failure part-way through the body arrived with no indication of what was being read — one line below a branch that already said `fetch nzb from indexer`.

- Both usenet clients get the same three lines; leaving NZBGet with the identical bug seemed worse than the symmetry costing a few lines.
- **No retry.** A single transient fetch failure still fails the grab permanently, which is what leaves the book quietly back on Wanted. That is a behaviour change rather than a reporting fix, so it is called out in #2105 as a follow-up rather than bundled here.
- `TestPendingGrab_DispatchesDownloadAndClearsPending` used one `httptest` server as both indexer and SAB, so the "NZB" it fetched was SAB's own JSON response — it only passed because nothing validated the body. It now serves a real NZB on the `.nzb` path; what the test asserts is unchanged.

- Retry a transient NZB fetch before failing the grab (#2105).
- `httpsec.NewDialContext` dials `ips[0]` only, with no fallback to the other resolved addresses. On a dual-stack indexer host reached from an IPv4-only pod network that is a hard failure whenever the resolver orders AAAA first. Latent rather than the cause here — happy to open a separate issue if it's wanted.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed — n/a, none changed
- [x] Wiki pages updated — new section in `docs/Troubleshooting-Wiki.md`, alongside the existing error-203 entry

## Test plan

- [x] `go test ./cmd/... ./internal/...` — 54 packages, all pass
- [x] `gofmt -l .` clean, `go vet ./...` clean
- [x] `golangci-lint run --timeout=5m` (v2.11.4) — 0 issues
- [x] `govulncheck ./...` — no vulnerabilities affecting the code
- [ ] `cd web && npm run build` — not run; the change is backend-only and touches no frontend file

Each new test was watched failing against the unfixed code first: the client-level ones failed with the download client being called and, for the read path, with a bare `unexpected EOF`.

Found while debugging a real grab failure on a self-hosted install: the release's NZB was valid and SAB accepted the identical `mode=addfile` upload when handed the real bytes, so the failure was entirely in what the indexer returned. Details and the ruling-out of a truncated NZB are in #2105.


vavallee (commented):
This is a good report and a good fix. I verified both halves against v1.32.0 before reading the patch: `sabnzbd/client.go` and `nzbget/client.go` both ended on a bare `return io.ReadAll(...)` one line below a branch that already routes through `nzbfetch.Error`, and nothing on the 200 path looked at the body at all. Your ruling out of a truncated NZB was more work than I'd have asked for, thank you.

The validation design is the part I care most about and I think you got it right. Root element only, non strict decoder, charset passthrough, BOM trimmed. A hard rejection here turning a working grab into a failed one really is worse than the bug, and you clearly weighed that.

One question before I merge.

**Gzip bodies.** `rootIsNZB` decodes the raw bytes. Transport level gzip is fine, Go decompresses that for us since we don't set Accept-Encoding ourselves. What I'm unsure about is an indexer serving the NZB as a gzip file, `Content-Type: application/x-gzip` with no Content-Encoding header, which some newznab implementations do. Those bytes would reach `rootIsNZB` compressed, fail validation, and a grab that works today would start failing, because SAB is perfectly happy to accept a gzipped NZB upload.

I don't know whether anything on the Prowlarr redirect path actually serves that, which is why this is a question rather than a change request. If you've already checked, say so and I'll merge as is. If not, the guard is cheap: sniff the `1f 8b` magic and either accept it or gunzip before validating.

Small thing while you're in there: an empty 200 body gives a message ending in "is not an NZB: " with nothing after it. Worth a distinct phrasing for the empty case.

Two notes beyond this PR.

Your `httpsec.NewDialContext` finding is real and I'd like it as its own issue please. Dialing `ips[0]` with no fallback is a hard failure on a dual stack indexer host from an IPv4 only pod network whenever the resolver orders AAAA first, and that's a shape plenty of people run.

On the retry follow up, agreed it doesn't belong here. A transient fetch failure permanently failing the grab is the thing that actually makes this silent, so I do want it, just separately.

Catching that `TestPendingGrab_DispatchesDownloadAndClearsPending` was only passing because the "NZB" it fetched was SAB's own JSON response is the detail that tells me you read the code rather than just the failure.

helios57:
I had not checked the gzip case, so I checked it. You were right, and the exposure is wider than gzip.

Read it out of SAB rather than reasoning about it. `process_single_nzb` is documented as **"Supports NZB, NZB.BZ2, NZB.GZ and GZ.NZB-in-disguise"**, and it keys off the leading bytes, not the upload filename:

```python
with open(path, "rb") as nzb_file:
    check_bytes = nzb_file.read(2)

if check_bytes == b"\x1f\x8b":
    # gzip file or gzip in disguise
    nzb_fp = gzip.GzipFile(path, "rb")
elif check_bytes == b"BZ":
    # bz2 file or bz2 in disguise
    nzb_fp = bz2.BZ2File(path, "rb")
```

Bindery uploads under a `.nzb` filename regardless, so the extension was never what saved this — the magic sniff is. **bzip2 is the same regression**, so both are handled.

Confirmed as a real break rather than a theoretical one: `TestAddURL_GzippedNZBReachesSABVerbatim` was watched failing against the unfixed check, with a gzipped NZB rejected as "a body that is not an NZB" followed by a screenful of binary.

What changed:

| | |
|---|---|
| gzip / bzip2 body | unwrapped **for the check only** |
| what reaches the client | the original compressed bytes, unchanged |
| decompression | through a `LimitReader`, never expanded whole |
| archive that won't decompress | says so; raw bytes stay out of the message |
| empty 200 | its own phrasing, no dangling colon |

Two things worth flagging beyond what you asked for.

**The compressed bytes still go to the client verbatim.** Decompressing for validation and then uploading the expanded copy would have been the easy wrong move — SAB handles both, so it would have passed every test I had and quietly changed what gets spooled. `TestAddURL_GzippedNZBReachesSABVerbatim` asserts the uploaded payload is byte-identical to what the indexer served, so that can't drift later.

**Unwrapping is bounded.** Expanding an indexer-controlled archive to answer "is the root `<nzb>`?" is a decompression bomb waiting to happen, so the decompressed stream is read through a `LimitReader` — the root element is settled in the first few hundred bytes either way.

Your `endedAt` comment about a helper tripping `bodyclose` caught me in the act: I added an `ok200` helper and the linter flagged it for exactly the reason documented there. Call sites are literals again.

Also closed the one branch Codecov flagged — NZBGet had no test for the body-read path that SABnzbd already had.

#2120 landed while I was working; no interaction. Your tests serve a real NZB so validation passes them, and `TestAddURL_NonNZBBodyIsRejectedBeforeSAB` asserts the indexer-caused case never reaches the string your change improves.

Follow-ups filed as asked: **#2156** for the dialer and **#2157** for the retry. On #2156 I was explicit in the issue that I never observed it firing — RFC 6724 deprioritises IPv6 on a host with no global v6 address, so `ips[0]` came back IPv4 on my box and the missing fallback was never reached. It is a code-reading find, not an outage, and the issue says so.

Rebased onto current main and squashed to a single commit. Full suite, `golangci-lint` v2.11.4 and `govulncheck` all clean against the rebased tree.

Issue #2029, opened 2026-08-15
Download-client health is qBittorrent-only: every other client type stores a fabricated OK, and the UI would not show it anyway


vavallee:
## Current behaviour

Three things compound into "the client is fine, imports never happen, and nothing anywhere says otherwise".

**1. The health check returns a fabricated pass for five of six client types.** `internal/downloader/health.go:154`:

```go
func CheckDownloadClientHealth(ctx context.Context, client *models.DownloadClient, downloadDir, audiobookDownloadDir string) models.DownloadClientHealth {
	if client == nil || client.Type != "qbittorrent" {
		return models.DownloadClientHealth{Status: HealthOK, Message: "Download client path check not required"}
	}
	return checkQbittorrentCategoryPath(ctx, client, downloadDir, audiobookDownloadDir)
}
```

For SABnzbd, NZBGet, Transmission, Deluge and rTorrent this stores `HealthOK` without checking anything. The message asserts the check was "not required", which is not true for NZBGet or rTorrent: `downloader.CheckCompletedPathVisibility` introspects both of them today (`pathcheck.go:60-72`) and is already wired into the Test action.

**2. The UI would not display the result even if it were real.** `web/src/pages/settings/ClientsTab.tsx:107`:

```tsx
{c.type === 'qbittorrent' && c.health?.status === 'error' && (
```

So a stored health error on any other client type renders nothing.

**3. Health is never refreshed on its own.** The only writers are `refreshClientHealthAsync` on create (`download_clients.go:169`) and update (`:205`), and `refreshClientHealth` on an explicit Test (`:241`). A client that was healthy at setup and broke afterwards — a seedbox remount, a rotated password, a container that moved — stays green until somebody presses Test on a hunch.

## Why it matters

This is the mechanism behind the most common support shape: the connection is verified, the client accepts grabs, and nothing ever imports. Bindery can already diagnose it — `CheckCompletedPathVisibility` produces a precise verdict including the remapped path — but nothing asks the question unless the user already suspects the answer.

rTorrent makes it sharper. It shipped in v1.31.0, seedbox installs are the case most likely to need a path remap, and it is one of the five types that stores a fake OK.

This is the download-client twin of #1935, which is the same failure for indexers.

## Proposed change

[...]

Issue #1936, opened 2026-08-12
Auto-grab decides from whatever indexers happened to answer: a failed indexer silently shrinks the candidate pool


vavallee:
## Symptom

Three indexers are configured. Two are hard-failing (one suspended account, one rate limit). Auto-grab runs, sees the two results the third indexer returned, picks the best of those, and grabs it. Nothing records that two thirds of the candidate pool was never asked.

The book ends up with whatever the surviving indexer had. If the good release was on one of the other two, it is simply not grabbed, and no history row, no log line above debug, and no field on the book says why.

## Current state (verified on `main`)

`internal/indexer/searcher.go:239` — `SearchBook` returns `[]newznab.SearchResult` and no error:

```go
hits, err := client.BookSearch(ctx, c.Title, c.Author, cats)
if err != nil {
    slog.Warn("indexer search failed", "indexer", idx.Name, "error", err)
    return
}
```

A failing indexer contributes zero results and is otherwise indistinguishable from one that answered with nothing. `scheduler.searchAndGrabFormat` (`internal/scheduler/scheduler.go:619`) takes what comes back, filters it, ranks it, and grabs the top candidate. There is no signal available at that point to tell "nothing matched" apart from "we couldn't ask".

The interactive path is fine — `SearchBookWithDebug` keeps per-indexer errors and the panel shows them. It's the automatic path, the one that acts without a human looking, that has no idea.

## The question this needs answering first

What *should* auto-grab do when part of the pool is unreachable?

1. **Grab anyway, but record it.** Stamp the grab (history event detail or a field on the download) with "N of M indexers failed" and the reason. Cheapest, no behaviour change, makes the decision auditable after the fact.
2. **Grab anyway unless the winner is weak.** Proceed when the best candidate clears some quality bar, defer when it doesn't and a hard failure occurred — a marginal release is exactly the case where the missing indexer might have had better.
3. **Defer the book to the next scan when any indexer hard-failed.** Safest for release quality, but a permanently suspended account would stall every grab indefinitely, so this only works on top of #1935's health state and probably wants to ignore indexers already known to be down.

My preference is (1) on its own merits, and it is a prerequisite for either of the others — none of them can be built without the searcher first telling its caller that an indexer failed. Whether to then act on the signal is a policy call worth making deliberately rather than as a side effect.

## Implementation note

[...]

Issue #1934, opened 2026-08-12
A rate-limited indexer is queried again on the very next search: the 5xx retry hint is parsed and thrown away


vavallee:
## Symptom

An indexer answers a search with a Newznab 500:

```
indexer error 500: Request limit reached. Retry in 485 minutes.
```

Bindery records nothing. The next search — and every search for the following eight hours — sends that indexer another request it has already refused.

## Current state (verified on `main`)

1. **The classification exists but has no memory.** `newznab.IsRateLimitError` (`internal/indexer/newznab/types.go:51`) recognises the 5xx range, and it is consulted at `newznab/client.go:387,459,471,504` — but only to abort tier fall-through *within one search*. Nothing outlives the call.

2. **The retry hint is discarded.** "Retry in 485 minutes" arrives as free text in `IndexerError.Description` (parsed at `newznab/client.go:903`). The one piece of information that says exactly when to try again is never read.

3. **`Searcher` has no per-indexer state.** `internal/indexer/searcher.go:31-44` holds a client factory and a connection cache and nothing else. All three fan-out loops (`SearchBook` :249, `SearchQuery` :315, `SearchBookWithDebug` `debug.go:104`) query every enabled indexer unconditionally.

4. **Every search path re-hits it.** The scheduled wanted scan (`scheduler.go:911`) is paced, but that pacing is per book and knows nothing about indexer state; on top of it sit on-add searches, bulk, series add, recommendations, and interactive search. On indexers that count refused requests against the quota, this can stop the window ever clearing.

5. **The cost is a silently degraded decision, not just wasted requests.** `searcher.go:260-263`:

   ```go
   hits, err := client.BookSearch(ctx, c.Title, c.Author, cats)
   if err != nil {
       slog.Warn("indexer search failed", "indexer", idx.Name, "error", err)
       return
   }
   ```

   `SearchBook` returns `[]newznab.SearchResult` with no error, so a rate-limited indexer just contributes zero results and auto-grab proceeds from the smaller pool. For as long as the lockout lasts, books can be grabbed from worse releases while nothing anywhere records that the better indexer was never asked. (Filed separately: the visibility half.)

## Proposed fix

Per-indexer cooldown held on the shared `*Searcher` (one instance, `cmd/bindery/main.go:317`, used by both the scheduler and the API handlers):

[...]

PR #2127, opened 2026-08-20
feat(indexer): remember and report whether an indexer still answers (#1935)

### LINKED ISSUE #1935: An indexer whose account is suspended looks healthy in Settings: no stored health, no notification, still enabled
## Symptom

An indexer answers every search with:

```
indexer error 101: Account suspended
```

In Settings → Indexers it looks completely normal: enabled, no badge, no warning. Nothing is emailed, pushed, or posted. The only place the failure appears is the Search details panel on a book page, and only if you happen to run an interactive search and expand it.

An expired subscription or a revoked API key can therefore sit there for weeks, quietly removing an indexer from every automatic grab decision.

1. **`models.Indexer` has no health fields** (`internal/models/indexer.go`) — no last error, no last-success timestamp, no consecutive-failure count. The only stored state relevant to whether it works is `Enabled`, which only the user ever writes.

2. **Errors are logged and dropped.** `internal/indexer/searcher.go:260-263` and `internal/indexer/debug.go:~130` both do `slog.Warn("indexer search failed", ...)` and return. The debug path additionally puts the message in `IndexerDebug.Error`, which lives in `lastDebugStore` (`internal/api/indexers.go:22`) — the most recent interactive search only.

3. **Nothing notifies.** `internal/notifier/` contains no reference to indexers at all. Every other operationally interesting event (grab, import, import failure, stalled download) has an event type in `internal/models/settings.go:79-88`; indexer failure has none.

4. **The Test button is manual and ephemeral.** `IndexersTab.tsx:179` calls `api.testIndexer(id)` and renders the result into component state. Nothing is persisted, so the answer is gone on reload and you have to already suspect a problem to ask the question.

5. **Auth errors never self-heal.** `newznab.IsAuthError` (`internal/indexer/newznab/types.go:40`) already separates the 1xx range from rate limits precisely because they mean different things. A 5xx clears on its own; 100 (bad credentials), 101 (account suspended) and 102 (VPN forbidden) require a human, and the human is not being told.

## Proposed scope

- [ ] Persist per-indexer health: last error text, its Newznab code, when it happened, and the last time the indexer answered successfully. New columns on `indexers` plus fields on `models.Indexer`.
- [ ] Write it from the searcher's error branches, and clear it on any successful search.
- [ ] Show it on the Indexers tab: a badge on any indexer whose last attempt failed, with the message and the timestamp, so the state is visible without pressing Test.
- [ ] Notify on a hard auth failure (`IsAuthError`) — a new event type alongside the existing ones. Once per transition into the failed state, not once per search, or a 12-hour wanted scan would send hundreds.
- [ ] Do not auto-disable the indexer. Flipping `Enabled` is the user's switch, and silently turning it off would make the recovery path ("why is this off? I never touched it") worse than the failure.

## Relation to #1934

#1934 stops the *wasted requests* to a rate-limited indexer and deliberately excludes auth errors, because benching an indexer on a timer is the wrong response to a problem that only a human can fix. This issue is that other half: making the failure visible and telling someone. The two share a natural home for state — if #1934's in-memory cooldown grows a persisted form, it belongs in the columns this issue adds.

vavallee:
Closes #1935. First of the three-part service-health arc (#1935 → #1936 → #2029); the other two read the state this adds.

> **Merge order:** this carries migration **077**, which assumes #2123 (migration 076) lands first. If that one is rejected or reordered, renumber before merging.

## The bug

An indexer answering every search with `indexer error 101: Account suspended` looked completely normal in Settings: enabled, no badge, no warning.

- `models.Indexer` had no health fields. The only stored state relevant to whether it works was `Enabled`, which only the user ever writes.
- The searcher logged at WARN and returned. `IndexerDebug.Error` kept the message, but only in `lastDebugStore`, for the most recent interactive search.
- `internal/notifier/` had no concept of indexers at all.
- The Test button answers the question, but you have to already suspect the answer to press it.

So a lapsed subscription or a revoked key quietly removed an indexer from every automatic grab decision for as long as nobody looked.

## The change

Migration 077 adds `last_error`, `last_error_code`, `last_failure_at`, `last_success_at`. The searcher writes them from every error branch across all three fan-out paths (`SearchBook`, `SearchQuery`, `SearchBookWithDebug`) and **clears them on any success**, which is what stops a badge outliving the problem it describes.

Two things keep it from becoming noise:

- **Writes are debounced.** An unchanged state is rewritten at most every 5 minutes, so a 12-hour wanted scan searching per book does not issue an UPDATE per book per indexer. A state *change* is always written immediately.
- **The notification is edge-triggered** on entry into a hard auth failure. A persistent one notifies once; only after a recovery can it notify again.

**Only the Newznab 1xx range notifies.** 100, 101 and 102 mean bad credentials, suspended account and forbidden VPN, none of which heal on their own. A 5xx rate limit is recorded and shown in amber but not notified: it clears by itself and #1934 already stops asking. A transport failure stores no code at all, because that was not the indexer rejecting us.

The event reuses `EventHealth` rather than adding an enum, since "a configured service stopped working" is the same thing to a user's notification rules whether it is a download client or an indexer. The hardcoded `"Download Client Unhealthy"` title at `notifier.go:85` now branches on the payload.

**The indexer is never auto-disabled**, as the issue asks. `Enabled` is the user's switch, and flipping it for them turns "why is this failing" into the worse question "why is this off when I never touched it".

## Two things worth a close look

**`updated_at` is deliberately not bumped** by either health writer. #1934's cooldown treats a bumped `updated_at` as the user having edited the row and clears itself on the assumption they fixed something. Health is written by the searcher on a schedule nobody chose, so bumping it would keep resetting cooldowns and undo that fix. There is a test pinning this.

**#1934's in-memory cooldown stays in memory.** Its doc comment points here for persistence, but I have left it: cooldown is rate-limit pacing and health is operator visibility. They want different lifetimes, and folding them together would mean a restart wiping health along with the cooldown. Easy to revisit now that the columns exist.

## Testing

- `internal/indexer/health_test.go` covers 10 cases: failure recorded with its code, transport failure stores no code and does not notify, 50 repeat failures write and notify once, an unchanged state still refreshes after the interval, success clears and a later failure re-notifies, repeat successes write once, rate limits record but stay silent, a nil store is completely inert, id 0 is skipped, plus the `NeedsAttention` split.
- `internal/db/indexer_health_test.go` covers the column round trip, NULL code for transport failures, and the `updated_at` guard above.
- `internal/notifier/health_title_test.go` checks that the title names the right subject for clients and indexers.
- Full suite green (59 packages), frontend typecheck clean, lint 0 errors.


vavallee:

PR #2129, opened 2026-08-20
feat(scheduler): record which indexers failed to answer an auto-grab (#1936)

### LINKED ISSUE #1936: Auto-grab decides from whatever indexers happened to answer: a failed indexer silently shrinks the candidate pool

vavallee:
Closes #1936. Second of the service-health arc (#1935 → **#1936** → #2029).

> **Stacked on #2127.** Base is `feat/1935-indexer-health`, so review that one first; this diff is only the scheduler-side change. Retarget to `main` once #2127 merges.

## The bug

`Searcher.SearchBook` returns results and no error:

```go
hits, err := client.BookSearch(ctx, c.Title, c.Author, cats)
if err != nil {
    slog.Warn("indexer search failed", "indexer", idx.Name, "error", err)
    return
}
```

A failing indexer contributes zero results and is otherwise indistinguishable from one that answered with nothing. `searchAndGrabFormat` takes what comes back, filters it, ranks it and grabs the top candidate, with no signal available at that point to tell "nothing matched" from "we could not ask".

With two of three indexers down, the book gets whatever the survivor had. If the better release was on one of the other two it simply is not grabbed, and no history row, no field on the download and no log line above debug says why.

The interactive path never had this problem: `SearchBookWithDebug` already collects per-indexer errors and the panel shows them. It was only the automatic path, the one that acts with nobody watching, that had no idea.

## The change

Rather than a third search implementation, this projects what the debug path already gathers. `SearchBookWithOutcomes` returns results plus `[]IndexerDebug`, and the scheduler reaches it through an **optional interface**, the same shape `librarySnapshotter` uses in `internal/api/helpers.go`:

```go
type outcomeSearcher interface {
	SearchBookWithOutcomes(...) ([]newznab.SearchResult, []indexer.IndexerDebug)
}
```

A searcher that only implements `Searcher` keeps working and reports no outcomes, which is what keeps every existing stub compiling and every unwired caller behaving exactly as before.

Grabs decided on an incomplete pool are stamped with `indexersFailed: "2 of 3"` plus per-indexer names and errors, and logged at Info rather than debug.

**Skipped indexers are not failures.** One that is disabled, or benched by #1934's rate-limit cooldown, was never contacted. Counting it would flag a problem already being handled, and would make a normal install with one disabled indexer look permanently degraded.

## Scope

Deliberately **option 1** from the issue: this records, it does not decide. The grab is byte-for-byte what it was. Whether to defer a marginal grab while part of the pool is down is a policy question worth answering on purpose rather than acquiring as a side effect, and doing it safely needs #1935's health state so a permanently suspended account cannot stall every grab forever.

## Testing

`internal/scheduler/scheduler_indexer_outcomes_test.go`, six cases: the optional interface is used when present and falls back cleanly when absent, the failure record's shape and JSON round trip, a healthy pool adding no keys at all, skipped indexers excluded, and a stale error on a skipped entry ignored.

**One testing limitation worth flagging.** I first wrote these as end-to-end `searchAndGrabFormat` tests asserting the recorded history row, but history is only written after a successful send to a download client, and there is no download-client fake in the scheduler suite today (the existing freeleech tests deliberately point at `127.0.0.1:1` and observe the download row instead). So the payload construction is extracted into `describeIndexerFailures` and tested directly, with the search-path wiring tested separately. A genuine end-to-end assertion would need a fake SABnzbd or NZBGet server, which felt like a bigger change than this PR should carry. Happy to add it if you would rather have it.

Full suite green (59 packages).


vavallee:

PR #2131, opened 2026-08-20
fix(downloader): check health for every client type, and keep checking (#2029)

### LINKED ISSUE #2029: Download-client health is qBittorrent-only: every other client type stores a fabricated OK, and the UI would not show it anyway

vavallee:
Closes #2029. Third of the service-health arc (#1935 → #1936 → **#2029**). Independent of the other two: no shared files, no migration, so it can merge in any order.

## The bug

Three things compounded into "the client is fine, imports never happen, and nothing anywhere says otherwise".

**1. The check returned a fabricated pass for five of six types.**

```go
if client == nil || client.Type != "qbittorrent" {
    return models.DownloadClientHealth{Status: HealthOK, Message: "Download client path check not required"}
}
```

That message is not true. NZBGet and rTorrent both expose a completed path, and `CheckCompletedPathVisibility` has introspected them since #1182. It was simply never reached from the health path.

**2. The UI would not have shown it anyway.** `ClientsTab.tsx:107` gated the health block on `c.type === 'qbittorrent'`, so a stored error on any other type rendered nothing.

**3. Health was never refreshed.** Only written on create, update, or an explicit Test. A client that was healthy at setup and broke afterwards, a seedbox remount, a rotated password, a container that moved, stayed green until somebody pressed Test on a hunch.

## The change

The type gate is gone from all four places it appeared (`health.go:137`, `:154`, the `CheckingHealth` message, and both `download_clients.go` refreshers).

qBittorrent keeps `checkQbittorrentCategoryPath` because it validates both the ebook and audiobook categories rather than a single path. Everything else dispatches through `CheckCompletedPathVisibility`:

| visibility | health |
|---|---|
| `PathVisible` | `HealthOK` |
| `PathNotVisible` | `HealthError` |
| `PathUnknown` | **`HealthUnknown`** (new) |

`HealthUnknown` is the point. SABnzbd, Transmission and Deluge genuinely cannot be introspected, and "we could not check" is honest where "not required" was not. It renders muted rather than red, because it is not an error. An introspectable client that is merely unreachable also lands here rather than as a path error, preserving the existing "treat an introspection failure as can't-tell to avoid false alarms" behaviour.

A cron job re-probes every enabled client **every 15 minutes**, which is frequent enough to notice a broken mount within a download's lifetime, rare enough not to hammer remote clients. `HealthStore.Set` already edge-triggers its notification, so a persistent failure notifies once rather than every cycle.

### #1984 leftovers folded in

This makes those two reach every client rather than one, so they belong here: `rtorrentPathVisibility` no longer hardcodes `models.MediaTypeEbook`, and `nzbgetPathVisibility` no longer passes an empty expected-path hint. `qbittorrentExpectedHint` becomes `clientExpectedHint`, because every introspectable type wants the same sentence when the check fails.

rTorrent makes the whole thing sharper: it shipped in v1.31.0, seedbox installs are the case most likely to need a path remap, and it was one of the five types storing a fake OK.

## Worth flagging in review

**`downloadHealth` moves earlier in `main.go`.** The scheduler registers its cron jobs during `Start()` (line ~517), which ran before the health store was constructed (line ~578). The store only depends on the notifier, so it moves up next to it. No behaviour change beyond ordering, but it is the one structural edit in a file worth being careful with.

## Testing

`internal/downloader/health_all_types_test.go`, six cases: the three non-introspectable types report unknown and no longer claim the check was "not required", unreachable NZBGet/rTorrent report unknown rather than a false path error, a nil client is unknown, the Checking placeholder is no longer qBittorrent-specific, the expected-path hint names both directories when audiobooks are separate, and the async refresh now considers every enabled type while still skipping disabled ones.

Full suite green (59 packages), frontend typecheck clean, lint 0 errors, 590 vitest tests pass.


vavallee:

Issue #1733, opened 2026-08-04
Hardcover client ignores contribution roles — narrators and translators become the author


vavallee:
### Problem

The Hardcover client never asks which contribution role a name has, so narrators and translators can become the author. Reported on Discord #bug-reports by .v.e.g.a. (2026-08-03), verified by them against `main` @ 96b70c7 by direct source read; re-verified here.

Repro from the report: Will Wight's *Cradle* books "Blackflame" and "Dreadgod" kept landing under Travis Baldree — the series' audiobook narrator, not the author. Confirmed via the misfiled row's own `narrator` field reading "Travis Baldree". After moving them manually, a "Refresh All Metadata" run re-broke the same two books through the second code path below.

### Two sites, one root cause

**1 — `toBook()` (`internal/metadata/hardcover/client.go:1195`):**

```go
if len(b.Contributions) > 0 {
    a := c.toAuthor(b.Contributions[0].Author)
    bk.Author = &a
}
```

Takes index 0 unconditionally. Used by list-sync and single-book fetch.

**2 — `GetAuthorWorksByName()` (`client.go:194`):**

```graphql
where: { contributions: { author: { name: { _eq: $author } } } }
```

Matches *any* contribution role by name — no filter for Author vs Narrator/Translator/Illustrator. This is the path behind per-author Refresh Metadata and Refresh All, so the exposure isn't limited to list-sync, and it's why the manual fix didn't stick.

No query anywhere in the client requests the contribution role field at all, so neither site *could* filter today.

### Fix

Add the contribution/role field to both GraphQL queries and filter to the author role. Site 2 matters more — it's a structural query condition, not just a wrong index pick — and fixing it is what makes manual corrections stop reverting on refresh.

The same report's other finding against this query (name-only matching merging two different real people who share a pen name) is a distinct identity bug, filed separately.

Issue #1256, opened 2026-06-22
Hardcover numeric slug collides with DB-id lookup, returning the wrong book/editions/author


vavallee:
## Summary

`toBook`/`toAuthor` build the provider `ForeignID` as `"hc:" + slug`, where slug is Hardcover's human slug. Some real titles have a purely numeric slug (Orwell's *1984* → slug `1984`, *2001: A Space Odyssey* → `2001`). When that `ForeignID` is passed back to `GetBook`/`GetEditions`/`GetAuthor`, the code strips `hc:` and calls `hardcoverNumericID(id)`; for a numeric slug this succeeds and the query switches from a slug lookup to a **primary-key lookup**. So `GetBook("hc:1984")` fetches whatever book has database id 1984, and `GetEditions("hc:1984")` fetches editions of book_id 1984 — a completely unrelated book and its editions/cover/ratings.

The aggregator routes `hc:`-prefixed IDs here via `providerForForeignID`, so this fires in normal rebind / detail / edition flows for any title with a numeric slug.

## Details

`hardcoverNumericID` only checks `strconv.Atoi` success — there is no way to tell a numeric slug from a DB id. `GetAuthor` (`client.go:252`), `GetBook` (`:299`), `GetEditions` (`:367`) all switch to the id/book_id query for any numeric value.

## Fix

Prefer the slug lookup and only fall back to a numeric primary-key lookup when the slug matches nothing. This is backward compatible: a numeric slug resolves via slug; a genuine DB-id fallback (slug was empty at creation, stored as `hc:<id>`) still resolves via the id retry when the slug query returns no rows.

Found by an automated multi-agent code sweep; confirmed by two independent verifiers.

PR #2132, opened 2026-08-20
fix(hardcover): fetch a linked author's works by identity, not by name (#1734)

### LINKED ISSUE #1734: Hardcover author works are fetched by name even for linked authors — distinct authors sharing a pen name get merged
### Problem

`GetAuthorWorksByName` (`internal/metadata/hardcover/client.go:190`) selects works purely by author *name*:

```graphql
where: { contributions: { author: { name: { _eq: $author } } } }
```

Two different real people who publish under the same name are indistinguishable to that query. Reported on Discord #bug-reports by .v.e.g.a. (2026-08-03): two distinct authors both publishing as "J.A. Andrews" were merged into one author record, and the wrong one picked up ~44 unrelated books. Verified the query is unchanged on `main` @ 96b70c73.

### Why this is separate from the contribution-role issue

The role filter (filed separately) fixes narrators being mistaken for authors. It does nothing for this: both J.A. Andrewses hold genuine *author* contributions, so a role-filtered query still merges them. The missing dimension is identity, not role.

### Fix direction

Once an author row is linked to a Hardcover author (it carries a Hardcover foreign id), works should be fetched by that author *id*, not by name — name search belongs only in the initial resolve/link step where a human (or confidence scoring) picks which Hardcover author is meant. The relink-upstream candidates flow (`/author/{id}/relink-upstream/candidates`) already models "which of these same-named identities did you mean" and is the right UX for untangling an already-merged record.

A migration story for affected libraries is the hard part: an author row that already absorbed two identities' books has no recorded provenance for which book came from which identity, so the fix mostly prevents new merges; existing ones likely need manual relink plus per-book rebind.

vavallee:
Closes #1734.

## The bug

`GetAuthorWorksByName` selected works purely by author name:

```graphql
contributions: {author: {name: {_eq: $author}}}
```

Two different real people publishing under the same name are indistinguishable to that query, so it merges them. Reported on Discord by .v.e.g.a: two distinct authors both publishing as "J.A. Andrews", one of whom picked up ~44 books that were not theirs, re-applied on every metadata refresh.

The contribution-role filter from #1733 does nothing here, as the issue notes: both Andrewses hold genuine *author* contributions, so a role-filtered query merges them just as completely. The missing dimension was identity, not role.

## The change

`GetAuthorWorksByIdentity` selects on the linked author instead: slug first, numeric-id fallback, the same ordering `GetAuthor` uses and for the same reason (#1256: a purely numeric value may be a slug, or the DB-id fallback `toAuthor` emits when the slug is empty).

**It deliberately does not fall back to the name query when an identity finds nothing.** Widening there would re-merge the identities at exactly the moment the caller asked for precision; an identity matching no works means the author genuinely has none. There is an explicit test for this.

### Getting the identity to the metadata layer

The aggregator's supplemental seam was handed only a name:

```go
type authorWorksByNameProvider interface {
	Name() string
	GetAuthorWorksByName(ctx context.Context, authorName string) ([]models.Book, error)
}
```

...even though the full `models.Author` was already in scope at the call site. It now resolves the provider's own id from either the author's primary foreign id or the `author_identifiers` rows, which the sync loads into a **transport-only** `ProviderIdentifiers` map (no column, nothing persists it, the same shape as `Book.ProviderISBNs`). A relink moves the primary id to the new provider and keeps the previous one as an identifier, so both places are worth checking.

The identity also joins the author-works cache key, or a relink would be served the pre-relink answer.

**The Hardcover client also implements `worksProvider` now.** Without it, an author whose primary identity is already Hardcover fell through to `SearchBooks` (`aggregator_author_works.go:145-158`), another name-shaped lookup with the same problem. That was a second path to the same bug and it is closed here.

Name matching stays for the initial resolve step, where a human or a confidence score is about to pick which upstream author was meant, and where a name is genuinely all there is to go on.

## Migration story

**Prevention only**, as the issue anticipates. An author row that already absorbed two identities has no recorded provenance for which book came from which, so existing merges need the `/author/{id}/relink-upstream/candidates` flow followed by a per-book rebind. I have said so in the changelog entry rather than leaving users to discover it.

## Testing

- `internal/metadata/hardcover/author_identity_test.go` covers 6 cases: the query selects by slug and no longer mentions name, the role filter from #1733 survives, the numeric-id fallback runs only after the slug misses, **no fallback to the name query ever**, an empty identity issues no query at all, `GetAuthorWorks` routes through the identity path, and `GetAuthorWorksByName` keeps its shape.
- `internal/metadata/author_identity_supplement_test.go` covers 5 cases at the seam: identity preferred, the primary foreign id used when it is already the provider's, name used for an unlinked author, an identity with no works not widening, and the cache key being order-stable and identity-distinct.
- `internal/models/author_identity_test.go` covers 7 cases on `ProviderIdentity`.

Full suite green (59 packages).


vavallee:

PR #2134, opened 2026-08-20
feat(library): give books the provider identity map authors already have (#1705)

### LINKED ISSUE #1705: Duplicate volume rows after relinking an author to OpenLibrary alongside Hardcover series books
From discussion #1682 (Aiz0). Split out from the volume-collapse fix in #1704, which is a different cause.

### Current behaviour
With a Hardcover-linked series whose books were created by the series fill, using **Find better metadata** on the author and selecting their OpenLibrary entry, then refreshing metadata, produces two rows for the same volume — one from Hardcover (`hc:…`) and one from OpenLibrary (`OL…W`).

The reporter's screenshot shows two entries for volume 1 of *The Mimosa Confessions*. Their words: "I don't know how to fix that except for removing the hardcover book, but then i can't add the missing books to the series, I'd have to create it manually." So the workaround costs them the series link.

### Expected behaviour
Refreshing an author's catalogue from a second provider should reconcile against books already present from another provider rather than minting parallel rows, or there should be a supported way to merge the two.

### Why this is separate from #1704
#1704 fixes fuzzy title matching collapsing *distinct* volumes onto one row — the opposite failure. This is cross-provider duplication of the *same* volume, and the identity question ("is this OpenLibrary work the same book as this Hardcover edition?") is the same one behind #1691 (ABS retitle minting duplicates). Worth solving once.

### Steps to reproduce
1. Enable Enhanced Hardcover series, link a series, fill it.
2. On the author, Find better metadata → pick their OpenLibrary entry.
3. Refresh metadata.

### Deployment method
Docker (ghcr.io)

### Bindery version
v1.28.1 (`sha-90a07d0`)

### Trace logs
Not captured — reported via Discussions with screenshots. Reproduction steps above are precise enough to reproduce locally.

vavallee:
Closes #1705. Partially addresses #1691. Carries migration **078**.

## The bug

`books.foreign_id` is one column holding one provider's id, with the provider encoded in the prefix. Authors got a second identity table in migration 052 precisely because one column cannot express "this row is both `OL123W` and `hc:456`". Books never did, and that asymmetry is #1705.

The reported sequence: a Hardcover linked series fills its books, the user runs Find better metadata on the author, picks their OpenLibrary entry, and refreshes. `relinkExistingAuthorToUpstream` rewrites the *author's* foreign_id and preserves the old one in `author_identifiers`, but the books keep their `hc:` ids. The refresh fetches OpenLibrary works whose ids no row has ever held, the exact lookup misses every one, and the canonical title fallback misses whenever the two providers disagree about the title. So the same volume is created a second time. The reporter was left with two rows for volume 1 of *The Mimosa Confessions* and, in their words, no way out that did not cost them the series link.

## Why the table alone is not the fix

Worth being precise about this, because it caught me out. Widening the lookup to `book_identifiers` does not by itself help: the incoming OpenLibrary id is **genuinely unknown** to the existing row, so no amount of indexing finds it.

What connects them is `HardcoverForeignID`. `mergeAuthorWorks` already joins the two catalogues in memory before any of this and hands the OpenLibrary work the Hardcover id of the same book. That id is exactly what the existing row was created with. It was simply dropped at persist time because there was nowhere to put it.

So the fix is both halves: record it at create, and probe **both** ids in the sync. There is a test pinning this, and I verified it fails without the second probe, reproducing the reporter's exact symptom (two rows, `hc:mimosa-1` and `OL500W`, same volume).

## What else changed

- **Matches record what they learned.** A title match is a guess that happened to pay off, so attaching the ids turns it into an exact match next time. That matters most for the case that produced it: one volume seen through two providers whose titles agree today and may not after either side is edited.
- **A conflict is reported, not forced.** An id already claimed by a different row happens for duplicates created before this change and for genuine mistakes. Re-pointing it would move a book's identity out from under whoever is looking at it, so both rows are left alone and it is logged.
- **#1691 gets a partial fix from the same table.** `mergeUpstreamBook` rewrites a row's ForeignID to the upstream id, destroying the `abs:` key that step 2 of the ABS match ladder looks for. After that only provenance can match, and a provenance miss drops to the normalized title, which a rename in ABS has just invalidated. The old `abs:` id is now kept as an identifier and step 2 consults the table. The rest of #1691 stays open.

## Scope

**Prevention only.** Rows already duplicated have no record of which came from where, and merging them is destructive in a way that should be asked about rather than inferred. The changelog entry says so plainly rather than leaving users to find out.

The author merge path (`AuthorAliasRepo.Merge`) still reparents books without deduplicating them. That is a separate decision about whether an author merge should also merge books, and it is not made here.

## Testing

- `internal/db/book_identifiers_test.go` covers 8 cases: create records the primary id and the Hardcover link, both ids resolve to one row while the primary column stays untouched, unknown and blank ids resolve to nothing, a conflict is reported and the identity does not move, repeat attachment is idempotent, the FK cascade clears identifiers with their book (the failure migration 068 had to repair for authors), and the provider classification.
- `internal/api/authors_cross_provider_dedup_test.go` reproduces #1705 end to end and pins the opposite direction too: a genuinely distinct volume still creates its own row, which is the #1704 failure mode in reverse.

Full suite green (59 packages), golangci-lint 0 issues.


vavallee:

PR #2185, opened 2026-08-25
fix(calibre): stop recording co-authors as aliases of the primary author (#1684)

vavallee:
## Summary

A Calibre book credited to several people had every credit after the first written into `author_aliases` pointing at the primary author. An alias asserts "another name for this same person", so from then on every book by that co-author resolved back to their collaborator, and the co-author never got an author row at all. Which author survived came down to Calibre book id processing order, which is why the symptom looked arbitrary and why both reporters lost whole back-catalogues under a pen name.

This stops the importer minting those rows, and stops an unattributed alias deciding who an incoming Calibre credit is, so libraries already polluted recover on their next import without a single row being deleted. Refs #1684 (the scanner half of the issue is not in here, see below).

## What was wrong

`recordSecondaryAuthors` (`internal/calibre/importer.go`) recorded `cb.Authors[1:]` as aliases of `cb.Authors[0]`. `resolveAuthor` resolves exact `authors.name`, then `author_aliases.name`, then creates. A co-author who had not yet been created in their own right therefore resolved to the primary author's row forever.

The doc comment stated the intent plainly ("so future imports that see the same co-author-as-primary resolve back to the same Bindery row"). That intent is the bug: a Bindery book carries a single `author_id`, so there was never anywhere to put a second credit, and parking it in the alias table did not record a co-author, it destroyed an author.

Evidence in the issue: paulohenriquesg lost ~25 Kevin J. Anderson / Rebecca Moesta titles into a "Kem Antilles" pseudonym row; edsilberberg's logs are full of `calibre import: alias record skipped error="alias \"X\" already points at author N (refusing to reassign to M)"`, which is this function colliding with itself.

## What changed

| Change | Why |
|---|---|
| `recordSecondaryAuthors` deleted; a DEBUG line replaces it | Co-authors are not modelled. Dropping the credit loses nothing that was ever stored; keeping it lost the author. |
| New `aliasBindsAuthor` gates step 2 of `resolveAuthor` | An alias only decides identity when something backs it up. Repairs already-polluted installs with no data loss. |
| New `AuthorAliasRepo.GetByName` | `LookupByName` returns only an id; the trust decision needs `source_ol_id`. |

`aliasBindsAuthor` mirrors ABS's existing `trustedAuthorAlias`, which the Calibre path never had:

| Alias shape | Binds? | Reason |
|---|---|---|
| `source_ol_id` set | yes | An author merge or a provider record asserted these are the same human |
| Name matches the author's own name (exact or fuzzy-auto via `textutil`) | yes | A spelling or punctuation variant is what an alias table is for |
| Latin-script alias on a non-latin author name | yes | Exactly what `saveAlternateNames` writes so "Murakami" reaches "村上春樹"; the two can never look alike to a matcher |
| Unattributed and resembling nothing | no | The shape every co-author alias had. Logged at INFO and ignored |

## Scope decision: existing bad alias rows

**They stay. Nothing is deleted, and no migration.**

I checked `009_author_aliases.sql` (the only migration that touches the table) for a provenance column. There is only `source_ol_id`, and it is NULL for both the co-author rows *and* for everything `saveAlternateNames` writes (`internal/api/authors.go:3249`). Name similarity does not separate them either: the whole point of a "Haruki Murakami" alias on "村上春樹" is that the names look nothing alike, which is also true of "Kevin J. Anderson" on "Kem Antilles". **There is no discriminator, so any delete would be a guess**, and the rows it would take out by mistake are the ones non-latin authors depend on.

Ignoring them at resolution time gets the same user-visible outcome with none of that risk: the real author is created on the next import, the row stays for indexer search expansion (`scheduler.go:626` feeds aliases into `MatchCriteria.AuthorAliases`), and anyone who wants it gone can remove it from the author page, which already has a per-alias delete. The `TestImporter_IgnoresUntrustedPreExistingAlias` test asserts both halves: right author, row still there.

Worth noting the exact-name-shadow idea from the issue is already true and does not help: `resolveAuthor` tries `authors.name` before aliases, so an alias colliding with a real author name is unreachable there anyway. The reporters' broken installs are the case where the author row does *not* exist, which is why the fix has to be about trust rather than shadowing.

## Scope decision: the scanner half (deliberately not here)

`internal/importer/scanner.go` builds its author index from `authors.name` only (`authorNames` ~2394, `authorWordIndex` ~2470, `matchingAuthors` ~2484); nothing there reads `author_aliases`, which is why edsilberberg's follow-up sees `author_not_in_library` for names that exist only as aliases.

Leaving it for its own PR, because:

1. **Most of that symptom is this bug.** The names his scan could not find are names this importer minted. Once the real author rows exist, the scan finds them by name with no alias support at all.
2. **It is a behaviour change for every user, not a bug fix.** Teaching the index to read aliases means files start attracting to authors by a name that is not theirs. That wants its own tests and its own decision about which aliases qualify, and it would land better *after* the trust rule here settles what an alias is allowed to claim.
3. Mixing it in would make a contained importer fix into a change to the matching path everyone's library scan runs.

So `Refs #1684` rather than `Closes`. What remains on the issue: teach the library scan's author index to consult `author_aliases`. Happy to file the follow-up if you want it tracked separately.

## ABS importer: checked, no change needed

`internal/abs/import_author_matcher.go:442` logs the same "alias record skipped" message, but its `recordSecondaryAuthors` is guarded by `authorNamesAutoMatch(name, canonical.Name)`. It only aliases a co-author whose name already matches the canonical author's, so it records spelling variants of one person, never a distinct collaborator. It cannot produce this defect, and both reporters' log lines are `calibre import:`.

## Test that encodes the bug

`TestImporter_SecondaryAuthorsBecomeAliases` asserted the old behaviour directly ("secondary author should be recorded as alias"). It is now `TestImporter_SecondaryAuthorsAreNotAliases` and asserts the opposite, with a comment saying so. Flagging it rather than quietly dropping it.

## Before / after

`TestImporter_CoAuthorLaterCreditedAloneGetsOwnAuthor` is the user-visible regression. Source change stashed, tests kept:

```
=== RUN   TestImporter_SecondaryAuthorsAreNotAliases
    importer_test.go:359: co-author recorded as alias of author 1; co-authors must not become aliases
--- FAIL: TestImporter_SecondaryAuthorsAreNotAliases (0.08s)
=== RUN   TestImporter_CoAuthorLaterCreditedAloneGetsOwnAuthor
    importer_test.go:396: co-author never got an author row; authors = map[Kem Antilles:1]
--- FAIL: TestImporter_CoAuthorLaterCreditedAloneGetsOwnAuthor (0.08s)
=== RUN   TestImporter_IgnoresUntrustedPreExistingAlias
    importer_test.go:442: book was filed under the pseudonym via an untrusted alias
--- FAIL: TestImporter_IgnoresUntrustedPreExistingAlias (0.08s)
=== RUN   TestImporter_TrustedAliasesStillResolve
--- PASS: TestImporter_TrustedAliasesStillResolve (0.25s)
FAIL
```

Restored:

```
--- PASS: TestImporter_SecondaryAuthorsAreNotAliases (0.08s)
--- PASS: TestImporter_CoAuthorLaterCreditedAloneGetsOwnAuthor (0.08s)
--- PASS: TestImporter_IgnoresUntrustedPreExistingAlias (0.09s)
--- PASS: TestImporter_TrustedAliasesStillResolve (0.24s)
    --- PASS: TestImporter_TrustedAliasesStillResolve/provenanced_alias
    --- PASS: TestImporter_TrustedAliasesStillResolve/name_variant
    --- PASS: TestImporter_TrustedAliasesStillResolve/latin_alias_of_non-latin_author
PASS
```

`TestImporter_TrustedAliasesStillResolve` passes both ways by design: it is the guard that this fix does not trade one duplicate-author bug for another.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (`docs/Troubleshooting-Wiki.md` gains a symptom entry with the cleanup steps, `docs/User-Guide-Wiki.md` says what happens to multi-author Calibre books, godoc on the new helpers)
- [x] `changelog.d/1684-coauthor-aliases.md` added

## Test plan

- [x] `go test ./internal/...`
- [x] `go vet ./...`
- [x] `gofmt -l internal/` clean
- [x] `go build ./...`
- [ ] `cd web && npm run build` (no frontend change)


vavallee:
Took the fuzzy-tier note in 6ca54286. You are right that the reason belongs next to the code: an exact-only test would reject "RR Haywood" against "R.R. Haywood", which is the case the alias table exists for.

The rest checked out on my read too, in particular that the untrusted row goes permanently inert without being deleted, because the real author is found by name at step 1 on every later import.

PR #2246, opened 2026-08-26
fix(hardcover): carry author-work language from default editions (#2241)

### LINKED ISSUE #2241: A strict `unknownLanguageBehavior` silently deletes the entire Hardcover catalogue, because Hardcover author-works cannot carry a language
### Summary

Hardcover's author-works query deliberately does not select `language` —
`authorWorksProjection` (`internal/metadata/hardcover/client.go:193`) says so explicitly:

> *NB: do NOT select `language` here. It is an *edition* field; the `books` type has no
> `language`, so requesting it makes Hardcover reject the whole query … until that's
> added, supplemental books carry no language.*

So every work from a Hardcover-primary author arrives with `Language == ""`. A metadata
profile with `allowedLanguages` set and `unknownLanguageBehavior = fail` then rejects all
of them (`IsLanguageAllowed` returns `!unknownFail` for an empty code), and the only rows
that survive the sync are the ones the Audible supplement contributed — which carry no
Hardcover `SeriesRefs`, so no series are created either.

The user sees no error. The catalogue is simply wrong, and `metadata.primary_provider =
hardcover` is effectively inert.

### Version

v1.33.0, `metadata.primary_provider = hardcover`, profile `{allowedLanguages: "eng",
unknownLanguageBehavior: "fail"}`.

### Observed

Same author, same profile except `unknownLanguageBehavior`:

| | `fail` | `pass` |
| --- | --- | --- |
| total works | 203 | 203 |
| added | 61 | 97 |
| skipped_language | 139 | 21 |
| surviving `hc:` rows | **0** | 88 |
| surviving `audible:` rows | 61 | 9 |
| series created | **0** | 23 |

With `fail`, an author with 23 Hardcover series produced a catalogue of Audible leftovers
and no series at all.

### Why the existing mitigations don't cover this

Two mechanisms already exist for the blank-language problem, and both are
OpenLibrary-shaped:

1. **Edition sampling is inert under a Hardcover primary.**
   `FillMissingAuthorWorkLanguages` (`internal/metadata/aggregator_author_works.go:27`)
   type-asserts `a.primary` to `workLanguageFiller`, and `FillMissingWorkLanguages` is
   implemented **only** by the OpenLibrary client
   (`internal/metadata/openlibrary/client.go:731`). With Hardcover primary the assertion
   fails and it returns 0 without a word. The comment above the call site reasons entirely
   about OpenLibrary's `languages` gap.
2. **The majority-language fallback did not engage.**
   `applyAuthorMajorityLanguageFallback` requires ≥3 resolved languages and a ≥0.9
   majority share (`authors.go:3528`, `3534`). In the run above it left 139 works blank,
   presumably because the resolved subset was small and mixed.

### Suggested fixes, any of which would do

- **Treat "this provider structurally cannot report a language" as different from
  "unknown".** The filter is meant to catch untagged foreign translations from
  OpenLibrary; applying it to a provider that never reports a language at all is a
  category error. Exempting works whose `MetadataProvider` cannot supply one would fix
  this without weakening the OL case the setting was built for (#232, #891).
- **Implement `FillMissingWorkLanguages` for the Hardcover client** by traversing to a
  default edition, which is where the language actually lives — the projection comment
  already names this as the missing piece.
- **At minimum, warn.** A sync that skips most of the catalogue on the language filter
  while the primary provider reports no languages is a configuration error worth a WARN
  naming both facts, and worth surfacing in the profile editor when Hardcover is primary.

The `skipped_language` counter is already in the sync summary, but at DEBUG per book and
buried in a WARN line otherwise — nothing connects it to the provider's inability to
report a language.

kevinatlee:
## Summary

Hardcover author-work queries now select language through the default ebook and audio edition relations, allowing the existing `hcBook`/`toBook()` path to populate `Book.Language`. This prevents strict unknown-language profiles from rejecting matching Hardcover works while continuing to reject known non-allowed languages.

Closes #2241.

## Why minimal

- Reuses the default-edition language fields and conversion logic already used by Hardcover list/shelf queries.
- Adds no database schema or model changes.
- Adds no per-book request or API fan-out.
- Does not change reconciliation or deletion behavior.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared; this restores existing profile semantics without adding configuration or an API surface

## Test plan

- [x] `go test -race ./internal/metadata/hardcover -count=1`
- [x] `go test ./internal/metadata/... -count=1`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run --timeout=5m` (v2.11.4; 0 issues)

A repository-wide race run was also attempted locally. The changed Hardcover package passed, but the overall command did not complete cleanly on macOS: `internal/api` reached its 30-minute timeout and an existing downloader test hit the `/var` → `/private/var` temporary-directory symlink safety guard. Neither failure exercised this diff; the focused race test and all metadata tests pass.

### DIFF
--- a/changelog.d/2241-hardcover-author-language.md
+++ b/changelog.d/2241-hardcover-author-language.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Hardcover author language filtering** (#2241) — derive work language from default editions so strict unknown-language profiles retain matching books.
--- a/internal/metadata/hardcover/client.go
+++ b/internal/metadata/hardcover/client.go
@@ -192,12 +192,10 @@ func (c *Client) SearchBooks(ctx context.Context, query string) ([]models.Book, 
 // returns no supplemental results.
 // authorWorksProjection is the field set every author-works query returns.
 //
-// NB: do NOT select `language` here. It is an *edition* field; the `books`
-// type has no `language`, so requesting it makes Hardcover reject the whole
-// query ("field 'language' not found in type: 'books'", validation-failed)
-// and the entire author-works supplement fails (#1036-adjacent report). A
-// book's language can only be derived by traversing to a default edition;
-// until that's added, supplemental books carry no language.
+// NB: do not select `language` directly here. It is an *edition* field; the
+// `books` type has no `language`, so requesting it makes Hardcover reject the
+// whole query ("field 'language' not found in type: 'books'",
+// validation-failed). Derive it through the default-edition relations instead.
 const authorWorksProjection = `
 		id
 		title
@@ -213,6 +211,8 @@ const authorWorksProjection = `
 		audio_seconds
 		default_audio_edition_id
 		default_ebook_edition_id
+		default_ebook_edition { language { language } }
+		default_audio_edition { language { language } }
 		book_series(order_by: { position: asc }) {
 			position
 			series { id name }
--- a/internal/metadata/hardcover/client_test.go
+++ b/internal/metadata/hardcover/client_test.go
@@ -664,11 +664,16 @@ func TestGetAuthorWorksByName_WithToken(t *testing.T) {
 				t.Fatalf("query did not request Hardcover field %q: %s", field, req.Query)
 			}
 		}
-		// Regression: `language` is an edition field, not a `books` field.
-		// Requesting it makes Hardcover reject the whole query
-		// (validation-failed) and the author-works supplement fails entirely.
-		if strings.Contains(req.Query, "language") {
-			t.Fatalf("author-works query must not request the invalid books.language field: %s", req.Query)
+		// #2241: language belongs to editions, not books. Selecting both
+		// default-edition relations keeps the query valid and prevents strict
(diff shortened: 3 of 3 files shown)


kevinatlee:
Follow-up from real-world testing after #2246: the original blank-language fix was necessary, but we found that Hardcover's default edition can itself be a translation, so using that edition's language as authoritative work-level language can produce false non-English classifications.

Penelope Douglas reproduced this strongly: 49/78 author works were rejected by an English-only profile, including `Corrupt (por)`, `Hideaway (por)`, `Kill Switch (por)`, `Bully (Spanish)`, and `Nightfall (Spanish)`. An existing English Audible `Corrupt` row was consequently offered by catalogue reconciliation as a language-rejected stale row.

No files were deleted; #2247's preview/safeguards exposed the bad evidence before apply.

I've documented the reproduction in #2753 and submitted #2754 to resolve language from safer edition evidence while preserving the strict-profile behavior #2246 originally fixed.

Issue #2208, opened 2026-08-25
Metadata profile filters only screen newly discovered books, and the series fill path skips them entirely


vavallee:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

*From @ianepreston's report on #2093, plus a read of the code. Both gaps verified first-hand on current main. This is the reason #1780's problem statement no longer fits the evidence: he has Hardcover configured **and** `skipPartBooks: true` and still ended up with three "Expanse Boxed Set" rows and a "Leviathan Falls - Carton of 10 Signed Copies".*

### Current behaviour

The five metadata profile filters that went live in v1.32.0 (`SkipPartBooks`, `SkipMissingDate`, `MinPopularity`, `MinPages`, `SkipMissingISBN`) reach far less of the library than the settings imply. Two separate gaps, both structural.

**1. Enabling a filter never cleans an existing catalogue.** Every filter in the sync loop is gated on the book being new. `internal/api/authors.go:2255`:

```go
if existing == nil && skipPartBooks && isPartBookTitle(b.Title) {
```

The other four filters in that loop carry the same `existing == nil` condition. So turning a filter on after an author is already in the library screens only works discovered from that point onward. The bundle rows, the undated works, the unpopular ones already sitting there stay forever, through every subsequent refresh. There is no reconciliation pass and no affordance anywhere in the UI to apply a filter to what is already stored.

That is the opposite of how a filter setting reads. Someone who notices box sets in their library, finds the setting, and turns it on will reasonably expect the box sets to go away.

**2. The series fill path applies no profile filters at all.** `isPartBookTitle` has exactly one call site in the entire codebase, inside `fetchAuthorBooks`:

```
$ grep -rn "isPartBookTitle(" --include=*.go internal/ | grep -v _test
internal/api/authors.go:151:func isPartBookTitle(title string) bool {
internal/api/authors.go:2255:  if existing == nil && skipPartBooks && isPartBookTitle(b.Title) {
```

`ensureHardcoverCatalogBook` (`internal/api/series.go:1345`) creates books through `h.books.Create` at `:1489` having consulted no metadata profile. It sets `author.MetadataProfileID` and defaults a language, and that is the extent of it: no part-book check, no popularity floor, no page count, no ISBN requirement, no language filter. So "add all" on a series tab, and the per row add beside it, bypass every filter the author's profile declares.

[...]

Issue #2147, opened 2026-08-20
SetMaxOpenConns(1) serializes reads as well as writes, forfeiting the point of WAL


vavallee:
`internal/db/db.go:41-42` (and `:69-70` for `OpenMemory`).

```go
db.SetMaxOpenConns(1)
db.SetMaxIdleConns(1)
```

Every query in the process, read or write, goes through one physical connection. WAL's central property is that readers do not block a writer and a writer does not block readers, and pinning the pool to one connection gives all of that back.

It shows up wherever concurrent requests overlap: the UI polls `/queue` every 5s and the authors refresh status every 2s, OPDS clients read the catalogue, and a library scan or import is writing throughout. All of it queues behind the same connection.

### The original reason no longer applies, but a different one does

The comment at `db.go:149-166` explains the pin as protecting pragma state. That reason is now stale: #1727 moved `foreign_keys` and `busy_timeout` into `connectionPragmaDSN`, and the driver re-applies DSN pragmas on every connection open, which is exactly what makes a pool safe. The comment even says so: *"SetMaxOpenConns(1) does not prevent this: it bounds how many connections exist at once, not whether the one connection is ever replaced."*

There is a real blocker underneath it though, and it is not the one documented. Migrations run:

```go
database.Exec("PRAGMA foreign_keys=OFF")   // db.go:637
... apply migration ...
database.Exec("PRAGMA foreign_keys=ON")    // db.go:644
```

`PRAGMA foreign_keys` is connection scoped. With a pool, the `OFF` could land on one connection and the migration body execute on another, which still has foreign keys enforced. Migrations that rebuild a table by create-copy-swap depend on that `OFF`, so this would not fail loudly, it would fail as a constraint violation in the middle of a schema change.

### What a fix has to do

* Keep the pool pinned to 1 for the whole of `migrate()`, and only raise `SetMaxOpenConns` after migrations have completed.
* Leave `OpenMemory` pinned at 1 permanently. In-memory databases are per connection, so a pool would give each goroutine its own empty database. The existing comment at `db.go:60-63` already says this and it stays true.
* Carry a test that foreign key enforcement is on for **every** pooled connection, not just the first, since that is the property most likely to regress silently.
* Decide the write path. SQLite still allows one writer at a time; with a pool, concurrent writes surface as `SQLITE_BUSY` and lean on `busy_timeout(5000)`. Worth checking whether 5s is enough under a scan, or whether writes should be funnelled through a dedicated single connection with reads pooled.

[...]

PR #2247, opened 2026-08-27
feat(author): add safe catalogue reconciliation (#2208)

kevinatlee:
## Summary

- add an explicit **More → Reconcile catalogue…** preview/apply flow for one author; normal metadata refresh remains non-destructive
- compare local rows with a fresh effective catalogue (primary provider plus configured supplements) and the current metadata profile, reporting candidates, protection counts, indeterminate rows, and per-book reasons
- let users deselect individual preview rows and submit only the chosen IDs; every candidate starts selected
- treat partial primary or supplemental results conservatively: absence is never deletion evidence after a provider failure, and unresolved language/edition evidence is kept
- recompute the preview on apply, intersect it with the requested IDs, and repeat every safeguard in one guarded database delete
- preserve imported/owned books and any row with `filePath`, `ebookFilePath`, `audiobookFilePath`, or a `book_files` record; no filesystem deletion is involved
- document the UI/API behavior and add a changelog fragment

The branch is rebased onto current `main`, including the merged Hardcover language fix from #2246. Reconciliation remains a separate explicit operation and does not make refresh destructive. Refs #2208; the separate series-fill behavior described there is intentionally outside this PR.

## Suggested review order

1. `internal/db/books.go` and `internal/api/catalogue_reconciliation.go` — deletion boundary and apply-time safeguards
2. `internal/metadata/aggregator_author_works.go` and OpenLibrary completeness reporting — conservative catalogue evidence
3. `internal/api/catalogue_reconciliation_test.go` and `internal/db/books_reconciliation_test.go` — destructive/error-path coverage
4. `web/src/components/CatalogueReconciliationModal.tsx` — explicit preview and per-row selection
5. route wiring, API docs, user guide, and changelog fragment

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed (not applicable: no deployment/config change)

## Test plan

- [x] `go test -timeout=30m ./cmd/... ./internal/...`
- [x] reconciliation API race tests, `-count=10`
- [x] guarded DB deletion race tests, `-count=10`
- [x] focused coverage: 100% statements for every function in `catalogue_reconciliation.go` and for `DeleteMetadataOnlyWantedByIDs`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `golangci-lint run ./...` (0 issues)
- [x] pinned `govulncheck ./...` (no called vulnerabilities)
- [x] frontend typecheck, lint, production build, and full test suite (67 files / 657 tests)
- [x] real-binary HTTP smoke suite (`go test -count=1 -timeout=60s ./tests/smoke/...`)

The full backend run used `TMPDIR=/private/tmp` on macOS so the existing rTorrent deletion safety test sees a direct path instead of the `/var` → `/private/var` system symlink.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ Clean-room Go rewrite, modern React UI, MIT-licensed, actively developed.
 - Series support with position tracking, edition tracking (format / ISBN / publisher / page count), Calendar view of upcoming releases, and multiple library roots.
 - Library scan with four-tier matching: ASIN → title + author → series name + position → fuzzy title. Honours librarian sort-suffix form (`Title, The`) and series-annotated filenames (`[Mistborn, Book 1]`).
 - Author aliases (`RR Haywood` / `R.R. Haywood` / `R R Haywood` merge into one canonical row), and metadata re-bind to correct a wrong match without delete-and-re-add.
+- Explicit author-catalogue reconciliation with a selectable preview: remove chosen stale metadata-only Wanted rows after changing provider or metadata profile while always protecting imported books and every row with a tracked file.
 
 **Search & downloads**
 - Newznab + Torznab indexers queried in parallel, deduplicated, then composite-ranked by format quality, edition tags (RETAIL / UNABRIDGED / ABRIDGED), year match, grab count, size, and ISBN exact-match bonus.
--- a/changelog.d/2208-author-catalogue-reconciliation.md
+++ b/changelog.d/2208-author-catalogue-reconciliation.md
@@ -1,0 +1,2 @@
+### Added
+- **Safe author catalogue reconciliation** (#2208) — preview, select, and explicitly remove stale metadata-only Wanted rows after a provider or metadata-profile change while protecting imported and file-bearing books.
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -890,6 +890,8 @@ func main() {
 		r.Put("/author/{id}/genres", authorHandler.ApplyGenres)
 		r.Delete("/author/{id}", authorHandler.Delete)
 		r.Post("/author/{id}/refresh", authorHandler.Refresh)
+		r.Get("/author/{id}/catalogue-reconciliation", authorHandler.PreviewCatalogueReconciliation)
+		r.Post("/author/{id}/catalogue-reconciliation", authorHandler.ApplyCatalogueReconciliation)
 		r.Get("/author/{id}/relink-upstream/candidates", authorHandler.RelinkCandidates)
 		r.Post("/author/{id}/relink-upstream", authorHandler.RelinkUpstream)
 		r.Get("/author/{id}/series", authorHandler.ListSeries)
--- a/docs/API.md
+++ b/docs/API.md
@@ -39,6 +39,10 @@ GET    /api/v1/author/{id}                        author detail
 PUT    /api/v1/author/{id}                        update monitored / metadata profile
 DELETE /api/v1/author/{id}                        remove (with optional file delete)
 POST   /api/v1/author/{id}/refresh                re-pull works from OpenLibrary
+GET    /api/v1/author/{id}/catalogue-reconciliation
+                                                    preview stale metadata-only Wanted rows
+POST   /api/v1/author/{id}/catalogue-reconciliation
+                                                    recheck and remove selected preview rows
 GET    /api/v1/author/{id}/relink-upstream/candidates
                                                     search metadata candidates for manual relink
 POST   /api/v1/author/{id}/relink-upstream        re-bind to a different foreign ID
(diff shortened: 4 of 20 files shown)


vavallee:
Read the whole thing and ran it locally. This is careful work, and the deletion boundary in particular is built the way I'd want it built: ownership checked before anything else, apply recomputes the preview and intersects rather than trusting the client's ids, and every predicate repeated inside the SQL so an import that lands between preview and apply wins the race. The route sits next to `DELETE /author/{id}`, which is the right precedent for gating. Thanks for the suggested review order too, it made this much faster to go through.

Two things I want changed, then one limitation I think needs saying out loud rather than fixing.

### 1. An empty `ForeignID` is treated as a real identifier

`buildCatalogueReconciliation` keys both maps on the provider work's id:

```go
acceptedIDs[work.ForeignID] = struct{}{}   // line 270
rejectedIDs[work.ForeignID] = reason       // line 278
```

and then looks a local book up by its own `ForeignID` plus its `book_identifiers` rows (lines 332 to 339). Nothing filters an empty string on either side, so `""` becomes a shared key. You already guard for this exact case in the editions fan-out at line 242, so I think it's just an oversight rather than a decision.

I wrote a throwaway test against your branch to be sure. A provider work with no id, rejected on language, plus a local Wanted book with no id:

```
candidates=[{BookID:1 Title:A Locally Created Book Reason:language_not_allowed}]
```

The local book is offered for deletion under a reason it never earned. The mirror case also reproduces: an accepted empty-id work grants an unrelated empty-id local book a spurious Kept.

Reachability is genuinely narrow, and I want to be fair about that: `books.foreign_id` is `TEXT NOT NULL UNIQUE`, so at most one such row can exist in a database, and I could not find a creation path that writes one. So this is latent rather than something users will hit next week. I still want it closed, because it is a one line guard on a code path whose failure mode is deleting the wrong row, and because "no current path creates one" is a property of today's code rather than a property this function enforces.

### 2. `ListBookIdentifiers` runs once per local book

Line 332 issues a query per row inside the loop. On a 500 book author that's 500 round trips, and with `SetMaxOpenConns(1)` (#2147) they serialize behind every other reader. A single fetch keyed by book id before the loop would be a small change and keeps preview latency flat.

### 3. Provider call volume, and a limitation worth documenting

I measured this rather than guessing. A 300 work author with `minPages` set costs **300 `GetEditions` calls per preview** at concurrency 4. Apply re-fetches the catalogue but reuses cached editions, so it is 2 catalogue fetches and 300 edition calls total rather than 600, which is better than I expected.

I'm not asking you to change it: it's the same `RunBounded(..., authorAutoSearchConcurrency, ...)` shape already used at `internal/api/authors.go:330`, so it is consistent with the codebase. Worth knowing that it lands on the providers we just had to add throttling to in #2277.

The one I do want said explicitly: `searchAuthorWorks` requests `limit=200`, and `GetAuthorWorksSnapshot` returns complete only when **both** halves are complete. So for any OpenLibrary author with more than 200 works, `ProviderComplete` is false permanently, the warning banner always shows, and reconciliation can never remove a `not_in_current_catalogue` row. That is the safe direction and I would not want it inverted. But it means the feature's headline capability quietly does not apply to exactly the prolific authors whose catalogues are worst, and a user staring at the banner has no way to know it will never clear. A sentence in the user guide saying so would be enough.

### Not your problem

The red `licenses` check is infrastructure. `go install github.com/google/go-licenses@v1.6.0` failed on `sum.golang.org` returning an HTTP/2 `INTERNAL_ERROR` while verifying a `genproto` go.mod, so the drift comparison never ran. Your branch touches no `go.mod`, `go.sum`, `package.json` or `THIRD_PARTY_LICENSES.md`. I'll re-run it.

Fix 1 and 2 and I'm happy to merge this.

kevinatlee:
Addressed in b0d43b5d after rebasing onto current `main`:

- Blank or whitespace-only provider and local IDs are no longer indexed or matched during reconciliation. Added focused regressions for both the rejected-empty-ID false candidate and accepted-empty-ID false kept-match cases.
- Reconciliation now fetches all identifiers for the author in one query and indexes them by book ID before the local-book loop. Added a DB test covering grouping and author scoping; the existing cross-provider alias behavior remains covered.
- The user guide now states that OpenLibrary `searchAuthorWorks` uses `limit=200`, so authors over 200 works remain partial, the warning may remain, and `not_in_current_catalogue` rows stay protected.

The branch is rebased onto current upstream `main`. All CI checks are green, including lint, Go/frontend validation, smoke, licenses, patch coverage, and all six race shards.

vavallee (approved):
All three land. Checked locally rather than taking CI's word for it.

Empty and whitespace ids are now out of `acceptedIDs`, `rejectedIDs` and `indeterminateIDs`, and out of the per book `ids` slice. I reverted just those four guards on a scratch copy and both new tests fail there, so they are real regressions rather than tests that would have passed before the fix.

`ListBookIdentifiersByAuthor` is scope equivalent to the loop it replaces: `ListByAuthorIncludingExcluded` selects on `WHERE author_id = ?`, and the new query joins `books b ON b.id = bi.book_id WHERE b.author_id = ?`. Same book set, same columns, same per book ordering, one query instead of one per book. Moving the error context from the book id to the author id is the right call given there is no longer a per book call to attribute it to.

The 200 work ceiling being written down is what I wanted. That is the honest description of the behaviour until the OpenLibrary side paginates.

Branch builds and vets clean here, and `go test ./internal/api/... ./internal/db/...` is green.

Approving. The regression tests that actually fail without the fix are the part most PRs skip, so thanks for doing it properly.

PR #2292, opened 2026-08-31
fix(importer): merge audiobook import into existing ebook folder instead of splitting into "Title (2)"

### LINKED ISSUE #1959: Let ebook and audiobook share one book folder instead of appending ' (2)'

schmitzkr:
## Summary

Fixes #1959. Follow-up to your review on the issue.

Under the shared-folder layout (`BINDERY_AUDIOBOOK_DIR` unset, so both templates resolve to the same folder), importing a book's audiobook after its ebook lands it in a `Title (2)` sibling. `UniqueDir` can't tell "same book, other format" from a clash with an unrelated book.

In the plain whole-folder placement path, if the computed audiobook destination is exactly the folder this book already occupies via its own on-disk ebook (`existingEbookDir`), it now merges into it. That needed placement functions accepting an existing directory, since `CopyDirCtx`/`HardlinkDir`/`moveDirCtx` all refuse one; that guard is left untouched and the new `*Merge` siblings are only reachable once the caller has verified the destination is this same book's folder. Same-named files are skipped rather than overwritten, with the skipped names surfaced on the History entry and `bookImported` payload. Move mode keeps the source when anything was skipped.

**The rollback hazard you found is fixed.** You were right: the merge reassigns `destDir` to the ebook's folder, and the post-placement rollback was still `os.RemoveAll(destDir)` for copy/hardlink, so a persistent `SetFormatFilePath` failure deleted the ebook. I reproduced it with your recipe (`PRAGMA query_only = ON` after the ebook import) before fixing it.

Of your two suggested fixes I took "skip the rollback when `destDir` already existed" rather than reusing `rollbackPlacedFiles`, because this merge recurses into pre-existing subdirectories, so "what was placed" isn't the flat basename list that helper takes and a correct unwind would also have to avoid removing pre-existing parent dirs. Skipping is provably safe and self-correcting: the placed files stay, the failure message points at them (as move mode already did), and a retry is idempotent because the merge skips whatever is already there. It is also the same invariant the `*Merge` functions document.

Flatten is still deliberately untouched, per your confirmation, with a test pinning it so the limitation cannot be removed without reworking that rollback first.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated: n/a, no env var, config or upgrade-path change
- [ ] Wiki pages updated: n/a, fixes existing documented behaviour

## Test plan

- [x] `gofmt -l .`, `go vet ./...` clean
- [x] `golangci-lint run --timeout=5m` (pinned v2.11.4): 0 issues
- [x] `govulncheck` (pinned rev): 0 vulnerabilities in this code
- [x] Frontend and Docker steps of `make check` not run: no frontend or Dockerfile surface here

Tests in `internal/importer/scanner_shared_format_folder_test.go`, the first three table-driven across hardlink/copy/move:

- `AudiobookMergesIntoExistingEbookFolder`: both formats end up in one folder
- `AudiobookMergeSkipsSameNamedFile`: collision skipped not overwritten, recorded in History; move keeps the source
- `AudiobookMergeIntoExistingSubdirectory`: recurses into a pre-existing same-named subdirectory
- `AudiobookDoesNotMergeIntoStaleEbookRow`: an ebook row whose file is gone does not trigger a merge
- `AudiobookTemplateFlattenDoesNotMerge`: pins the flatten limitation; ebook survives, import still succeeds
- `MergeRollbackDoesNotDeleteEbook`: your repro, in copy and hardlink

One note: `TestIndexerUpdate_WriteOnlyAPIKey` (`internal/api`, untouched here) failed once during a full-suite run on a loaded machine, then passed on re-run, 5/5 isolated on this branch and 20/20 on clean `main`. Looks like a pre-existing flake under load rather than anything from this change, but flagging it in case CI hits it.

## AI disclosure

Written with Claude Code (Anthropic's AI coding assistant). It root-caused the issue from the importer and renamer source, implemented the change and the tests, and reproduced your rollback finding before fixing it. I reviewed the design decisions and the diff. Being straight about the limits: the branch has only ever run against synthetic temp directories, not a real library.


schmitzkr:
CodeQL's 4 alerts are the `os.Stat(src)` entries and the move's `os.RemoveAll(src)`. Same shape as `copyDirPublicCtx`, `HardlinkDir` and `moveDirCtx`, which the `*Merge` variants mirror deliberately; per-file work still goes through `os.Root`. Flagged only because the lines are new. Happy to restructure if you would rather they not add to the branch baseline.

Codecov was fair, fixed in `ef6f0c5f`: the placements were only reached via the scanner's happy path, so the guards went untested. Direct unit tests now.

vavallee:
Rollback fix verified. I ran my own repro against `ef6f0c5f` rather than relying on yours, in both copy and hardlink mode, and the epub survives with the m4b beside it:

```
copy: ebook survived, folder holds [We Who Wrestle with God - Jordan B. Peterson.epub book.m4b]
hardlink: ebook survived, folder holds [We Who Wrestle with God - Jordan B. Peterson.epub book.m4b]
```

Skipping the rollback rather than reusing `rollbackPlacedFiles` is the right call, and for the reason you give: the merge recurses into pre-existing subdirectories, so the flat basename list that helper takes would not describe what to unwind, and a correct version would also have to track which parent directories it created. Putting `mergedIntoExistingFolder` first in the switch is what makes the invariant hold rather than merely be documented.

`renamer_merge_test.go` exercising the placement functions directly, `NeverRemovesTheDestination` in particular, is what stops a future edit quietly reintroducing this. That is the part I would have asked for next.

On the red CodeQL: not blocking, and not yours. The four new `go/path-injection` alerts are the `os.Stat(src)` at the top of each `*Merge` function plus the `os.RemoveAll(src)` in `MoveDirMergeCtx`. `main` already carries about twenty open alerts of that same rule in this same file, on the identical constructs in the non-merge siblings (renamer.go 792, 809, 880, 1140, 1191 and onward), none of them dismissed. Your merge walks themselves go through `os.OpenRoot`, same as the code they mirror. I will dismiss the four rather than ask this PR to solve a pattern that predates it.

Local run on your head: `go build ./...` and `go vet ./internal/...` clean, `go test ./internal/importer/... ./internal/api/` green. `TestIndexerUpdate_WriteOnlyAPIKey` passed here in a full package run, so I am inclined to agree with your read that it is load flake rather than anything you did.

Good to merge from my side. Thanks for the writeup on the design decision, and for being straight about the synthetic-only testing.

vavallee (approved):
Approving. Verification and the CodeQL read are in the comment above.

Issue #2306, opened 2026-09-02
Re-bind to OpenLibrary does not update release date or language from the selected work


yarafie:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Re-binding a book from a Hardcover record to the correct OpenLibrary work successfully changes the metadata provider / foreign ID, but the book's release date and language are not refreshed.

Example:

Book: Paying the Piper — David Drake

Before re-bind:
Provider: Hardcover
Foreign ID: hc:paying-the-piper
Release date: 0007-01-01
Language: unknown

Re-bound to:
Provider: OpenLibrary
Foreign ID: OL1000175W

The re-bind succeeds and the book now shows OpenLibrary OL1000175W as Current, but it still shows:

Published: Jan 1, 7
Language: unknown

OpenLibrary shows this work as first published in 2002. The book's Edit Metadata dialog can also suggest "en", but the stored language remains unknown.

Refreshing the author after the re-bind does not correct the date either.

This appears to be specific to the OpenLibrary single-work metadata path used by Re-bind. Search/author-work enrichment already uses OpenLibrary first_publish_year, while GetBook() fetches /works/{id}.json and does not populate ReleaseDate or derive Language from editions.

### Expected behaviour

Re-bind should refresh the book from the newly selected metadata identity, including release date and language when OpenLibrary can provide or derive them.

For OpenLibrary, the single-work/Re-bind path should ideally use the same enrichment already available elsewhere in Bindery:

- first_publish_year for ReleaseDate
- edition-derived language when the work itself has no language

In the example above, re-binding to OL1000175W should result in a 2002 release year and English language rather than retaining the stale imported values.

This is especially important because Re-bind clears metadata locks and represents an explicit request to replace metadata from the newly selected record.

### Steps to reproduce

1. Have an existing book with stale/incorrect metadata, e.g.:
   Paying the Piper — David Drake
   Provider: Hardcover
   Foreign ID: hc:paying-the-piper
   Release date: 0007-01-01
   Language: unknown

2. Open the book page and click Re-bind.

3. Select OpenLibrary and enter:
   OL1000175W

4. Complete the Re-bind.

5. Observe that:
   - Provider changes to OpenLibrary.
   - Current foreign ID changes to OL1000175W.
   - Re-bind is recorded in History.
   - Release date remains Jan 1, 7.
   - Language remains unknown.

[...]

PR #2308, opened 2026-09-02
fix(metadata): fill OpenLibrary release date and language on the single-work fetch (#2306)

vavallee:
## Summary

Fixes the Re-bind half of #2306.

An OpenLibrary work record carries neither a publication date nor a language, so `GetBook` returned a book with neither set. `BookHandler.Rebind` copies upstream metadata only where the upstream record has it, so a book re-bound to an OpenLibrary work kept the stale date and language of the record it was replacing.

Extending `workResponse` would not have helped. The endpoint genuinely does not carry the fields, which the reporter's own example shows:

```
GET /works/OL1000175W.json
  keys: authors, covers, created, description, key, last_modified,
        latest_revision, revision, subjects, title, type
  first_publish_date = None
  languages = None

GET /search.json?q=key:/works/OL1000175W
  {"first_publish_year": 2002, "language": ["eng"], "key": "/works/OL1000175W", ...}
```

`GetBook` now enriches from the search index, which is the same source `GetAuthorWorks` already uses for both fields, so a re-bound book agrees with what an author sync would have produced for it. Language goes through the existing `pickPreferredLanguage`, so it is stored as `eng` consistently with the rest of the OpenLibrary path.

The enrichment is best-effort: it never overwrites a value already present, it verifies the returned doc's key is the work that was asked for before copying anything off it, and a failed or empty lookup leaves the fields alone rather than failing the Re-bind that called it.

## Out of scope

The issue's second observation, that a subsequent author refresh does not repair the date either, is real and left alone here. The existing-book branch of author sync (`internal/api/authors.go:2362`) updates ratings, cover URL and Hardcover-sourced genres and deliberately leaves the rest, so it never touches `ReleaseDate` or `Language`. Widening that would rewrite dates across every existing book on every refresh, which is its own decision rather than a detail of this fix.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated: n/a, no env var, config or upgrade-path change
- [ ] Wiki pages updated: n/a, fixes broken behaviour rather than changing documented behaviour

## Test plan

- [x] `gofmt -l` clean, `go vet ./internal/metadata/...` clean
- [x] `go build ./...`
- [x] `go test ./internal/metadata/... ./internal/api/ ./internal/abs/`

Three tests in `internal/metadata/openlibrary/client_http_test.go`:

- `TestGetBook_HTTP_EnrichesDateAndLanguageFromSearch`: the reported case, asserting the query sent is `key:/works/OL1000175W` and that 2002 and `eng` land on the book. Fails against the unpatched `client.go`.
- `TestGetBook_HTTP_SearchEnrichmentIsBestEffort`: the search index answers 500 and `GetBook` still returns the work it did fetch.
- `TestGetBook_HTTP_SearchEnrichmentRejectsDifferentWork`: a doc for another work donates nothing.

The existing `GetBook` HTTP tests register no `/search.json` handler, so they exercise the best-effort path as a side effect and still pass unchanged.


vavallee:
Holding this out of v1.33.3. A review pass turned up three things, and two of them are mine to fix before this goes anywhere near a release.

**The early return never fires, so every `GetBook` pays an extra search request.** `workResponse` (`types.go:95-103`) has no date and no language field, and nothing between building `b` and the enrichment call assigns either, so `if b.ReleaseDate != nil && b.Language != ""` can never be true. I confirmed that by reading the struct rather than taking it on trust. That is a 50% request increase on `GetBook`, against `/search.json`, which is the endpoint OpenLibrary throttles hardest, with no pacer in front of it. The ABS import calls `GetBook` once per matched item, so a 500 book library is 500 extra unpaced calls, and `getJSON` retries retryable statuses three times, so a 429 wave multiplies.

**Worse: the ABS importer would overwrite accurate per item data with work level approximations.** It merges `GetBook` output field by field through `applyConflictField`, and `release_date` and `language` are both in `bookConflictFields`. Today an OpenLibrary book always comes back with both empty, so the ABS value is kept and auto resolved. Fill them in and the branch flips to "values differ": the book row is overwritten and an unresolved conflict is filed. ABS reports `en`, the OpenLibrary search index reports `eng`, so that is a conflict on essentially every item that carries a language. And a work's `first_publish_year` approximated to January 1st would replace the edition's own publication date. An audiobook published 2019-05-14 of a 1996 work becomes 1996-01-01.

That second one is the real problem, and it is not a bug in the enrichment so much as a scoping error in where I called it. #2306 is about the Rebind path. The enrichment belongs there, not on every `GetBook` in the codebase.

**Also, two of the three tests pass against the pre-fix code.** `SearchEnrichmentIsBestEffort` and `SearchEnrichmentRejectsDifferentWork` both assert only that the fields are empty, which is what the old `GetBook` returns for any input, and neither asserts the search endpoint was reached. They need a hit counter.

One more thing I want to check before this lands: nothing here has ever exercised `q=key:/works/{id}` against the real API. Every test serves it from a mock that echoes back whatever doc the test wrote, so they prove nothing about whether OpenLibrary's search accepts that query shape at all. The unescaped slashes are meaningful to the query parser. If it does not work, the guard fires, the enrichment silently no-ops, #2306 is not actually fixed, and every `GetBook` still pays the request.

v1.33.3 ships without this. I will rework it against the Rebind path for v1.34.0 and verify the query by hand first.

PR #2310, opened 2026-09-03
fix(series): exclude hidden books from series fill and genre apply (#2302)

### LINKED ISSUE #2302: Book exclusion is ignored by every series query, and series fill re-queues excluded books
### Pre-flight

Reported by hoxtonia in Discord #support: "if i exclude a book in author view then that exclusion does not carry across to the series page. i think a book exclusion should be universal?"

### Current behaviour

`books.excluded` is honoured by every book listing query in `internal/db/books.go` (`List`, `ListForUser`, `ListByAuthor` at `books.go:365`, `ListByStatus`, the filtered list at `books.go:291`, and so on, all of which carry `WHERE ... excluded = 0`). No query in `internal/db/series.go` does.

The three series read paths:

* `ListBooksInSeries` (`series.go:282`) selects thirteen columns from `books` and `excluded` is not one of them. No `WHERE` beyond `sb.series_id = ?`.
* `ListWithBooksForUser` (`series.go:76`), behind `GET /api/series`, scopes on `owner_user_id` only.
* `GetByIDForUser` (`series.go:320`), behind `GET /api/series/{id}`, same.

Verified against a fresh in-memory database with one author, one series and two wanted books, the second of them excluded:

```
ListByAuthor returned 1 books:
   "The Eye of the World" excluded=false
ListBooksInSeries returned 2 books:
   "The Eye of the World" excluded=false status="wanted"
   "New Spring" excluded=false status="wanted"
ListWithBooksForUser series "The Wheel of Time" has 2 books:
   "New Spring" excluded=false
   "The Eye of the World" excluded=false
GetByIDForUser series "The Wheel of Time" has 2 books:
   "New Spring" excluded=false
   "The Eye of the World" excluded=false
```

Two separate problems in that output. The excluded book is returned, and it is returned with `excluded=false`, because the column is never selected. So the JSON the series page receives asserts the book is not excluded, and no client-side filter can recover the truth either.

### The part that is not cosmetic

`SeriesHandler.Fill` at `series.go:507` builds its work list from `ListBooksInSeries`, and `queueSeriesBook` (`series.go:565`) gates only on status:

```go
switch b.Status {
case models.BookStatusImported, models.BookStatusDownloading, models.BookStatusDownloaded:
    return false, models.Book{}, nil
}
if err := h.books.MarkWantedMonitored(ctx, b.ID); err != nil {
```

An excluded book sitting at `wanted` therefore gets `MarkWantedMonitored` called on it and is handed to `fanOutSeriesSearches` → `SearchAndGrabBook`. Series fill actively re-acquires the exact books the user excluded, which is the opposite of what the action means. Same shape as the series-fill bypass already noted on #2208.

`SeriesHandler` genre apply (`internal/api/genre_apply.go:108`) reads the same unfiltered list and writes genres onto excluded books too. Harmless by comparison, but it is the same missing filter.

### Expected behaviour

Exclusion should be universal, as the reporter puts it. Concretely:

1. `ListBooksInSeries` filters `excluded = 0`. That fixes series fill and genre apply at once, since both consume it for action rather than display.
2. `ListWithBooksForUser` and `GetByIDForUser` either filter `excluded = 0`, or select `b.excluded` and let the UI show excluded books greyed out the way the author view does. Worth deciding rather than assuming: on the author page an excluded book disappears, so filtering is the consistent answer, but a series with a gap at position 3 and no explanation is its own kind of confusing.

Either way `b.excluded` should be added to the selected columns so the API stops asserting `false` for books that are excluded.

### Steps to reproduce

1. Add an author whose books belong to a series.
2. On the author page, exclude one book. It disappears from the author view.
3. Open that series. The excluded book is still listed.
4. Press Fill on the series. The excluded book is marked wanted and monitored, and a search is dispatched for it.

### Deployment method

Code path is deployment independent. Confirmed by reading and exercising `main` at b67394fd.

thejdubb02:
Fixes #2302.

The bug:
ListBooksInSeries (internal/db/series.go) selects from books but never filters books.excluded, unlike every author-level listing in books.go (List, ListByAuthor, ListByStatus, and so on all carry excluded = 0). Two callers act on that list:
- SeriesHandler.Fill (internal/api/series.go) builds its acquisition work list from it, and queueSeriesBook gates only on status, so an excluded book at "wanted" gets MarkWantedMonitored then SearchAndGrabBook. Series fill re-acquires the exact books the user excluded, the opposite of what excluding means.
- The series genre apply (internal/api/genre_apply.go) writes genres onto excluded books too.

The fix:
Filter excluded = 0 in ListBooksInSeries so both action paths skip excluded books, matching the author-level queries. The one caller that must still see every book regardless of the flag is remainingSeriesBooksAfterRunRollback (internal/abs/import_rollback.go), which counts the books a rollback leaves behind to decide whether the series is now empty; an excluded book still keeps the series non-empty, so that path calls a new ListBooksInSeriesIncludingExcluded. This mirrors the existing ListByAuthor / ListByAuthorIncludingExcluded split.

Testing:
- Added TestSeriesListBooksExcludesHidden (internal/db/series_test.go): with one series and two linked wanted books, one excluded, ListBooksInSeries returns only the kept book and ListBooksInSeriesIncludingExcluded returns both. It fails without the filter (ListBooksInSeries returns both), so it pins the bug.
- go test ./internal/db/ ./internal/abs/ ./internal/api/ pass; go build ./..., gofmt, go vet and golangci-lint are clean.

Scope note: this fixes the non-cosmetic behaviour (fill re-acquiring excluded books, genre apply touching them). The two series-display reads, ListWithBooksForUser and GetByIDForUser, also return excluded books, and since excluded is not in their SELECT the API reports excluded=false for them. Whether the series page should hide those (consistent with the author view) or show them greyed out is a UI call, so I left it out of this change; happy to follow up whichever way you prefer.

### DIFF
--- a/internal/abs/import_rollback.go
+++ b/internal/abs/import_rollback.go
@@ -1044,7 +1044,9 @@ func (i *Importer) hasMatchingProvenanceOutsideRun(ctx context.Context, entity m
 }
 
 func (i *Importer) remainingSeriesBooksAfterRunRollback(ctx context.Context, seriesID int64, runOwnedBookIDs map[int64]struct{}) (int, error) {
-	books, err := i.series.ListBooksInSeries(ctx, seriesID)
+	// Count every book still in the series, including excluded ones: an excluded
+	// book left behind still keeps the series non-empty (#2302).
+	books, err := i.series.ListBooksInSeriesIncludingExcluded(ctx, seriesID)
 	if err != nil {
 		return 0, err
 	}
--- a/internal/db/series.go
+++ b/internal/db/series.go
@@ -278,14 +278,33 @@ func (r *SeriesRepo) UpdateTitle(ctx context.Context, id int64, title string) er
 	return nil
 }
 
-// ListBooksInSeries returns all books linked to the given series, with status.
+// ListBooksInSeries returns the books linked to a series, skipping books the
+// user has excluded. Callers that act on the result (series fill, genre apply)
+// must not touch excluded books, the same way the author-level listings in
+// books.go filter them (#2302).
 func (r *SeriesRepo) ListBooksInSeries(ctx context.Context, seriesID int64) ([]models.Book, error) {
+	return r.listBooksInSeries(ctx, seriesID, false)
+}
+
+// ListBooksInSeriesIncludingExcluded is ListBooksInSeries without the excluded
+// filter, for callers that must account for every book in the series regardless
+// of its excluded flag (e.g. counting how many books an import-run rollback
+// leaves behind, where an excluded book still keeps the series non-empty).
+func (r *SeriesRepo) ListBooksInSeriesIncludingExcluded(ctx context.Context, seriesID int64) ([]models.Book, error) {
+	return r.listBooksInSeries(ctx, seriesID, true)
+}
+
+func (r *SeriesRepo) listBooksInSeries(ctx context.Context, seriesID int64, includeExcluded bool) ([]models.Book, error) {
+	where := "WHERE sb.series_id = ? AND b.excluded = 0"
+	if includeExcluded {
+		where = "WHERE sb.series_id = ?"
+	}
 	rows, err := r.db.QueryContext(ctx, `
 		SELECT b.id, b.foreign_id, b.author_id, b.title, b.sort_title, b.status, b.monitored,
 		       b.image_url, b.release_date, b.language, b.media_type, b.created_at, b.updated_at
 		FROM series_books sb
 		JOIN books b ON b.id = sb.book_id
-		WHERE sb.series_id = ?`, seriesID)
+		`+where, seriesID)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/db/series_test.go
+++ b/internal/db/series_test.go
@@ -135,6 +135,86 @@ func TestSeriesLinkBook(t *testing.T) {
 	}
 }
 
+// TestSeriesListBooksExcludesHidden covers #2302: ListBooksInSeries must skip
+// excluded books so series fill and genre apply do not act on them, while
+// ListBooksInSeriesIncludingExcluded still returns them for callers that count
+// every book in the series (e.g. import-run rollback).
+func TestSeriesListBooksExcludesHidden(t *testing.T) {
+	database, err := OpenMemory()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer database.Close()
+
+	ctx := context.Background()
+	authorRepo := NewAuthorRepo(database)
+	bookRepo := NewBookRepo(database)
+	seriesRepo := NewSeriesRepo(database)
+
+	author := &models.Author{
+		ForeignID: "OL-2302-A", Name: "Robert Jordan", SortName: "Jordan, Robert",
+		MetadataProvider: "openlibrary", Monitored: true,
+	}
+	if err := authorRepo.Create(ctx, author); err != nil {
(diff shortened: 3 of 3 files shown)


thejdubb02:
The SAST - Frontend check is failing on `npm audit` in web/ (browserslist and @humanfs/node), which are pre-existing frontend dependencies unrelated to this Go-only change. Looks like #2309 already covers bumping them. The Go, CodeQL, dependency-review, and sign-off checks here are green.

vavallee (approved):
Approving. I checked this against the tree rather than taking it on trust and it holds up.

There are exactly three callers of `ListBooksInSeries` and you found all three. Series fill (`internal/api/series.go:507`) and genre apply (`internal/api/genre_apply.go:108`) both act on the list and both should skip excluded books; the rollback counter is the one that must not. Getting that third one right is the part I would have flagged if it had been missed: `remainingSeriesBooksAfterRunRollback` decides whether a series is now empty, and an excluded book still occupies the series, so counting only the visible ones would have had a rollback delete a series that still had a book in it. Swapping it to the explicit `IncludingExcluded` variant rather than leaving it on the filtered call is the right shape, and mirroring the existing `ListByAuthor` split means the next person reads it the same way.

Verified locally: build and vet clean, `internal/db`, `internal/abs` and `internal/api` all pass, and your test does pin the bug. I dropped just the `b.excluded = 0` clause with the test file left in place and `TestSeriesListBooksExcludesHidden` fails, so it is catching the thing it claims to.

One thing missing: this needs a `changelog.d/` fragment. `changelog.d/2310-series-excluded-books.md`, starting with `### Fixed`, one bolded lead sentence and user facing prose ending in `(#2302)`. See the two files already in there for the shape. Add that and I will merge.

On your scope question, and thanks for asking rather than guessing. Show them, do not hide them. The author view hides excluded books because that list is a catalogue you are curating down. A series is a fixed set of things, so hiding book 3 of 7 makes the series itself look wrong, and the user then has no way to see what they excluded or to undo it. But that is a real second bug and not just a UI preference: `excluded` is not in the SELECT for `ListWithBooksForUser` or `GetByIDForUser`, so the API reports `excluded=false` for every book on the series page and the frontend could not mark them even if it wanted to. Leave it out of this PR, I will file it separately.

Nice report to work from too. Filing the scope note yourself saved me working out whether you had missed those two reads or left them deliberately.

vavallee:
Merging this without the fragment, so you are not left waiting on a formality for a fix that is already correct.

We are cutting v1.33.3 now and this is in it. The fragment convention exists to keep PRs off `CHANGELOG.md` so they do not all conflict on one file; at release time I write the section by hand anyway, so I have folded your entry straight into the assembly and credited you there. Nothing lost, one less round trip.

For your next one, still add the fragment. It only stops mattering when a release happens to be going out the same day.

Thanks for this. The rollback caller was the part that made it worth reviewing carefully rather than rubber stamping, and you got it right without being told.

vavallee:
Filed the display half as #2324, as promised.

Turns out it is worse than "the series page cannot show an excluded badge". Because `excluded` is never selected, an excluded book still counts toward the amber "N missing" pill on the series card, which enables Fill, and Fill then correctly skips it thanks to your change and returns `queued: 0`. So the UI now prints "Nothing to fill" right next to a badge that says "1 missing". Your fix is what made that contradiction visible, which is the right order to fix them in.

Shipped in v1.33.3. Thanks again.

PR #2317, opened 2026-09-03
fix(authors): only save latin alternate-name aliases for a non-latin author (#2268)

### LINKED ISSUE #2268: saveAlternateNames writes OpenLibrary alternate names as aliases with no provenance, minting other authors' names
### Summary

`saveAlternateNames` (`internal/api/authors.go:3402`) writes every all-ASCII entry of
OpenLibrary's `alternate_names` into `author_aliases` with **no provenance**: the row is
created as `&models.AuthorAlias{AuthorID: author.ID, Name: name}`, leaving `SourceOLID`
empty. For a pseudonym or collaboration record, those alternate names are the real
people, so adding one author silently mints alias rows claiming other authors' names.

This is the root of the collapse reported in #1684. That issue is fixed on the
**consumption** side in v1.32.2 (`aliasBindsAuthor`, `internal/calibre/importer.go:765`,
refuses to trust an alias that nothing backs up), so the rows are currently inert. This
issue is about not writing them in the first place.

### Why it looks correct and is not

The function exists for transliteration: it lets "Murakami" on a release name reach an
author stored as "村上春樹", which is a real requirement. The filter it uses for that is
`isAllASCII(name)`, applied to the alias.

That test never looks at the author. It is intended to mean "this is a latin-script
rendering of a non-latin name", but it is true of *any* ASCII string, so for an author
whose own name is already ASCII it admits everything OpenLibrary lists. A pen name's
alternate names are exactly that: the collaborators' real names.

`aliasBindsAuthor` already encodes the correct test for this on the read side:

```go
return !isAllASCII(canonical.Name) && isAllASCII(alias.Name)
```

The write side never got the `!isAllASCII(canonical.Name)` half.

### Evidence

From #1684, @paulohenriquesg's library (~1,550 books, one title credited to the joint
pseudonym "Kem Antilles", pen name of Kevin J. Anderson and Rebecca Moesta):

- author row created `2026-08-18T06:20:16.202Z`
- alias "Kevin J. Anderson" at `06:20:16.220Z`, alias "Rebecca Moesta" at `06:20:16.226Z`,
  both `source_ol_id` NULL, roughly 20 ms after author creation
- reproduced after editing the Calibre book to credit **only** "Kem Antilles", which is
  what identifies enrichment rather than co-author credits as the source

### Proposed fix

Apply the same test the read side uses. An alternate name should only be persisted when
it plausibly is a transliteration, which at minimum means the canonical name is not
already latin-script:

- gate the write on `!isAllASCII(author.Name)`, mirroring `aliasBindsAuthor`
- and/or stamp `SourceOLID` on rows written from enrichment, so the provenance the read
  side looks for actually exists and enrichment aliases become distinguishable from
  hand-created ones

Worth deciding explicitly whether an enrichment-sourced alias should count as
"backed up" at all. Stamping `SourceOLID` would make these rows bind again under the
current `aliasBindsAuthor`, which would reintroduce #1684 unless the write is narrowed
at the same time. If both changes are made, narrow the write first.

### Not in scope

Cleaning up existing rows. There is no migration and deletion is manual, per alias, from
the author page. That is tracked in the discussion on #1684 and should be decided
separately, since anything automatic is deleting user-visible data on an assumption.

### Acceptance

An author whose own name is latin-script gains no aliases from enrichment. A non-latin
author still gains its latin-script alternate names, so the "Murakami" case in the
existing doc comment keeps working. Both covered by tests.

Filed off #1684, credit to @paulohenriquesg for the timing evidence that isolated the
write path.

thejdubb02:
## Summary

Fixes #2268. `saveAlternateNames` persists an author's OpenLibrary "alternate names" as `author_aliases` rows, keeping only the all-ASCII (latin) ones, so a non-latin primary name (for example 村上春樹) can be reached by a latin release name (Murakami). It never checked the author's own name, so a latin-script author (a pen name, or a two-author collaboration credit) had other real authors' alternate names minted as its aliases.

The read side already guards exactly this: `aliasBindsAuthor` (internal/calibre/importer.go) binds an unattributed latin alias only when the canonical name is non-latin (`!isAllASCII(canonical.Name) && isAllASCII(alias.Name)`). This adds the missing write-side half: `saveAlternateNames` returns early when the author's own name is all-ASCII. The latin same-name variants that legitimately bind (initials, comma-sort, punctuation) are written elsewhere with provenance (`recordAuthorCreateAlias` stamps a `SourceOLID`), so nothing that should bind is lost, only the unattributed OL dump for latin authors is dropped.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` (no env, config, or upgrade change)
- [ ] Wiki (no new user-facing surface, only the incorrect-alias behaviour is fixed)

## Test plan

Backend-only change, no frontend touched.

- `go build ./...` and `go vet ./internal/api/` clean, `gofmt` clean.
- `go test ./internal/api/ ./internal/calibre/` pass (the affected packages, including the read-side `aliasBindsAuthor`).
- Added `TestSaveAlternateNames_OnlyBindsNonLatinAuthor`: a latin-script author with two collaborator names as alternates mints zero aliases (this fails without the guard, minting both names), and a non-latin author still gets its latin alternate names, so the feature it exists for is preserved.

### DIFF
--- a/changelog.d/2268-author-alias-latin-guard.md
+++ b/changelog.d/2268-author-alias-latin-guard.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Author aliases** (#2268): a latin-script author (a pen name, or a two-author collaboration credit) no longer has other real authors' OpenLibrary alternate names saved as its aliases. Alternate-name aliases are kept only for authors whose primary name is non-latin, which is the case they exist for.
--- a/internal/api/authors.go
+++ b/internal/api/authors.go
@@ -3484,6 +3484,17 @@ func (h *AuthorHandler) saveAlternateNames(ctx context.Context, author *models.A
 	if h.aliases == nil || len(author.AlternateNames) == 0 {
 		return
 	}
+	// These latin-script aliases exist only so a non-latin primary name can be
+	// reached by a latin release name (e.g. "Murakami" -> "村上春樹"). For a
+	// latin-script author every alternate name is just another latin name, and
+	// minting it as an alias would file unrelated real authors (a pen name, or a
+	// co-author credit) under this one. Only save them when the author's own name
+	// is non-latin, mirroring the read-side guard in aliasBindsAuthor
+	// (importer.go), which binds an unattributed latin alias only when the
+	// canonical name is non-latin.
+	if isAllASCII(author.Name) {
+		return
+	}
 	for _, name := range author.AlternateNames {
 		if !isAllASCII(name) {
 			continue
--- a/internal/api/authors_test.go
+++ b/internal/api/authors_test.go
@@ -8071,3 +8071,65 @@ func TestFetchAuthorBooks_StillWidensAnUnownedBook(t *testing.T) {
 		t.Errorf("mediaType = %q, want both — an unowned book still merges", books[0].MediaType)
 	}
 }
+
+// TestSaveAlternateNames_OnlyBindsNonLatinAuthor is the regression test for
+// #2268: saveAlternateNames minted every all-ASCII OpenLibrary alternate name
+// as an alias without checking the author's own name, so a latin-script author
+// (a pen name, or a two-author collaboration credit) had unrelated real
+// authors' names filed under it as aliases. Latin aliases must only be saved
+// when the author's primary name is non-latin, mirroring aliasBindsAuthor on
+// the read side.
+func TestSaveAlternateNames_OnlyBindsNonLatinAuthor(t *testing.T) {
+	database, err := db.OpenMemory()
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { _ = database.Close() })
+
+	ctx := context.Background()
+	authorRepo := db.NewAuthorRepo(database)
+	aliasRepo := db.NewAuthorAliasRepo(database)
+	h := &AuthorHandler{authors: authorRepo, aliases: aliasRepo}
+
+	// A latin-script author (here a pen name) whose OpenLibrary "alternate
+	// names" are in fact other real authors. None of these may become aliases.
+	penName := &models.Author{
+		ForeignID:      "ol:kem-antilles",
+		Name:           "Kem Antilles",
+		SortName:       "Antilles, Kem",
+		AlternateNames: []string{"Kevin J. Anderson", "Rebecca Moesta"},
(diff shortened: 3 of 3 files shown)


vavallee:
Merging. I checked the premise rather than taking the description, and it holds all the way down.

Verified the write side really does mint junk: reverting your guard and running your test gives

```
latin-script author minted 2 aliases, want 0:
  [{Name:Rebecca Moesta SourceOLID:} {Name:Kevin J. Anderson SourceOLID:}]
```

Two real, separate authors filed as aliases of a third. That is the bug stated better than either of us stated it.

Your provenance argument checks out too. `saveAlternateNames` creates rows with no `SourceOLID`, while `recordAuthorCreateAlias` stamps `SourceOLID: author.ForeignID` on the same-name variants that legitimately bind, so the early return drops only the unattributed dump and nothing that should match.

One thing I went looking for and want to record, because it changes how urgent this is rather than whether it is right. I expected these rows to be actively harmful on the search path, since `internal/api/indexers.go:428` and `internal/scheduler/scheduler.go:749` both read every alias through `ListByAuthor` and feed the names straight into `MatchCriteria.AuthorAliases` with no `aliasBindsAuthor` call anywhere near them. They are not harmful, because `filterRelevant` guards it independently at `internal/indexer/searcher.go:777`:

```go
authorTokenSets := [][]string{authorTokens(author)}
if !isAllASCIILower(surname) {
    for _, alias := range aliases {
```

Aliases only enter the token sets when the surname is non latin. So a latin author's junk aliases were being written, stored, and then ignored by both readers. Dead rows rather than wrong matches.

Worth knowing that the two read guards do not agree on the rule: `aliasBindsAuthor` and your new write guard both test the full name, `filterRelevant` tests the surname alone. They diverge for a name like "村上 Haruki" where the surname is latin and the given name is not. Pre-existing, not yours, and I am not asking you to touch it.

Two follow ups I will file rather than hold this up for:

- Rows already written stay written. They are inert for the reasons above, so cleanup is hygiene rather than a fix, but somebody should decide whether a migration clears them.
- The three-guard divergence deserves one shared helper so the next person does not have to work out which rule applies where.

Good change, and thanks for including the read-side reasoning in the description. That is what made it quick to check.

Issue #2154, opened 2026-08-21
search observability: scheduled and bulk paths are silent, return before completion, and leave last-debug stale


SturmB:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

*Split from #1634 at the maintainer's request. This keeps search observability separate from the missing-format wanted query.*

### Current behaviour

Search observability differs by initiation path, and the two non-interactive paths provide no reliable evidence that work ran:

1. **Scheduled and bulk searches emit no search log line.** A search that returns no results is indistinguishable from a search that never ran. The first visible line may be `auto-grabbing book`, and that exists only when a release was found and approved.
2. **Bulk search reports success before doing the work.** `POST /api/v1/book/bulk` with `action: "search"` returns `{"results":{"N":{"ok":true}}}` immediately, while the actual indexer sweep continues asynchronously and may auto-grab a release later. In the reproduction attached to #1634, the response arrived first and the grab appeared roughly 18 seconds later, so `ok:true` means accepted/queued rather than completed successfully.
3. **`GET /api/v1/search/last-debug` is incomplete and easy to read from the wrong bucket.** `lastDebug.set` is called by the interactive `POST /book/{id}/search` handler, but scheduled and bulk searches never update it. The state is also partitioned by user id: a browser search writes the signed-in user's bucket, while an API-key request reads user 0. Reading the endpoint through an API key after clicking Search in the browser can therefore return an unrelated older search even though the interactive path did update its own bucket.

The third point corrects the mechanism stated in #1634: the UI does use the interactive handler. The remaining defect is that scheduled/bulk paths do not populate the debug state, and different auth contexts silently read different partitions.

### Expected behaviour

All search initiation paths should leave an accurate, attributable diagnostic trail:

- log search start/completion at `INFO`, including initiation path, term/book, requested media type, indexer count, raw-result count, shown/approved count, and elapsed time;
- make the bulk response describe acceptance/queueing rather than completed success, or expose a job/result identifier that can be checked after the asynchronous sweep finishes;
- either populate debug state from scheduled and bulk searches too, or replace the single “last debug” view with a model that identifies initiation path and auth/user scope so an old or cross-bucket payload cannot be mistaken for the search just triggered.

### Steps to reproduce

1. Set Runtime level to DEBUG.
2. Trigger a bulk search with `POST /api/v1/book/bulk` and `action: "search"`.
[...]

PR #2326, opened 2026-09-04
feat(scheduler): log every automatic search, with what started it (#2154)

vavallee:
Addresses point 1 of #2154. Points 2 and 3 are confirmed but deliberately not in this PR — see below.

## What was wrong

I traced all three points in @SturmB's report against `main`. All three are real, and the mechanism in each case is exactly as described.

**Point 1, the one this PR fixes.** `Scheduler.searchAndGrabFormat` had three INFO lines and every one of them is conditional:

- `"searching with an incomplete indexer pool"` — only when an indexer hard-failed
- `"skipping a multi-book pack for a single-book search"` — only when a pack was a candidate
- `"auto-grabbing book"` — only on success

A clean no-result search fell through `if best == nil { ... return }` and logged nothing whatsoever. So "the sweep never ran" and "the sweep ran and found nothing" were genuinely indistinguishable, which is what the report says.

## The fix

One INFO line on **every** return path out of `searchAndGrabFormat`, carrying the fields the report asked for:

```
level=INFO msg="book search finished" origin=bulk book=Dune book_id=42 format=ebook
  indexers=3 raw_results=17 after_filters=4 approved=1 outcome=grabbed elapsed_ms=2841
```

`outcome` distinguishes the ways a search can end: `no results`, `nothing approved`, `grabbed`, `already grabbed`, `no download client for protocol`, and the error paths. The counters fill in as the search progresses, so an early return leaves them at zero — an honest reading of how far it got rather than a fabricated one.

### Why the origin is on the context

`origin` is the field that actually answers "did the thing I just clicked run?", and it is the awkward one: the call chain crosses three packages and the API handlers hold a `BookSearcher` interface, not a `*Scheduler`, so a parameter would have to be threaded through an interface every caller implements. The bounded pools in `bulk.go`, `series.go` and `authors.go` already carry their own context into the search goroutine, so the context is the thing that is already going where the value needs to go.

`indexer.WithSearchOrigin` / `indexer.SearchOriginFrom`, same shape as the existing context-carried values in this codebase. Eight production call sites, all tagged:

| origin | site |
| --- | --- |
| `scheduled` | the periodic wanted sweep |
| `bulk` | `POST /api/v1/book/bulk` and `/wanted/bulk` |
| `series-fill` | series Fill |
| `author` | author auto-search and add-with-search |
| `book` | a book flipping to wanted |
| `recommendation` | an accepted recommendation |
| `requeue` | the re-search after a stalled download is blocklisted |
| `unknown` | the fallback, a real value so a gap reads as a gap |

`TestEverySearchCallSiteTagsItsOrigin` asserts that structurally by reading the source, so a fan-out added later fails the suite instead of quietly logging `origin=unknown`.

## Verification

Five tests, and the log capture runs at `LevelInfo` on purpose — #2154 asks for evidence at the level an operator actually runs at, so a line demoted to Debug has to fail rather than pass.

Negative controls, all three run:

| control | result |
| --- | --- |
| demote the line to `slog.Debug` | 3 tests fail |
| remove the line entirely | 4 tests fail |
| untag one call site | the drift guard fails, naming the file and line |

Also asserted: a `both` book logs one line per format, not one for the pair, and the auto-grab kill switch keeps its own dedicated line without claiming a search ran.

## What is not here, and why

Points 2 and 3 are confirmed but are API contract changes, not logging, and each needs a decision I did not want to bury in this PR.

**Point 2 — bulk reports success before doing the work.** `BulkHandler.BooksBulk` sets `bulkItemResult{OK: true}` per id in the request loop, then calls `fanOutSearches` after it, and that helper is documented fire-and-forget by design. So `ok:true` means accepted, exactly as the report says. Fixing it means either renaming the field's meaning (a breaking response change) or introducing a job id the caller can poll. That is a design decision, not a bug fix.

**Point 3 — `last-debug` is partitioned and incompletely written.** `lastDebugStore` is keyed by user id and `set` is called from exactly one place, the interactive `SearchBook` handler. `LastSearchDebug` reads `auth.UserIDFromContext`, which is 0 for an API key. So an API-key reader and a browser session are looking at different buckets, and the scheduled and bulk paths write to neither. The correction in the report about the UI using the interactive handler is right.

The awkward part is that the store lives on `IndexerHandler` in `internal/api` while the searches that need to write to it run in `internal/scheduler`, so populating it from those paths means either moving the store or handing the scheduler a writer interface. That is worth doing and worth doing deliberately.

@SturmB — this gets you the evidence trail. Does the completion line carry what you need for the scheduled and bulk cases, and would you rather see point 2 solved by a job id you can poll or by the response field just saying "queued" honestly?

### DIFF
--- a/changelog.d/2154-search-observability.md
+++ b/changelog.d/2154-search-observability.md
@@ -1,0 +1,2 @@
+### Added
+- **Every automatic search now logs when it finishes** (#2154) — a scheduled sweep, a bulk search or a series fill that found nothing used to log nothing at all, so "the search never ran" and "the search ran and found nothing" looked identical from the log. Each one now emits a single INFO line naming what started it, the book and format, how many indexers were queried, how many results came back, how many survived filtering, and how long it took. Thanks to SturmB for the report.
--- a/internal/api/authors.go
+++ b/internal/api/authors.go
@@ -2740,7 +2740,7 @@ func runBookSearches(ctx context.Context, searcher BookSearcher, books []models.
 	// once as slots free up (#1515); shares the package pace with the bulk
 	// "search all" and series-fill fan-outs.
 	concurrency.RunBoundedPaced(ctx, books, maxConcurrent, searchPaceInterval, func(ctx context.Context, book models.Book) {
-		searcher.SearchAndGrabBook(ctx, book)
+		searcher.SearchAndGrabBook(indexer.WithSearchOrigin(ctx, indexer.OriginAuthor), book)
 	})
 }
 
@@ -3339,7 +3339,7 @@ func (h *AuthorHandler) AddBook(w http.ResponseWriter, r *http.Request) {
 	// context so the search goroutine is cancelled on shutdown rather than
 	// running against context.Background(). See #846.
 	if req.SearchOnAdd && h.searcher != nil {
-		go h.searcher.SearchAndGrabBook(h.bgCtx(), *book) // #nosec G118 -- intentional: search must outlive the request
+		go h.searcher.SearchAndGrabBook(indexer.WithSearchOrigin(h.bgCtx(), indexer.OriginAuthor), *book) // #nosec G118 -- intentional: search must outlive the request
 	}
 
 	writeJSON(w, http.StatusCreated, book)
--- a/internal/api/books.go
+++ b/internal/api/books.go
@@ -17,6 +17,7 @@ import (
 	"github.com/vavallee/bindery/internal/bookhydrate"
 	"github.com/vavallee/bindery/internal/db"
 	"github.com/vavallee/bindery/internal/importer"
+	"github.com/vavallee/bindery/internal/indexer"
 	"github.com/vavallee/bindery/internal/metadata"
 	"github.com/vavallee/bindery/internal/models"
 	"github.com/vavallee/bindery/internal/textutil"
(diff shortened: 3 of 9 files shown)


vavallee:
Two red checks, both accounted for:

**`Container Scan`** is failing on `main` too, not on this branch. Grype picked up a vulnerability DB built this morning (`v6.1.9`, 2026-09-04T06:30) carrying two new high alerts against `golang.org/x/crypto v0.55.0`, created 2026-09-03. Neither this PR nor #2325 touches `go.mod`, `go.sum` or the `Dockerfile`. `govulncheck` says none of the three advisories in that module are reachable from our code, so the practical risk is low, but grype is reachability-blind and will keep failing every PR until the module is bumped. `v0.56.0` is out and no Dependabot PR covers it. Filing that separately.

**`codecov/patch`** is measuring changed lines that are error branches: `outcome = "download client lookup failed"` and friends, plus the one-line context wrappers at the API call sites. Those branches had no coverage before this PR either, and reaching them means standing up a failing indexer repo and a failing download-client repo purely to assert a log string. `origin.go` is at 100% and `SearchAndGrabBook` at 100%; `searchAndGrabFormat` sits at 62.4%, which is where it already was.

Neither check is in the required set, which is `lint`, `validate (Go)` and `Security Summary`.

SturmB:
Yes, the completion line carries what I need for diagnosing scheduled and bulk searches. Thank you for confirming and separating the remaining contract questions.

For point 2, I would prefer a job ID that can be polled, with the initial response explicitly reporting the work as queued/accepted rather than completed. I am consuming this from automation, so simply renaming the response state would be more accurate but would still leave the caller sleeping and guessing when the sweep has finished.

The automated review’s suggested distinction between `filtered out` and `nothing approved` also sounds useful if it can be added without expanding the scope much.

Issue #2360, opened 2026-09-05
Frontend polling keeps running in a hidden tab, and WantedPage restarts its interval on every hover


vavallee:
Line numbers are `main` at `7917b7ec`.

## What is wrong

Three pages poll on a fixed 5 s `setInterval` and none of them stop when the tab is hidden:

- `web/src/pages/QueuePage.tsx:60`, two requests per tick (`listQueue` plus `listPending`)
- `web/src/pages/WantedPage.tsx:56`, one `listWanted` per tick
- `web/src/pages/settings/LogsTab.tsx:143`, one `logs` query per tick while auto refresh is on

`visibilitychange` appears exactly once in the whole frontend, in `AuthContext.tsx:39`, and none of these three use it. Leave the Queue tab open in a background window overnight and it makes about 1400 requests an hour against a server that has nothing new to say.

`WantedPage` has a second problem. Its effect deps are `[showExcluded, showResults, grabbingGuid, searchingId]`, so every hover-open of a results panel and every grab or search tears the interval down and starts a fresh 5 s window:

```ts
const interval = setInterval(() => {
  if (showResults !== null || grabbingGuid !== null || searchingId !== null) return
  ...
}, 5000)
return () => { cancelled = true; clearInterval(interval) }
}, [showExcluded, showResults, grabbingGuid, searchingId])
```

The guard inside the callback already handles the mid interaction case, so the deps only need to gate the interval, not restart it.

## Fix shape

A small `usePolling(fn, intervalMs)` hook that:

- pauses while `document.hidden` and resumes on `visibilitychange`
- keeps `fn` in a ref so unrelated state changes do not recreate the timer

Use it in the three pages above. Test with fake timers: ticks while visible, no ticks while hidden, one immediate catch up tick on becoming visible again.

This is the cheap version of #1163 (push updates over SSE), not a replacement for it. #1163 stays the real fix.

## Sequencing

Merge after the download state table PR and the `formatBytes` PR, both of which touch `QueuePage.tsx`. Next minor, not patch release material.

Issue #2359, opened 2026-09-05
22 raw window.confirm calls beside the unused ConfirmDialog component, four hardcoded in English


vavallee:
Line numbers are `main` at `7917b7ec`.

## What is wrong

22 raw `window.confirm` calls across 13 files, next to a `web/src/components/ConfirmDialog.tsx` that only `BookDetailPage.tsx` uses.

```
settings/AuthSettings.tsx:30      components/MergeAuthorsModal.tsx:55
pages/UsersPage.tsx:62           pages/AuthorsPage.tsx:134, 160, 208, 228
pages/AuthorDetailPage.tsx:267, 278, 322
pages/QueuePage.tsx:234, 260     pages/BooksPage.tsx:120
pages/SeriesPage.tsx:112         pages/settings/BlocklistTab.tsx:37
pages/settings/ABSTab.tsx:334    pages/settings/GeneralTab.tsx:749, 763
pages/settings/LogsTab.tsx:117   pages/settings/MetadataTab.tsx:298
pages/settings/QualityTab.tsx:119 pages/settings/IndexersTab.tsx:344
```

Four of them hardcode English and never reach i18n at all:

- `SeriesPage.tsx:112` `Delete "..." from Series? Linked books will stay in your library.`
- `LogsTab.tsx:117` `Delete backup ...?`
- `GeneralTab.tsx:749` `Regenerate the API key? ...`
- `GeneralTab.tsx:763` `Rotate the session signing secret? ...`

## What a user sees

A native browser dialog with the origin in the title bar, unstyled, untranslatable in those four cases, and blocking. It also cannot render the file lists and emphasis that `ConfirmDialog` was built for, which is why the book detail delete path grew its own modal and everything else did not.

## Fix shape

Route the call sites through `ConfirmDialog`. The component currently forces an "I understand" acknowledge checkbox, which is right for an irreversible file delete and too heavy for "delete backup X", so make `acknowledgeLabel` optional and skip the checkbox when it is absent. A small promise returning hook around it keeps each call site close to the shape it has now:

```ts
if (!(await confirm({ title, body, confirmLabel }))) return
```

Move the four hardcoded English strings into `en.json` keys while doing it. `en.json` only, other locales fall back.

## Sequencing

Merge after the download state table PR and the `formatBytes` PR, both of which touch `QueuePage.tsx`. Mechanical but wide, so worth landing on its own. Next minor, not patch release material.

Issue #2350, opened 2026-09-05
Seven copies of the byte formatter in web/src, four different behaviours


vavallee:
Line numbers are `main` at `7917b7ec`. There is no `web/src/util/format.ts`.

## What is wrong

Seven copies of a byte formatter across the frontend, in four different behaviours:

| Where | Name | Behaviour |
|---|---|---|
| `SearchPage.tsx:6`, `HistoryPage.tsx:56`, `BookDetailPage.tsx:23` | `formatSize` | `<= 0` gives `''`; GB one decimal, MB zero decimals |
| `QueuePage.tsx:281`, `WantedPage.tsx:184` | `formatSize` | `>` not `>=` so exactly 1 GiB reads as MB; MB one decimal; zero reads `0 KB` |
| `settings/RootFoldersTab.tsx:8` | `formatBytes` | log based units, no clamp |
| `settings/LogsTab.tsx:8` | `formatBackupSize` | log based units with a `Math.min` clamp, `<= 0` gives `0 B` |

Three visible consequences:

1. The same 800 MB release reads `800 MB` on Search and History but `800.0 MB` on Queue and Wanted.
2. Queue and Wanted render `0 KB` for a zero size, where Search, History and Book detail render nothing.
3. `RootFoldersTab` has no unit clamp, so a root folder with 1 PB or more free renders `1.1 undefined`:

```ts
const units = ['B', 'KB', 'MB', 'GB', 'TB']
const i = Math.floor(Math.log(bytes) / Math.log(1024))
return `${(bytes / Math.pow(1024, i)).toFixed(1)} ${units[i]}`
```

`LogsTab` got the clamp right, `RootFoldersTab` did not, and neither one knows about the other five.

## Fix shape

One `web/src/util/format.ts` exporting `formatBytes(n: number): string`: `<= 0` returns `''` for the release size callers, log based units with the `Math.min` clamp, one decimal above KB. Keep a second entry point for the two callers that want `0 B` rather than an empty string (`RootFoldersTab`, `LogsTab`). Replace all seven copies. Unit tests for the boundaries: 0, 1023, 1024, 1.5 GB, 1 PB.

## Sequencing

Merge after the download state table PR, which also touches `QueuePage.tsx`. Not patch release material, this is next minor cleanup. Touches seven files, so worth merging on its own.

Issue #2342, opened 2026-09-05
Download state labels are four partial hardcoded lists in QueuePage instead of one shared table


vavallee:
Line numbers are `main` at `7917b7ec`.

## What is wrong

Book statuses were consolidated into `web/src/components/bookStatus.ts` so every surface agrees on one table. Download states never got the same treatment. `QueuePage.tsx` carries four partial, hand maintained lists of the same enum:

- `statusLabels` (`:197-206`), nine hardcoded English strings, no i18n at all
- `statusChip` (`:211-220`), nine chip classes
- `FAILED_STATUSES` (`:222`), three states
- `isMatchable` (`:228`), two states

```ts
const statusLabels: Record<string, string> = {
  grabbed: 'Grabbed',
  downloading: 'Downloading',
  ...
```

Two of those lists are already wrong. They stop at `importBlocked` while `internal/models/download_state.go` defines eleven states, so `importExternal` and `importHeld` render as raw enum text (#2339). And `retryAllFailed` (`:244`) hardcodes a fifth list of one state, which is the bug in #2336.

The English strings also bypass i18n keys that already exist: `history.events.grabbed` and `history.events.importFailed` are translated in eight locales and say exactly the same thing.

## What a user sees

The queue page is English only for every status chip, in an app that ships eight locales. Plus the two symptoms already filed as #2339 and #2336.

## Fix shape

One `web/src/components/downloadStatus.ts` in the shape of `bookStatus.ts`: a table keyed by every state string the Go side defines, each entry carrying an i18n label key, a chip class, and flags for failed / retryable / matchable. Derive `statusLabels`, `statusChip`, `FAILED_STATUSES`, `isMatchable` and the retryable filter from it. Reuse `history.events.*` where the key exists, add `queue.status.*` for the rest, to `en.json` only so the other locales fall back to English rather than carrying filler.

Add a test that the table covers every state in `download_state.go`, so the next state added on the Go side fails CI instead of shipping a grey chip.

## Sequencing

Same PR as #2339 and #2336, all three are this table. Patch release material because of the two symptoms it carries.

PR #2377, opened 2026-09-05
fix(web): drive the queue chips from one download-state table (#2342, #2339, #2336)

### LINKED ISSUE #2336: "Retry all failed" skips Import Blocked rows the retry endpoint accepts, while the count beside it includes them
Line numbers are `main` at `c6895aed`. Small, but it makes the queue lie about what it did.

## What is wrong

The backend retry accepts two states. `DownloadRepo.ResetImportRetry` (`internal/db/downloads.go:184`) updates `WHERE id=? AND status IN (?, ?)` with `StateImportFailed` and `StateImportBlocked`, so a blocked row is retryable through `POST /queue/{id}/retry-import`.

The bulk button does not. `retryAllFailed` (`web/src/pages/QueuePage.tsx:244`) filters to one status:

```ts
const retryable = queue.filter(q => q.status === 'importFailed')
```

and its visibility condition at `:339` is the same single check.

Meanwhile the count sitting next to that button is built from three states: `FAILED_STATUSES = new Set(['failed', 'importFailed', 'importBlocked'])` (`web/src/pages/QueuePage.tsx:222`).

## What the user sees

A bar reading "3 failed" with a **Retry all failed** button that quietly acts on one of them and reports nothing about the other two. The row-level story is inconsistent with itself as well: `isMatchable` (`:228`) already treats `importFailed` and `importBlocked` as the same class for the match action, so the two states are siblings everywhere except here.

The stale comment at `:231` says "Retry only applies to importFailed", which was true of an earlier backend. The repository widened and the frontend did not follow.

## Repro

1. Get one download into Import Failed and one into Import Blocked.
2. The bar reads "2 failed".
3. Press Retry all failed. Only the Import Failed row is retried, with no message about the other.
4. Press Retry import on the blocked row individually. It is accepted, which proves the bulk path excluded a row the endpoint would have taken.

## Suggested fix

Widen the filter and the visibility check to the states `ResetImportRetry` actually accepts, and derive both from one predicate rather than repeating the literal:

```ts
const isRetryable = (status: string) => status === 'importFailed' || status === 'importBlocked'
```

`isMatchable` at `:228` is already exactly this predicate, so either reuse it under a clearer name or define the retryable set beside `FAILED_STATUSES`.

That still leaves plain `failed` counted in the badge but not retryable, which is correct: a grab that never produced a file has nothing on disk to retry the import of. Worth making the button label or the bar say so, since "3 failed, retry all failed" reading as "retry 1 of these 3" is the actual complaint. Related: #2295 asks for a retry on that plain `failed` case, which is the missing capability rather than this mismatch.

## Origin

Found while answering #2295.

### LINKED ISSUE #2339: Queue shows a raw enum chip for importExternal and importHeld
Line numbers are `main` at `7917b7ec`.

## What is wrong

`internal/models/download_state.go` defines eleven download states. The queue UI only knows nine. `statusLabels` (`web/src/pages/QueuePage.tsx:197-206`) and `statusChip` (`:211-220`) both stop at `importBlocked`, so `importExternal` and `importHeld` fall through to the `?? item.status` and `?? 'bg-slate-200 ...'` fallbacks at `:429-430`.

```ts
<span className={`... ${statusChip[item.status] ?? 'bg-slate-200 dark:bg-zinc-800 ...'}`}>
  {statusLabels[item.status] ?? item.status}
</span>
```

## What a user sees

A grey chip reading the raw enum, `importExternal` or `importHeld`. Both states are deliberately non-terminal with no automatic exit (see the doc comments on `StateImportExternal` and `StateImportHeld`), so the row sits in the queue indefinitely with a label that means nothing to anyone who has not read the Go source. Nothing on the page says the file is waiting on the external import tool, or that it is parked until the paired format shows up.

## How to trigger

Set `import.mode=external` and grab anything. The download settles in `importExternal` and stays there. For `importHeld`, set a book to `media_type=both` with pair gating on and let one format finish first.

## Fix shape

Give both states a real label, a chip colour, and a line of explanation. Since this is the same table `statusChip`, `statusLabels`, `FAILED_STATUSES` and `isMatchable` all partially duplicate, the natural home is one shared table keyed by every state the Go side defines, in the shape of `web/src/components/bookStatus.ts`.

## Sequencing

Lands together with the download-state table refactor and #2336 in one PR, since all three are the same table. Patch release material: it is a user visible wrong result on the queue page, and it is the visible surface of the pair gating hold.

vavallee:
The queue page kept four partial hand maintained copies of the download state enum (`statusLabels`, `statusChip`, `FAILED_STATUSES`, `isMatchable`) plus a fifth inline literal in `retryAllFailed`, and two of them had already fallen behind `internal/models/download_state.go`. This replaces all five with one table in `web/src/components/downloadStatus.ts`, in the shape of the existing `bookStatus.ts`, keyed by every state the Go side defines and carrying an i18n label key, a chip class, an optional explanation and the failed / retryable / matchable flags.

Three user visible fixes fall out of it:

* `importExternal` and `importHeld` get a real label, a chip colour and a line saying why the row is parked, instead of a grey pill reading the raw enum. Both are non terminal by design so those rows can sit in the queue indefinitely.
* **Retry all failed** now retries every state `DownloadRepo.ResetImportRetry` accepts (`importFailed` and `importBlocked`), so a blocked row is no longer skipped by a button whose neighbouring count included it. A plain `failed` grab produced no file, so it correctly stays out.
* The chip labels go through i18n instead of nine hardcoded English strings. `grabbed` and `importFailed` reuse the `history.events.*` keys that already say the same thing in Korean, the one other locale that translates them; the rest are new `queue.status.*` keys added to `en.json` only, so the other locales fall back to English rather than carrying filler.

Part of #2342, which stays open for the drift guard: the coverage test here is a hand copied list, not a real pin against `download_state.go`.
Closes #2339
Closes #2336

## Tests

New `web/src/components/downloadStatus.test.ts` asserts the table covers exactly the state list in `download_state.go` (the Go list is copied in with a comment pointing at the file), that the derived failed and retryable sets match `ResetImportRetry`, and that the two hand off states resolve to a label rather than the raw string.

Three new cases in `QueuePage.test.tsx` are the fails before / passes after ones:

* `retries every state the retry endpoint accepts, not just importFailed (#2336)`
* `offers Retry all failed when only blocked imports are queued (#2336)`
* `labels importExternal and importHeld instead of showing the raw enum (#2339)`

Verified they fail on `origin/main`'s `QueuePage.tsx` (3 failed, 28 passed) and pass with the change.

```
cd web
npm run typecheck   # clean
npm run lint        # 0 errors, 6 pre-existing warnings
npm test            # 68 files, 671 tests, all passing
npm run build       # clean
```

## Sequencing

First of four frontend PRs. Nothing to land before it. Patch release material: #2339 and #2336 are both user visible wrong results on the queue page.

It unblocks the next three, all of which touch `QueuePage.tsx` and say "merge after this one":

1. this PR
2. `formatBytes` consolidation (#2350)
3. `ConfirmDialog` rollout (#2359)
4. `usePolling` hook (#2360)

### DIFF
--- a/web/src/components/downloadStatus.test.ts
+++ b/web/src/components/downloadStatus.test.ts
@@ -1,0 +1,104 @@
+import { describe, it, expect } from 'vitest'
+import {
+  DOWNLOAD_STATUSES,
+  FAILED_STATUSES,
+  RETRYABLE_STATUSES,
+  downloadStatusBadge,
+  isMatchable,
+  isRetryable,
+} from './downloadStatus'
+
+// Stub t: return the key so we can assert which label path was taken without
+// depending on the loaded i18n resources.
+const t = ((key: string) => key) as unknown as Parameters<typeof downloadStatusBadge>[1]
+
+// Every DownloadState constant declared in internal/models/download_state.go.
+// Copied deliberately: if a state is added there and not here, the coverage
+// test below fails and whoever added it gets told to give it a label and a chip
+// instead of shipping a grey pill reading the raw enum (#2339).
+const GO_DOWNLOAD_STATES = [
+  'grabbed',
+  'downloading',
+  'completed',
+  'importPending',
+  'importing',
+  'imported',
+  'failed',
+  'importFailed',
+  'importBlocked',
+  'importExternal',
+  'importHeld',
+]
+
+describe('DOWNLOAD_STATUSES', () => {
+  it('covers every state internal/models/download_state.go defines', () => {
+    for (const state of GO_DOWNLOAD_STATES) {
+      expect(DOWNLOAD_STATUSES[state], `no table entry for "${state}"`).toBeDefined()
+    }
+  })
+
+  it('has no entry the Go side does not define', () => {
+    expect(Object.keys(DOWNLOAD_STATUSES).sort()).toEqual([...GO_DOWNLOAD_STATES].sort())
+  })
+
+  it('gives every state a label key and a chip class', () => {
+    for (const [state, entry] of Object.entries(DOWNLOAD_STATUSES)) {
+      expect(entry.labelKey, state).toBeTruthy()
+      expect(entry.chip, state).toBeTruthy()
+    }
+  })
+
+  it('explains the two non-terminal hand-off states', () => {
+    expect(DOWNLOAD_STATUSES.importExternal.descriptionKey).toBeTruthy()
+    expect(DOWNLOAD_STATUSES.importHeld.descriptionKey).toBeTruthy()
+  })
+})
+
+describe('derived state sets', () => {
+  it('counts the three failed states', () => {
+    expect([...FAILED_STATUSES].sort()).toEqual(['failed', 'importBlocked', 'importFailed'])
+  })
+
+  // internal/db/downloads.go ResetImportRetry updates
+  // "WHERE ... status IN (StateImportFailed, StateImportBlocked)", so the bulk
+  // retry must offer exactly those two — #2336 shipped with importFailed only.
+  it('matches the states ResetImportRetry accepts', () => {
+    expect([...RETRYABLE_STATUSES].sort()).toEqual(['importBlocked', 'importFailed'])
+    expect(isRetryable('importBlocked')).toBe(true)
+    expect(isRetryable('importFailed')).toBe(true)
+    expect(isRetryable('failed')).toBe(false)
+  })
+
+  it('treats importFailed and importBlocked as matchable and nothing else', () => {
+    expect(isMatchable('importFailed')).toBe(true)
+    expect(isMatchable('importBlocked')).toBe(true)
+    expect(isMatchable('failed')).toBe(false)
+    expect(isMatchable('importExternal')).toBe(false)
+  })
+})
+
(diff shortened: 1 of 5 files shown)


vavallee:
Correcting one thing in the description before this goes in: I said the `history.events.*` keys are translated in eight locales. They are translated in one, Korean. Only `ko.json` has that subtree; de, es, fr, id, nl and tl have none. Reusing the keys is still the right call, there is just less existing translation riding on it than I claimed. Description updated so the squash message does not carry it.

Also worth recording, and not blocking: `downloadStatus.test.ts` does not actually pin the table against `download_state.go`. `GO_DOWNLOAD_STATES` is a hand copied TS array, so adding a twelfth state on the Go side leaves the test green. The comment in `downloadStatus.ts` and the fix requirement in #2342 both claim otherwise. The Go side already has the pattern that would work, `internal/indexer/quality_editor_vocab_test.go` reads a TSX file from a Go test, so the durable version of this guard is a Go test reading `downloadStatus.ts`. Leaving #2342 open for that rather than closing it here.

PR #2378, opened 2026-09-05
perf(api,indexer): memoise the image cache total and stop recompiling cached regexes (#2340) (#2341) (#2344)

### LINKED ISSUE #2340: GET /system/status walks the entire image cache on every call
`GET /api/v1/system/status` calls `ImageProxyHandler.CacheSize` (`internal/api/imageproxy.go:279`) on every single request, and that method does a full `filepath.WalkDir` over the whole image cache with a `d.Info()` stat per entry.

`cmd/bindery/main.go:818`:

```go
r.Get("/system/status", func(w http.ResponseWriter, r *http.Request) {
	w.Header().Set("Content-Type", "application/json")
	cacheBytes, _ := imageProxyHandler.CacheSize()
```

The cache stores two files per cover (the image plus its `.ct` sidecar), so a 10k book library is roughly 20k stat calls per status request. The frontend calls `api.status()` from the root shell on mount (`web/src/App.tsx:76`) and from four more pages, so this fires on basically every navigation into the app.

How to trigger: populate the image cache on a large library, then hit `/api/v1/system/status` in a loop and watch the disk IO.

What a user sees: a slow status response and a busy disk on every page load, worst on spinning disks and network mounts.

Fix shape: memoise the total behind a mutex, seed it lazily, add the delta when `Serve` writes a new cache file, and recompute it from the existing `StartEviction` sweep goroutine (`imageproxy.go:246`) which already walks the tree once a day.

## Sequencing

Next minor, no dependencies. Lands with the `phraseRegex` compile fix and the `regexCache` bound in one perf PR since they are all hot path memoisation. Nothing depends on it.

### LINKED ISSUE #2341: phraseRegex and inOrderRegex compile a regex on every call, even on a cache hit
`phraseRegex` (`internal/indexer/release.go:133`) and `inOrderRegex` (`internal/indexer/inorder.go:35`) both end with:

```go
re, _ := regexCache.LoadOrStore(pattern, regexp.MustCompile(pattern))
return re.(*regexp.Regexp)
```

Go evaluates the arguments before the call, so `regexp.MustCompile` runs on every invocation and the compiled value is thrown away on a cache hit. The cache saves the allocation of the map entry and nothing else.

`WordBoundaryRegex` a few lines up at `release.go:111` already does it the right way:

```go
if v, ok := regexCache.Load(kw); ok {
	return v.(*regexp.Regexp)
}
re := regexp.MustCompile(...)
regexCache.Store(kw, re)
```

How to trigger: `ContainsPhrase` runs once per release inside the search filter loop, so a 500 result search compiles the same regex 500 times and discards 499 of them. Same for `containsInOrder`.

What a user sees: slower searches than necessary, more so on fan out sweeps that filter thousands of releases across indexers.

Fix shape: use the same Load then Store pattern in both functions so the compile only happens on a miss.

## Sequencing

Next minor, no dependencies. Lands with the `/system/status` memo and the `regexCache` bound in one perf PR. Nothing depends on it.

### LINKED ISSUE #2344: regexCache is an unbounded sync.Map with no eviction
`regexCache` (`internal/indexer/release.go:38`) is a bare `sync.Map` with no cap and no eviction:

```go
regexCache = sync.Map{} // map[string]*regexp.Regexp
```

Three functions write to it: `WordBoundaryRegex` keyed on a title or author token, `phraseRegex` keyed on the generated phrase pattern, and `inOrderRegex` keyed on the generated in order pattern. Every distinct token or word sequence the instance has ever searched stays resident for the life of the process, together with its compiled `*regexp.Regexp`.

How to trigger: run automatic searches on a large monitored library over weeks. Every new book title contributes fresh tokens and a fresh phrase pattern, none of which are ever reused once the book is imported.

What a user sees: RSS growing slowly and never coming back down, which matters most on a Raspberry Pi or a small VPS. Not a fast leak, just an unbounded one.

Every other cache in the codebase is capped: `internal/metadata/ttl_cache.go` caps on entry count, `internal/indexer/newznab/querycache.go` caps on bytes. This one was missed.

Fix shape: replace the `sync.Map` with a small capped map behind a mutex plus a FIFO ring of keys for eviction. No new dependency (MIT project, and `ttlCache` is unexported in `internal/metadata` so it is not reusable here).

## Sequencing

Next minor. Should land with or after the `phraseRegex` compile fix since both touch the same three functions. Lands in one perf PR with that and the `/system/status` memo.

vavallee:
Three small perf fixes on paths that run constantly. No behaviour change.

**`/system/status` walked the image cache on every request.** `ImageProxyHandler.CacheSize` did a full `WalkDir` with a stat per file, two files per cached cover, and the frontend calls `api.status()` from the root shell plus four pages. The total is memoised now: seeded on first use, adjusted by the exact on-disk delta after each cache write (measured on both sides of the write so a refetch of an expired entry does not double count), recounted by the daily eviction sweep, and otherwise refreshed at most once per `imageCacheSizeTTL` to catch changes made behind the handler's back.

**`phraseRegex` and `inOrderRegex` compiled on every call.** `regexCache.LoadOrStore(pattern, regexp.MustCompile(pattern))` evaluates the compile before the lookup and throws it away on a hit. `ContainsPhrase` runs once per release in the search filter loop, so a 500 result search paid 500 compiles. Both now use the Load then Store form `WordBoundaryRegex` already used.

**`regexCache` was unbounded.** A bare `sync.Map` keyed by title tokens and generated patterns, so every token an instance ever searched stayed resident with its compiled regex forever. Now a capped map plus a FIFO ring at 2048 entries. FIFO rather than LRU because a miss costs one compile. No new dependency.

Closes #2340
Closes #2341
Closes #2344

## Tests

New `internal/indexer/regexcache_test.go`:
- `TestPhraseRegexCompilesOncePerPattern` and `TestInOrderRegexCompilesOncePerPattern` count compiles through a `compileRegex` package variable. Confirmed failing on the old body: `phraseRegex compiled 50 times for one pattern, want 1`.
- `TestRegexCacheIsBounded`, `TestRegexCacheEvictsOldestFirst`, `TestRegexCacheStoreIsIdempotent` cover the cap, the FIFO order and the ring bookkeeping.
- `BenchmarkContainsPhraseCached` for the hit path.

New `internal/api/imageproxy_cachesize_test.go`:
- `TestImageProxy_CacheSizeIsMemoised` writes a file behind the handler's back and asserts the second call does not see it (a re-walk would). Fails on the old unconditional walk.
- `TestImageProxy_CacheSizeTracksWrites` asserts the memo equals what a walk would report after two misses plus one hit.
- `TestImageProxy_EvictionRecountsCacheSize` asserts the sweep refreshes the number.

Ran:

```
go build ./...                                              ok
go vet ./internal/api/... ./internal/indexer/...             ok
go test ./internal/api/... ./internal/indexer/... -count=1   ok (api 141.7s, indexer 0.6s)
golangci-lint run ./internal/api/... ./internal/indexer/...  0 issues
```

## Sequencing

Next minor, nothing blocks it. Lands before the other two PRs from this batch (series pagination, backfill revision marker), neither of which touches these files. Nothing depends on it.


vavallee:
Heads up: #2407 from @v01dst just landed and fixes the #2341 half of this independently. That was a collision I caused by filing #2341 and opening this PR eight minutes later.

This needs a rebase onto main. The #2341 hunks in `release.go` and `inorder.go` will conflict, and the resolution is to take the bounded cache from here, since `boundedRegexCache.load`/`store` already implements the check-then-compile shape their fix introduced. Nothing of theirs is lost, it just moves into the new type.

What is left of this PR after the rebase is #2340 (memoising the image cache total) and #2344 (bounding the cache). Retitling and dropping the #2341 reference at that point.

Issue #211, opened 2026-04-17
bindery doesn´t find books on my indexer, but manual search does


zippoking:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

I added different authors and when i want to download the e-book, bindery says "no results on any indexer. If i go to my indexers website and search manual for the e-book, i get results.

i made  sure that i added the correct categories in my indexer settings:

<img width="1252" height="516" alt="Image" src="https://github.com/user-attachments/assets/fc78b900-08e5-461e-be50-b409f80fbbb1" />

###
###
###
###

Example 1:
Author: Cornelia Funke
Title: Hinter verzauberten Fenstern

bindery shows:

<img width="906" height="573" alt="Image" src="https://github.com/user-attachments/assets/f508d667-1d62-42ea-93b2-7e5afe7ab89d" />

Manual search without any filter on the indexer shows:
<img width="2240" height="1130" alt="Image" src="https://github.com/user-attachments/assets/324863ad-dde4-4a5f-a0da-5eb9b034aa7f" />

Manual search with ebooks filter
<img width="2216" height="952" alt="Image" src="https://github.com/user-attachments/assets/3e9137f8-9e35-4e62-a569-af58089fe623" />
 
###

Example 2:
Author: Cornelia Funke
Title: Gespensterjäger

bindery search:

<img width="925" height="598" alt="Image" src="https://github.com/user-attachments/assets/ec7b4d47-42af-41de-96c3-d18580acef7b" />

manual search on indexer:

<img width="2255" height="918" alt="Image" src="https://github.com/user-attachments/assets/fa9f550a-0388-4b22-a760-1b380dd2ba76" />

###

Example 3:
Author: Juli Zeh
Title: Die Stille ist ein Geräusch

bindery search:
<img width="940" height="612" alt="Image" src="https://github.com/user-attachments/assets/9c2073f0-f411-4c84-8a8b-a527108ba0aa" />

manual search:
<img width="2294" height="1057" alt="Image" src="https://github.com/user-attachments/assets/a18a764f-5397-4c01-af64-c0b4ae009477" />

### Expected behaviour

I expect bindery to list the same search resultsin comparison to a manual search on my indexer.

What is the search term bindery uses on my indexer? There are no logs.

### Steps to reproduce

1. add author
2. look for books
3. try to download book
4. no results found

### Trace logs

```text
[...]

Issue #2312, opened 2026-09-03
Per-indexer daily query cap: one wanted sweep can exhaust a tracker's whole API allowance


vavallee:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I checked the roadmap and this isn't already planned

Fast-tracking this. Reported by DrCrypt in Discord: a large ebook collection on a 7 day search interval "ends up basically blowing through all my API calls on all my trackers and getting all my other services locked out for the day whenever it runs". That last part is what moves it up the list. Bindery is not just degrading itself, it is spending a budget shared with everything else pointed at that tracker.

This is item 2 of #1097. That issue asked for three things, #1117 shipped item 1 (`search.interval`) and said in as many words that "per-indexer rate limiting and batch cooldowns are tracked as separate follow-ups in that issue", but the issue was then closed as completed, so the follow-ups have been tracked nowhere since June. Second time it has been asked for.

### Current behaviour

There is no query budget anywhere in the codebase. `models.Indexer` has no quota field, no migration defines one, and the only search-related settings keys are `search.interval` and `search.preferredLanguage`.

What exists is reactive rather than preventive:

* `internal/indexer/cooldown.go` (#1934) benches an indexer for 1h, bounded to [1m, 24h] and honouring `Retry-After`, but only after it has already returned a rate-limit error. The calls are spent by then, and in this report the account is already locked out.
* `searchPaceInterval` (3s, `internal/scheduler/scheduler.go:225`) with `scheduledWantedSearchConcurrency = 2` spaces launches apart. That bounds the rate at roughly 1200 book-searches an hour. It does nothing about the total.

Worth naming to avoid confusion: the Hardcover throttle added in #2277 is a metadata-provider pacer and has no bearing on indexer queries.

`wantedSearchQueue` (`internal/scheduler/scheduler.go:1116`) returns every Wanted book minus excluded and in-flight ones, and `searchWanted` hands the whole list to `RunBoundedPaced`. No per-sweep cap, no carry-over between sweeps. Per book per indexer the cost is one query when tier 1 matches and up to four when it falls through the text tiers (`internal/indexer/newznab/client.go:377-444`), so a sweep costs between `books × indexers` and `books × indexers × 4` queries in one continuous burst.

### Why search.interval is not the answer

[...]

PR #2391, opened 2026-09-05
fix(indexer): search localized and original book titles

IfritDMC:
## Summary

Avoid sending combined localized/original display titles to indexers. Bindery now builds a bounded title-candidate list, uses the localized title for automatic searches, and exposes the original-language fallback only to interactive searches.

Refs #211.

## Implementation notes

- Derives at most two safe title candidates from metadata and a single spaced slash outside brackets/parentheses.
- Keeps the existing four-tier cascade for the primary title and limits the fallback to two generic requests.
- Marks original-language results as `manualOnly`, so the scheduler and auto-grab path cannot select them.
- Preserves same-language edition aliases as automatic fallbacks.
- Exposes `matchedTitle`, `manualOnly`, candidates, and per-indexer attempted titles in the API/debug UI.
- Adds no database migration or dependency.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` not required: no env vars, persisted config, or upgrade path changed
- [x] Added `changelog.d/multilingual-title-search.md`
- [ ] Wiki pages updated — the README documents the user-facing behaviour; happy to add a wiki follow-up if preferred

## Test plan

- [x] `go test ./internal/indexer/...`
- [x] `go test ./internal/api -run TestSearchBook -count=1`
- [x] `go test -count=1 ./internal/scheduler`
- [x] `go vet ./cmd/... ./internal/...`
- [x] Frontend: 659 tests, typecheck, lint (0 errors; 6 pre-existing warnings), and production build
- [x] Isolated live Prowlarr search with auto-grab disabled: Spanish localized result plus English manual-only fallback; no download triggered

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ Clean-room Go rewrite, modern React UI, MIT-licensed, actively developed.
 
 **Search & downloads**
 - Newznab + Torznab indexers queried in parallel, deduplicated, then composite-ranked by format quality, edition tags (RETAIL / UNABRIDGED / ABRIDGED), year match, grab count, size, and ISBN exact-match bonus.
-- Smart matching — four-tier query fallback (`t=book` → `surname+title` → `author+title` → title), word-boundary keyword matching, contiguous-phrase requirement for multi-word titles, dual-author-anchor for ambiguous short titles, subtitle-aware (`Title: Subtitle`).
+- Smart matching — four-tier query fallback (`t=book` → `surname+title` → `author+title` → title), bounded localized/original-title candidates, word-boundary keyword matching, contiguous-phrase requirement for multi-word titles, dual-author-anchor for ambiguous short titles, subtitle-aware (`Title: Subtitle`). Original-language fallbacks are visible in interactive search but excluded from automatic grabs when the metadata profile selects a localized language.
 - SABnzbd, NZBGet, qBittorrent, Transmission, Deluge, rTorrent/ruTorrent — with **Use SSL** and **URL Base** for reverse-proxy subpaths.
 - Auto-grab sweep every 12h, immediate search on add or `wanted` flip, plus interactive per-book search and "Search all wanted" per author. Global kill-switch pauses auto-grab without losing your monitored list.
 - Quality profiles covering every format release parsing recognises (EPUB, MOBI, AZW3, PDF, plus AZW, DJVU, CBR, CBZ, FB2, LIT, RTF, TXT and the audio containers M4B, M4A, FLAC, MP3, OGG), language filter, regex-based custom formats, delay profiles, blocklist (consulted on every search; one-click add from History), and failure visibility in Queue and History.
--- a/changelog.d/multilingual-title-search.md
+++ b/changelog.d/multilingual-title-search.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Localized book searches** — search a safe localized title alias first and expose the original-language title as a clearly marked manual fallback, instead of sending combined bilingual titles to indexers.
--- a/internal/api/indexers.go
+++ b/internal/api/indexers.go
@@ -434,23 +434,27 @@ func (h *IndexerHandler) SearchBook(w http.ResponseWriter, r *http.Request) {
 	}
 
 	crit := indexer.MatchCriteria{
-		Title:            book.Title,
-		Author:           authorName,
-		MediaType:        book.MediaType,
-		ASIN:             book.ASIN,
-		AllowedLanguages: allowedLangs,
-		AuthorAliases:    authorAliases,
+		Title:               book.Title,
+		AllowManualFallback: true,
+		Author:              authorName,
+		MediaType:           book.MediaType,
+		ASIN:                book.ASIN,
+		AllowedLanguages:    allowedLangs,
+		AuthorAliases:       authorAliases,
 	}
 	if book.ReleaseDate != nil {
 		crit.Year = book.ReleaseDate.Year()
 	}
+	var bookEditions []models.Edition
 	if h.editions != nil {
 		if eds, err := h.editions.ListByBook(r.Context(), book.ID); err != nil {
 			slog.Warn("failed to load editions for search ISBN", "book_id", book.ID, "error", err)
 		} else {
+			bookEditions = eds
 			crit.ISBN = indexer.CriteriaISBN(book, eds)
 		}
 	}
+	crit.TitleCandidates = indexer.BuildTitleCandidates(*book, bookEditions, allowedLangs)
 
 	// For dual-format books (media_type='both'), run one search per format so
 	// each uses its own category tree (7xxx for ebooks, 3xxx for audiobooks).
@@ -523,7 +527,7 @@ func (h *IndexerHandler) SearchBook(w http.ResponseWriter, r *http.Request) {
 	beforeLang := len(results)
 	filterDesc := ""
 	if len(allowedLangs) > 0 {
-		results = indexer.FilterByAllowedLanguages(results, allowedLangs)
+		results = indexer.FilterByAllowedLanguagesInteractive(results, allowedLangs)
 		filterDesc = "allowed=" + strings.Join(allowedLangs, ",")
 	} else if s, _ := h.settings.Get(r.Context(), "search.preferredLanguage"); s != nil {
 		results = indexer.FilterByLanguage(results, s.Value)
--- a/internal/api/indexers_test.go
+++ b/internal/api/indexers_test.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"strconv"
 	"strings"
 	"sync"
@@ -477,6 +478,60 @@ func TestSearchBook_RedactsIndexerAPIKey(t *testing.T) {
 	}
 	if len(resp.Results) != 1 || resp.Results[0].NZBURL != "https://idx.example.com/dl?id=eb1" {
 		t.Fatalf("nzbUrl not redacted as expected: %+v", resp.Results)
+	}
+}
+
+func TestSearchBookBuildsBilingualInteractiveCandidates(t *testing.T) {
+	database, err := db.OpenMemory()
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Cleanup(func() { database.Close() })
+
+	ctx := context.Background()
+	authorRepo := db.NewAuthorRepo(database)
+	author := &models.Author{ForeignID: "OL1A", Name: "Brandon Sanderson", SortName: "Sanderson, Brandon", MetadataProvider: "openlibrary", Monitored: true}
+	if err := authorRepo.Create(ctx, author); err != nil {
+		t.Fatal(err)
+	}
+	bookRepo := db.NewBookRepo(database)
+	book := &models.Book{
+		Title:            "El imperio final / The Final Empire",
+		OriginalTitle:    "The Final Empire",
+		Language:         "spa",
+		ForeignID:        "OL1W",
+		AuthorID:         author.ID,
+		MediaType:        models.MediaTypeEbook,
+		MetadataProvider: "openlibrary",
+		Monitored:        true,
+	}
+	if err := bookRepo.Create(ctx, book); err != nil {
+		t.Fatal(err)
+	}
+
+	mock := &mockIndexerSearcher{}
+	h := NewIndexerHandler(
+		db.NewIndexerRepo(database), bookRepo, authorRepo,
+		db.NewMetadataProfileRepo(database), mock,
+		db.NewSettingsRepo(database), db.NewBlocklistRepo(database),
+	)
+	rec := httptest.NewRecorder()
+	req := withURLParam(httptest.NewRequest(http.MethodGet, "/indexer/book/1/search", nil), "id", strconv.FormatInt(book.ID, 10))
+	h.SearchBook(rec, req)
+	if rec.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d: %s", rec.Code, rec.Body.String())
+	}
+
+	criteria := mock.criteria()
+	if !criteria.AllowManualFallback {
+		t.Fatal("interactive search did not enable manual title fallbacks")
+	}
+	want := []indexer.TitleCandidate{
+		{Title: "El imperio final", Language: "spa"},
+		{Title: "The Final Empire", ManualOnly: true},
+	}
+	if !reflect.DeepEqual(criteria.TitleCandidates, want) {
+		t.Fatalf("title candidates = %#v, want %#v", criteria.TitleCandidates, want)
 	}
 }
 
--- a/internal/indexer/debug.go
+++ b/internal/indexer/debug.go
@@ -25,28 +25,30 @@ type SearchDebug struct {
 
 // SearchQueryDebug is the effective criteria sent into the searcher.
 type SearchQueryDebug struct {
-	Title            string   `json:"title,omitempty"`
-	Author           string   `json:"author,omitempty"`
-	Year             int      `json:"year,omitempty"`
-	ISBN             string   `json:"isbn,omitempty"`
-	ASIN             string   `json:"asin,omitempty"`
-	MediaType        string   `json:"mediaType,omitempty"`
-	AllowedLanguages []string `json:"allowedLanguages,omitempty"`
-	FreeText         string   `json:"freeText,omitempty"`
+	Title            string           `json:"title,omitempty"`
+	TitleCandidates  []TitleCandidate `json:"titleCandidates,omitempty"`
+	Author           string           `json:"author,omitempty"`
+	Year             int              `json:"year,omitempty"`
+	ISBN             string           `json:"isbn,omitempty"`
+	ASIN             string           `json:"asin,omitempty"`
+	MediaType        string           `json:"mediaType,omitempty"`
+	AllowedLanguages []string         `json:"allowedLanguages,omitempty"`
+	FreeText         string           `json:"freeText,omitempty"`
 }
 
 // IndexerDebug describes what happened for a single indexer: whether it was
 // contacted, the categories sent, how many results came back, and any error.
 type IndexerDebug struct {
-	IndexerID   int64  `json:"indexerId"`
-	IndexerName string `json:"indexerName"`
-	Enabled     bool   `json:"enabled"`
-	Skipped     bool   `json:"skipped,omitempty"`
-	SkipReason  string `json:"skipReason,omitempty"`
-	Categories  []int  `json:"categories,omitempty"`
-	ResultCount int    `json:"resultCount"`
-	DurationMs  int64  `json:"durationMs"`
-	Error       string `json:"error,omitempty"`
+	IndexerID       int64    `json:"indexerId"`
+	IndexerName     string   `json:"indexerName"`
+	Enabled         bool     `json:"enabled"`
+	Skipped         bool     `json:"skipped,omitempty"`
+	SkipReason      string   `json:"skipReason,omitempty"`
+	Categories      []int    `json:"categories,omitempty"`
+	AttemptedTitles []string `json:"attemptedTitles,omitempty"`
+	ResultCount     int      `json:"resultCount"`
+	DurationMs      int64    `json:"durationMs"`
+	Error           string   `json:"error,omitempty"`
 }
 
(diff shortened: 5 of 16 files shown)


vavallee:
Thanks, this is a lot of careful work and the checklist and test plan are the most complete I have had on a PR here. The live Prowlarr run with auto-grab disabled is exactly the right way to test this.

Giving it a proper review rather than a skim, because of where it sits: `manualOnly` is a safety gate on the automatic grab path, so if an original language result can reach an auto-grab anywhere, that lands wrong files in someone library unattended. I want to walk every path from a search result to a grab before this goes in, not just the scheduler one.

Two things worth knowing while I do that:

- **The branch is based on `7917b7ec`, and main has moved.** v1.33.4 shipped today and #2407 changed the regex cache in `internal/indexer/release.go` and `inorder.go`, while #2385 changed the wanted sweep in `internal/scheduler/scheduler.go`. GitHub still calls it mergeable so there is no textual conflict, but I am checking for a semantic one in `scheduler.go` specifically.
- **Query volume is the thing I am most likely to come back to you about.** #2385 just put previously unsearched dual format books back into the sweep, and #2313 is open because people get banned from private trackers over query counts. If the candidate list multiplies automatic queries per book at all, I will want that number before merging, even if the multiplier is small.

Container Scan is red on every PR in this repo right now including main, so ignore it. Nothing else is failing.

Will come back with specifics.

vavallee:
Reviewed properly. This is genuinely good work and I want it in, with two changes first.

The design choice to enforce `manualOnly` by never issuing the query, rather than filtering results afterwards, is the right one and stronger than what I went looking for. I traced all eight automatic paths (wanted sweep, `SearchAndGrabBook`, stall re-search, series fill, bulk search, add author with search on add, add single book, recommendations) and they funnel into `searchAndGrabFormat`, which leaves `AllowManualFallback` false, so a manual only result cannot exist in an automatic set at all. `TestSearchBookDoesNotUseManualFallback` is a real test: I reverted the gate and watched it fail with the exact query it should never have sent. The candidate builder also survived every hostile input I threw at it, CJK, RTL, unbalanced brackets, bare slashes, a 700 character title, with no panic and no empty or punctuation only query.

It also merges into current main cleanly. I did the merge locally and checked for a semantic conflict, not just a textual one: #2385 kept `searchAndGrabFormat` intact and moved only the kill switch up a level, so your hunk still lands in the right function, and #2407's regex cache does not intersect. Sign offs present, changelog fragment in the right shape, no new Go module or npm package.

## Two things before it goes in

**1. Please drop the three added lines in `FilterByAllowedLanguages` (`internal/indexer/searcher.go:1259`).**

They make the filter require the raw `newznab:attr language` value to normalize into the allowed set, and `normalizeLanguageCode` consults only `iso639TwoLetterAliases`, which is ten two letter codes. I probed it against `allowed = ["eng"]`: `"English"`, `"en-US"`, `"und"`, `"unknown"` and `"N/A"` are all dropped. The default metadata profile ships `allowed_languages = 'eng'` (`003_features.sql:55`), so this is live on every install that has never opened the profile editor, on both the automatic and interactive paths.

The tell is internal to the function. The release name check three lines below uses `releaseLanguageTags`, which does map `"english"` to `"eng"` at `:1121`. So the same word is understood in one half and rejected in the other. It also makes the comment above the function false, since that says the tag check can only ever be a negative signal, and this is a positive membership requirement.

I removed the three lines and the whole indexer and scheduler suite still passes, so nothing covers them. They are also not needed for your feature, since manual only results go through `FilterByAllowedLanguagesInteractive` anyway. If you do want a language attribute check, `LanguageFilterSpec` in `internal/decision/specs.go:226` already exists for it and would surface a rejection reason in the debug panel instead of dropping the result silently.

**2. Please make the edition alias `manualOnly: true` as well** (`title_candidates.go:91`, the `add(edition.Title, language, false)` call).

An edition alias in an allowed language is currently an automatic candidate. Both providers store `"eng"` on English editions and the default profile allows `eng`, so an ordinary English library picks up a second automatic candidate whenever an edition title differs from the book title after normalization. A series prefix or a subtitle is enough:

```
edition-alias-english  AUTO=["The Final Empire" "Mistborn: The Final Empire"]
```

That takes the worst case from 8 outbound requests per book per format per indexer to 10, and the fallback fires precisely when the primary found nothing, which is the population the sweep retries forever. #2385 shipped yesterday and put dual format books back in the sweep, so it doubles again for those, and #2313 is open because someone burned a day of private tracker allowance in one sweep.

If you want the alias to stay automatic that is a fair argument to make, but it needs to be an explicit decision in the PR body with the request count in it. As written the PR reads as though automatic volume is unchanged, and it is not.

## Smaller things, take or leave

The scheduler side of the gate has no test. The indexer level one is solid, but the property that actually keeps users safe is that `searchAndGrabFormat` never sets `AllowManualFallback`, and nothing pins it. Worth knowing why: `decision.Release` has no `ManualOnly` field, the grab takes `&results[i]` without checking it, and `pending_releases` JSON round trips the whole struct, so a manual only release that ever landed there would be re-grabbed unattended. Nothing reaches those today. This boolean is the only thing standing between them.

`dedupe` at `searcher.go:922` now includes `MatchedTitle` in the key, so the same GUID found by two indexers under different titles is listed twice. Harmless for the grab, but it weakens the deduplicated contract.

`separatorInsideGrouping` treats only ASCII brackets as grouping, so a Japanese title using full width parentheses splits inside them.

The `manual fallback` badge at `BookDetailPage.tsx:187` is hardcoded English while the rest of the file uses `t()`.

The changelog fragment wants a number in the filename per `changelog.d/README.md`.

And `searchBookTimeout` is 60s with a comment saying each per indexer search may issue up to 4 sequential calls. It can now issue 10, so a slow indexer that used to just fit may get truncated mid fallback.

Once those two blocking items are addressed I am happy for this to go in.

vavallee:
Heads up on a merge hazard between this PR and #2461 that git will not flag.

#2461 makes `internal/models/language.go` the single ISO 639 source and **deletes `indexer.iso639TwoLetterAliases`**, along with the two Audible copies. This PR adds `normalizeLanguageCode` in `internal/indexer/title_candidates.go`, whose body reads that map:

```go
func normalizeLanguageCode(language string) string {
	language = strings.ToLower(strings.TrimSpace(language))
	if alias, ok := iso639TwoLetterAliases[language]; ok {
		return alias
	}
	return language
}
```

The two changes touch different files and different regions, so **git merges them without a conflict and the result does not compile**. It is called at 10 or more sites in this branch.

Whichever lands second needs that body to become:

```go
return models.NormalizeLanguageCode(language)
```

That is not just a repair. The shared function strips region and script subtags and handles the ISO 639-2/T spellings, so the new `r.Language` check in this PR gets `pt-BR`, `zh-Hans`, `deu` and `German` handled, which the local copy does not do today.

No action needed right now if #2461 lands first; I will do the swap on merge and re-run the build. Flagging it because a clean merge that fails to compile has bitten this repo before (#2211 and #2216 broke `main` exactly this way), and the fix is obvious only if you know it is coming.

vavallee:
Reviewed this properly since it is a large change to the automatic search path. The title-candidate design is sound and the `manualOnly` gate genuinely holds: `AllowManualFallback` is set only at `internal/api/indexers.go:438`, the scheduler leaves it false, and `effectiveTitleCandidates` drops manual-only candidates before any query is issued. So the scheduler cannot grab an original-language fallback. That part does what the summary says.

One finding is blocking, and it is not in the title-candidate code.

## `FilterByAllowedLanguages` now drops releases in languages the profile allows

`searcher.go` adds this to the shared filter:

```go
if language := normalizeLanguageCode(r.Language); language != "" && !set[language] {
    ok = false
}
```

`normalizeLanguageCode` (`title_candidates.go:272`) returns its input unchanged when it is not one of the ten entries in `iso639TwoLetterAliases`. `set` is built through that same ten-entry map. So any language outside those ten fails to reconcile with itself.

`SearchResult.Language` is documented as "ISO 639-1 from newznab:attr language", and this filter is on the **automatic** path (`scheduler.go:776`), so this silently narrows results for existing installs.

I ran the same table against both branches:

| profile | `r.Language` | main | #2391 |
|---|---|---|---|
| `swe` | `sv` | kept | **dropped** |
| `eng` | `en-US` | kept | **dropped** |
| `eng` | `English` | kept | **dropped** |
| `eng` | `en` | kept | kept |
| `eng` | `eng` | kept | kept |
| `eng` | `de` | kept | dropped (intended) |

The `de` row is the improvement you were going for and it works. The first three are the regression. The Swedish one is the sharpest: a profile that explicitly **allows** Swedish drops a release the indexer tagged Swedish, because `swe` and `sv` never meet. The ten-entry map covers en/fr/de/nl/es/it/pt/ja/zh/ru and nothing else, so every other language is in that state.

This also sits awkwardly against the function's own contract three lines up: *"Untagged releases always pass ... the tag check can only ever be a negative signal."* A `Language` value the mapping does not understand is closer to untagged than to disallowed, and right now it is treated as disallowed.

## Both problems have the same fix

#2463 replaces `iso639TwoLetterAliases` with `models.NormalizeLanguageCode`, which handles exactly these cases: `"english"` to `eng`, `"swedish"`/`"svenska"` to `swe`, and region and script subtag stripping (`en-US` to `en` to `eng`). Pointing `normalizeLanguageCode` at it fixes the table above.

That is also the rebase this branch already needs. #2463 **deletes** `iso639TwoLetterAliases` from `searcher.go`, and `normalizeLanguageCode` lives in `title_candidates.go` and reads it. Different files, so git merges the two cleanly and package `indexer` then fails to compile. Same shape as #2211/#2216, which broke main once already.

So: rebase onto #2463 and route through `models.NormalizeLanguageCode`. One change, both issues.

## Smaller, lower confidence

`dedupe` appends `MatchedTitle` to the key so a primary-query copy cannot erase a fallback copy's provenance. Within one indexer that is safe, because `searchIndexerBook` only reaches the fallback when the primary produced nothing relevant, so `filterRelevantForCriteria` drops the primary copy. Across indexers I do not think it holds: indexer A can stop at the primary while indexer B falls through to the fallback, and the same release then carries two different `MatchedTitle` values, survives dedupe, and passes relevance under both. Worth a look, but it needs two indexers carrying the same release and diverging on which candidate hits, so I would not hold the PR for it.

Nothing else stood out. The four-tier cascade is preserved for the primary title, `BookSearchFallback` is correctly bounded to two requests, and `maxTitleCandidates` caps the fan-out.

vavallee:
Following up my earlier comment about the compile break. There is a second, worse problem on the same lines, and I measured it on both branches rather than reasoning about it.

The new check in `FilterByAllowedLanguages`:

```go
if language := normalizeLanguageCode(r.Language); language != "" && !set[language] {
    ok = false
}
```

`normalizeLanguageCode` returns its input unchanged when it misses `iso639TwoLetterAliases`, and `set` is built through that same ten-entry map. So any language outside those ten fails to reconcile **with itself**. One release, one profile entry, `r.Language` set:

| profile | `r.Language` | main | this branch |
|---|---|---|---|
| `swe` | `sv` | kept | **dropped** |
| `eng` | `en-US` | kept | **dropped** |
| `eng` | `English` | kept | **dropped** |
| `ger` | `deu` | kept | **dropped** |
| `por` | `pt-BR` | kept | **dropped** |
| `eng` | `de` | kept | dropped, and this one is the intended behaviour |

Only the last row is the improvement this change is going for. The rest are a profile dropping releases in a language it explicitly allows. `FilterByAllowedLanguages` is what the scheduler calls on the automatic search path, so this would show up as books quietly never grabbing.

`main` keeps all six only because it does not check `r.Language` at all, so this is new behaviour rather than a pre-existing bug.

**The fix is one line**, and it is the same line that fixes the compile break I flagged earlier. Once #2461 merges, `iso639TwoLetterAliases` no longer exists and the body becomes:

```go
func normalizeLanguageCode(language string) string {
	return models.NormalizeLanguageCode(language)
}
```

That handles all five failing rows: the region and script subtag strip covers `en-US` and `pt-BR`, the ISO 639-2 T-to-B map covers `deu`, and the language-name map covers `English` and `Swedish`. `sv` to `swe` is already in the two-letter map.

I am not going to push that to this branch. It is not mine to edit and the decision to merge it at all sits with the maintainer. Flagging it here so it is visible to whoever does.

Credit for spotting the reconciliation failure goes to the parallel review of this PR; I verified the table independently on both branches.

vavallee:
Correction and an addition to my two earlier comments, now that #2461 has merged and I can measure against real `main` rather than predict.

**My earlier "git merges this cleanly and it does not compile" warning was wrong, and the real situation is better in one way and worse in another.**

Better: merging this branch into `main` now produces ordinary textual conflicts, which git stops on and tells you about.

```
CONFLICT (content): Merge conflict in internal/indexer/debug.go
CONFLICT (content): Merge conflict in internal/indexer/searcher.go
```

Worse: **resolving those two files is not enough, and the merge will not tell you what is left.** `internal/indexer/title_candidates.go` is new on this branch, so it has no counterpart on `main` and can never conflict. It reads `iso639TwoLetterAliases`, which `main` no longer defines.

Resolving both flagged files toward `main` and building:

```
# github.com/vavallee/bindery/internal/indexer
internal/indexer/title_candidates.go:121:18: c.TitleCandidates undefined (type MatchCriteria has no field or method TitleCandidates)
internal/indexer/title_candidates.go:128:59: c.AllowManualFallback undefined (type MatchCriteria has no field or method AllowManualFallback)
internal/indexer/title_candidates.go:274:18: undefined: iso639TwoLetterAliases
```

The first two are this branch's own `MatchCriteria` fields, which live in the conflicted region of `searcher.go` and come back with a proper resolution. The third does not: it needs an edit to a file nothing flagged.

So the checklist for whoever rebases this:

1. Resolve `searcher.go` and `debug.go`, keeping `main`'s deletion of `iso639TwoLetterAliases` and this branch's `MatchCriteria` additions.
2. Change `normalizeLanguageCode` in `title_candidates.go` to `return models.NormalizeLanguageCode(language)`.
3. `go build ./internal/indexer/` before trusting the merge. This branch is also stale against #2312's query-budget change, which surfaces as `s.quota undefined` if you resolve the other way.

Step 2 is also the fix for the language regression in my previous comment, so it is not busywork: `models.NormalizeLanguageCode` handles `en-US`, `pt-BR`, `deu` and `English`, which the local ten-entry map does not.

Still not pushing any of this. Flagging it so the person who does has the whole picture, including the part the merge is silent about.

vavallee:
One more thing this branch needs, and it is not a merge artefact: **it predates the per-indexer daily query cap** (#2312, merged as `2d1f569e`).

`internal/indexer/searcher.go` on this branch has zero references to `s.quota`. `main` has four. The line this PR replaces:

```go
hits, err := client.BookSearch(ctx, c.Title, c.Author, cats)     // this branch
qctx := newznab.WithQueryBudget(ctx, s.quotaBudget(idx))         // main
hits, err := client.BookSearch(qctx, c.Title, c.Author, cats)
```

So `s.searchIndexerBook(ctx, client, c, cats)` takes the unbudgeted context, and the surrounding `quotaHold` gate, `flushQuota` defer and `budgetExhausted` branch are all things the branch has never seen. That is why resolving the conflict toward this branch's `searcher.go` fails on `s.quota undefined` rather than on the ISO 639 symbol.

Threading `qctx` through is mechanical. The part that is a decision, and the reason I do not think this is a one-liner any more:

**Should the fallback candidate's queries count against the daily cap?** They should, once `qctx` is threaded, and that is the correct answer. But it means a book whose primary title finds nothing now spends a second cascade: `BookSearch`'s four tiers, then `BookSearchFallback`'s two. For a user who set a cap because their indexer enforces one, searches that fall through consume roughly twice the budget they used to. That may be entirely acceptable — a fallback that never runs costs nothing, and it only runs when the primary found nothing relevant — but it is a product call worth making deliberately rather than inheriting from a rebase.

Taken with the language-filter finding above, this branch needs a real rebase onto current `main` rather than a conflict resolution: the ISO 639 swap, the query-budget threading, and a decision on the budget interaction. That is the contributor's call or a maintainer's, not something to patch in from the side.

The title-candidate design itself still looks right to me, and none of this touches it.

vavallee:
Pushed `b6341024` to this branch under the "allow edits from maintainers" setting. It is the one-line delegation from my earlier comments, and I want to correct something I said about it.

**I claimed that change "handles all five failing rows". On this branch as it stands, it handles three.** Measured after the push:

| profile | `r.Language` | before | after |
|---|---|---|---|
| `swe` | `sv` | dropped | **kept** |
| `eng` | `en-US` | dropped | **kept** |
| `por` | `pt-BR` | dropped | **kept** |
| `ger` | `deu` | dropped | still dropped |
| `eng` | `English` | dropped | still dropped |
| `eng` | `de` | dropped | dropped, intended |

The reason is that this branch's base predates #2463. `models.NormalizeLanguageCode` there has only the two-letter map, so it does the region and script subtag strip but not the ISO 639-2 T-to-B pairs (`deu` to `ger`) or the language names (`English` to `eng`). Those two maps are on `main` now, so **`deu` and `English` start passing the moment this branch rebases, with no further change to this file.** I should have checked that before saying "all five", rather than reasoning from what the function does on `main`.

The change is still the right one and is worth having now: it fixes three rows immediately, the other two follow from the rebase, and it removes this file's read of `indexer.iso639TwoLetterAliases`, which #2463 deleted.

That last part is the reason I did not simply leave this for the rebase. `title_candidates.go` is new on this branch, so it has no counterpart on `main` and can never produce a conflict marker. Without this commit, resolving the two files git *does* flag (`searcher.go`, `debug.go`) leaves a build that fails on a third file the merge said nothing about.

Still outstanding for whoever rebases, unchanged from my previous comment:

1. Resolve `searcher.go` and `debug.go`, keeping `main`'s deletion of `iso639TwoLetterAliases` and this branch's `MatchCriteria` additions.
2. `go build ./internal/indexer/` before trusting the merge — this branch is also stale against #2312's query-budget change, which surfaces as `s.quota undefined`.

Nothing else here is touched. The review of the substantive change, including the `manualOnly` flag and the `dedupe` keying on `MatchedTitle`, is still open and is not mine to close.

vavallee:
Hi @IfritDMC, thanks for this. The title candidate design is good and the manualOnly gate does what you say it does, I checked it end to end.

Main has moved under you though, and the branch needs a rebase before it can go in. Three things, all detailed in my comments above:

1. `iso639TwoLetterAliases` is gone from main. `normalizeLanguageCode` in `title_candidates.go` reads it, and since that file is new it never shows a conflict marker, so the build breaks in a spot the merge does not point at. `models.NormalizeLanguageCode` replaces it and also fixes the language filter bug below.
2. The new `r.Language` check in `FilterByAllowedLanguages` drops releases in languages the profile allows, because the old ten entry map cannot reconcile `sv` with `swe` or `deu` with `ger`. Switching to `models.NormalizeLanguageCode` handles it.
3. The branch predates the per indexer daily query cap, so `searcher.go` conflicts and `searchIndexerBook` needs the budgeted context threading through. Worth a thought on whether the fallback cascade should spend the same budget. I think it should, but it is your call to make explicit rather than inherit.

No rush and nothing here is a criticism of the approach. Happy to answer questions on any of it.

vavallee:
Checking in @IfritDMC. Main has moved again since the 6th and the three items above still stand (the rebase, the quota delegation, and the two rows my one line push did not cover). Happy to take the rebase myself if you would rather, just say so. Otherwise I will leave this open for you.

vavallee:
Hi @IfritDMC, a quick update since this has been quiet for a while and main keeps moving under it.

I've taken the automatic search half of your work into #2657: the title split, so a "translated / original" title is searched by its translated part. It's adapted from your `splitLocalizedOriginalTitle`, reduced to a single `SearchTitle` function, and you're credited as co-author on the commit and in the changelog. That's the part that fixes the "finds nothing" symptom, and it's small enough to keep current.

I've left out the original language fallback for interactive search, `manualOnly`, the candidate list and the new API fields. They were well built, but nobody has asked for the fallback yet, and it's the part that makes the rebase hard (the query cap and the language filter issue from my earlier comments). I'll close this one once #2657 merges. If you'd like to bring the fallback back as its own PR on top of that, I'd be glad to look at it.

Thanks for the careful work here.

vavallee:
Closing this in favour of #2657, which just merged and does what this PR set out to do: a "translated / original" title is searched by its translated half.

Thanks for the original work and for the discussion here, IfritDMC. Your approach is credited as a co-author on the merged commit, and the split is narrower than the first draft: it only applies when the slash is not inside brackets or parentheses, both halves carry letters or digits, and there is evidence of a translation, either the original title matching one half or the book language naming a non English language.

Issue #2053, opened 2026-08-17
Grabs recovered by host match leave IndexerID nil, so the download row loses indexer attribution and its seed-ratio override


vavallee:
Follow-up to #2038 (which closed #2039).

`signNZBURL` recovers the right indexer by matching the download URL's host and signs the URL with that indexer's key. It returns only the URL, though, so `indexerID` in `QueueHandler.grab` stays nil for exactly the callers the fallback exists for: API clients that post `{guid, nzbUrl}` with no id.

Two things downstream read that variable:

- `internal/api/queue.go` — `IndexerID: indexerID` on the persisted `models.Download`, so the row has no indexer attribution and the queue cannot say where the release came from.
- `internal/api/queue.go` — `SeedRatio: h.resolveSeedRatio(ctx, indexerID)`, which returns nil on a nil id, so a per-indexer seed-ratio override is silently skipped for these grabs.

Neither breaks the download, which is why it did not show up in #2039. The fix is to have the fallback hand back the indexer it resolved and assign it, keeping the existing behaviour where two indexers on one host disagree on the key (no match chosen, nothing signed, id left nil).

The end-to-end tests in the closed #2044 cover this case and can be lifted rather than rewritten.

Issue #2256, opened 2026-08-27
autoGrab.enabled is enforced per caller, so three grab paths still bypass it


vavallee:
### Summary

`autoGrab.enabled` is enforced by each caller rather than at the point where a grab
actually happens, so every new dispatch site starts life unguarded and stays that way
until someone notices. #2242 was one instance of this, found by a user whose library
downloaded 14 releases with the switch off. Fixing series fill closed that instance and
left the pattern intact.

There are **nine** non-test call sites of `Scheduler.SearchAndGrabBook`
(`internal/scheduler/scheduler.go:595`, reached through the `BookSearcher` interface in
`internal/api/helpers.go:19`). Three of them still do not consult the switch.

### Unguarded today

| Site | Enclosing function | Reached from |
| --- | --- | --- |
| `internal/api/bulk.go:375` | `BulkHandler.fanOutSearches` | `bulk.go:337`, `:459`, `:531` |
| `internal/api/authors.go:3287` | `AuthorHandler.AddBook` | add-book request path |
| `internal/api/recommendations.go:248` | `RecommendationHandler.Add` | add-from-recommendations |

`internal/api/bulk.go` and `internal/api/recommendations.go` contain no reference to
`autoGrab` at all. Verified with `grep -c autoGrab` on both files: zero.

### Guarded today

`internal/api/series.go:558` (via `autoGrabEnabled`, added in #2251),
`internal/api/books.go:600`, `internal/scheduler/scheduler.go:1036` and `:1305`, and the
author catalogue sync, where `internal/api/authors.go:2010` folds
`h.isAutoGrabEnabled(ctx)` into `autoSearchEnabled` before the queue is built, which
covers `runBookSearches` at `:2701`.

### Why this keeps happening

The switch means "do not start downloads without me asking". That is a property of
grabbing, not a property of any particular handler, but it is currently re-derived at
each call site. A caller added later has to know the rule exists. Series fill did not,
and neither did bulk, add-book or recommendations.

### Proposed fix

Enforce it inside `SearchAndGrabBook` itself, so no caller can bypass it:

- `Scheduler.SearchAndGrabBook` reads the switch once and returns without dispatching
  when it is off, logging at Info the way the sweep already does.
- Callers keep marking books wanted and monitored. That is correct in every path today
  and matches what the scheduler does; only the grab stops.
- Remove the now-redundant per-caller checks, or leave them as cheap early exits, but the
  authoritative check is the one at the bottom.

[...]

PR #2397, opened 2026-09-05
perf(scheduler): hoist sweep-invariant loads out of the per-book search and fix the grab history indexer id (#2370) (#2368)

### LINKED ISSUE #2368: Grab history records the requested indexer id, not the one the download row actually got
### Current behaviour

`QueueHandler.grab` resolves the indexer id twice and records the wrong one. At `internal/api/queue.go:931-937` the `signNZBURL` host-match fallback (#2053) overwrites the local `indexerID` when the caller did not supply one:

```go
nzbURL, hostMatched := h.signNZBURL(ctx, req.NZBURL, indexerID)
if hostMatched != nil {
    indexerID = hostMatched
}
```

The download row is then built with `IndexerID: indexerID` (`:942`), which is right. But the grab history event twelve lines later still reads the raw request field (`internal/api/queue.go:1003-1008`):

```go
h.recordHistory(ctx, models.HistoryEventGrabbed, req.Title, bookID, map[string]any{
    "guid":      req.GUID,
    "size":      req.Size,
    "indexerId": req.IndexerID,
})
```

### What a user sees

An API caller that omits `indexerId` (which is exactly the case the #2053 fallback exists to serve) gets a correct download row and a history row whose `indexerId` is null. History's indexer attribution for those grabs is simply missing, with no hint that the information was known at the time.

### Fix

`req.IndexerID` becomes `indexerID` on that one line.

### Sequencing

Not patch-release material on its own; it is a wrong value in an audit trail, not wrong behaviour. Batched with the sweep-invariant hoisting work because both are one-line-ish scheduler and queue changes with no user-facing surface. Nothing blocks it.

### LINKED ISSUE #2370: Wanted sweep reloads sweep-invariant tables once per book, and queries downloads once per in-flight state
Two halves of the same shape: the wanted sweep reloads tables that cannot change during the sweep, once per book, plus once per in-flight state.

### E4: `SearchAndGrabBook` reloads sweep-invariant tables per book

`searchAndGrabFormat` (`internal/scheduler/scheduler.go:688`) is called once per format per book from `searchWanted` (`:1090`). Per call it loads:

| line | load |
|---|---|
| `:695` | `s.indexers.List(ctx)` |
| `:705` | `s.settings.Get(ctx, "search.preferredLanguage")` |
| `:806` | `s.blocklist.List(ctx)` — the whole table |
| `:830` | `s.delayProfiles.List(ctx)` |

None of them can change meaningfully mid-sweep, and the blocklist grows monotonically, so a install with a long blocklist pays O(wanted books x blocklist rows) of decoding per tick just to rebuild the same `decision.NewBlocklistedSpec` over and over.

### E0: `wantedSearchQueue` issues one `ListByStatus` per in-flight state

`internal/scheduler/scheduler.go:1127-1139` loops `inFlightDownloadStates` and runs a separate query per state. Seven states, seven full scans of `downloads`, every sweep, to build one map.

```go
for _, st := range inFlightDownloadStates {
    active, derr := s.downloads.ListByStatus(ctx, st)
    ...
}
```

`DownloadRepo` has no multi-status list method; one `WHERE status IN (...)` would do it in a single query.

### Fix shape

- Load the four invariants once in `searchWanted` and pass them down (a small sweep-context struct is enough). `SearchAndGrabBook` has other callers (on-add hooks, the stall re-search, the API), so its exported signature has to keep working: load lazily when no context is supplied.
- Add `DownloadRepo.ListByStatuses(ctx, ...models.DownloadState)` in `internal/db/downloads.go` and use it in `wantedSearchQueue`.

Not included here: `blockStaleImportFailures` (`internal/importer/scanner.go:950`) loads the whole downloads table once per enabled client per 15s poll. Same shape, different package, worth its own pass.

### Sequencing

Tech debt, no behaviour change, next minor rather than a patch. Touches `internal/scheduler/scheduler.go` alongside #2365 and #2367, so it should merge after both of those. Nothing depends on it.

vavallee:
Three loads the wanted sweep repeated that it only ever needed to do once.

`searchWanted` builds a `sweepContext` holding the indexer list, the whole blocklist, the delay profiles and `search.preferredLanguage`, and hands the same snapshot to every book. The blocklist grows monotonically, so the old shape cost O(wanted books x blocklist rows) of decoding per tick just to rebuild the same spec. `SearchAndGrabBook` keeps its exported signature: its other callers (the on-add hook, the stall re-search, the API) pass no snapshot and each value is loaded lazily exactly as before, which is also what keeps a one-off search reading fresh state.

`wantedSearchQueue`'s seven `ListByStatus` calls collapse into one `DownloadRepo.ListByStatuses` over the same states.

And the one-word fix next door: the grab history event recorded `req.IndexerID` rather than the resolved `indexerID`, so a caller that omits the id got a correct download row and a null indexer in history, which is precisely the case the #2053 host-match fallback exists to serve.

No behaviour change in the sweep itself. The one distinction worth naming is that an empty blocklist still builds the spec while an unreadable one does not, matching what the per-book `err == nil` check used to do; `sweepContext.blocklistLoaded` carries that.

Closes #2370
Closes #2368

### Testing

- `TestSweepContext_ServesEveryBookFromOneLoad` builds the snapshot, detaches every repo, and asserts all four reads still answer correctly. An accessor that went back to the database would return empty, so this is the "one load per sweep, not one per book" assertion in a form that does not need a query counter.
- `TestSweepContext_NilLoadsPerCall` pins the other half: no snapshot means each value comes from its repo, and a Scheduler with no repos at all degrades to empty rather than panicking.
- `TestSweepContext_UnreadableBlocklistSkipsTheSpec` pins empty-vs-unreadable.
- `TestDownloadRepo_ListByStatuses` covers the union, the exclusions, agreement with `ListByStatus` for one status, and the no-statuses case.
- `TestQueueGrab_HistoryRecordsResolvedIndexerID` is the fail-before/pass-after for the history id. It grabs with no `indexerId` in the body, checks the download row got the resolved id first (so it cannot pass vacuously), then reads the history payload:

```
--- FAIL: TestQueueGrab_HistoryRecordsResolvedIndexerID (0.23s)
    queue_grab_history_indexer_test.go:89: grab history recorded a null indexerId; want 1 (the id the download row got)
```

Ran: `go build ./...`, `go vet` and `go test` over `internal/scheduler`, `internal/db`, `internal/api` (all ok), `golangci-lint run` over the same three (0 issues).

### Sequencing

Tech debt with no user-visible surface, so next minor rather than a patch. Merge after #2385 and #2388, which touch `internal/scheduler/scheduler.go` ahead of it; the overlap with #2385 is the `SearchAndGrabBook` / `searchAndGrabFormat` signatures, which both PRs extend for different reasons. Nothing depends on this.


vavallee:
Merged `main` in. This was conflicting because #2385 landed on main and restructured the same two functions this branch restructured. Seven hunks, four in `scheduler.go` and three in `scheduler_autograb_test.go`.

Both changes survive. Nothing was dropped from either side.

**`scheduler.go`**

1. `SearchAndGrabBook` / the chokepoint function. Main split the book level entry point into `neededFormats` plus `searchAndGrabFormats(ctx, book, formats)` and moved the #2256 kill switch into it. This branch had renamed the same body to `searchAndGrabBook(ctx, book, sweep)`. Kept main's shape and added the snapshot parameter to it: `searchAndGrabFormats(ctx, book, formats, sweep)`. The branch's `searchAndGrabBook` wrapper is gone, it only existed to carry the snapshot and `searchAndGrabFormats` now does that. `SearchAndGrabBook` calls it with `neededFormats(&book)` and a nil snapshot, so every one off caller still loads lazily.
2. The dispatch loop. Main's `for _, mediaType := range formats` won over the branch's two `NeedsEbook` / `NeedsAudiobook` ifs, with `sweep` passed down to `searchAndGrabFormat`.
3. The sweep fan out. Kept both: `newSweepContext` is still loaded once before the fan out, and the fan out still iterates `wantedSearch` values and passes `w.formats`.
4. The in flight filter. Main replaced the branch's `map[int64]bool` with the per format `inFlightFormatsByBook`, but built it by looping `ListByStatus` once per state, which is exactly the query pattern #2370 removed. Kept main's per format map and rebuilt it from the single `ListByStatuses(ctx, inFlightDownloadStates...)` call. So seven queries per sweep are still one query, and the map is still per format.

**`scheduler_autograb_test.go`**

All three hunks are the drift guard naming the chokepoint. Main's side wins outright, not as a preference but because it is now correct: the chokepoint really is `searchAndGrabFormats`, so the guard reads that function's body, asserts `autoGrabEnabled` is called before the first `searchAndGrabFormat(` dispatch, and errors on any call of `searchAndGrabFormat` in the package outside it. The branch's variant named `searchAndGrabBook`, which no longer exists, so it is genuinely superseded rather than merged. Added a line to the guard's comment noting the function also carries the per sweep snapshot now, since that changes its signature but not the property. Confirmed the property itself still holds and not just the test: `searchAndGrabFormats` is the only caller of `searchAndGrabFormat` in production code, and the switch check is still ahead of the loop.

One non conflicted follow up: `scheduler_wanted_queue_test.go` called `searchAndGrabFormats` with three arguments, updated to pass a nil snapshot.

**Verification**

```
go build ./...                     clean
go vet ./...                       clean
golangci-lint run ./internal/{scheduler,api,db}/...   0 issues
go test ./internal/scheduler/... ./internal/api/... ./internal/db/... -count=1
  ok  internal/scheduler  7.7s
  ok  internal/api      129.2s
  ok  internal/db        25.4s
```

Then both sides' own tests, run by name, to prove neither behaviour was quietly lost. All present, all passing:

- #2385's per format sweep: `TestWantedSearchQueue_ParkedFormatLeavesSiblingSearchable` (importHeld and importExternal), `TestWantedSearchQueue_SingleFormatDownloadingStillSkipped`, `TestWantedSearchQueue_BothFormatsInFlightSkipped`, `TestWantedSearchQueue_UndeterminedFormatBlocksWholeBook`, `TestWantedSearchQueue_FormatOnDiskNotSearched`, `TestWantedSearchQueue_SingleFormatBookIgnoresReleaseTokens`, `TestSearchAndGrabFormats_SearchesOnlyGivenFormats`, `TestDownloadMediaType`.
- This branch's hoisting: `TestSweepContext_ServesEveryBookFromOneLoad`, `TestSweepContext_NilLoadsPerCall`, `TestSweepContext_UnreadableBlocklistSkipsTheSpec`.
- #2256: `TestSearchAndGrabBook_HonoursAutoGrabSwitch` and `TestAutoGrabChokepointHolds`.

The one place the two changes genuinely disagreed rather than merely overlapped was hunk 4. Main's per format map needed a per download row, and the branch's single query returns exactly those rows, so the disagreement was in how the rows are fetched, not in what the filter decides. Everything else was two edits to the same lines with compatible intent.

PR #2398, opened 2026-09-05
fix(auth): sign the OIDC flow cookie and bind it to its provider (#2362)

### LINKED ISSUE #2362: OIDC flow state cookie is unsigned base64 JSON, so a cookie writer can sign the victim in as themselves
The `bindery_oidc_flow` cookie is base64 JSON with no signature:

```go
func encodeFlowStateRaw(fs flowState) (string, error) {
    b, err := json.Marshal(fs)
    ...
    return base64.RawURLEncoding.EncodeToString(b), nil
}
```

`internal/auth/oidc/provider.go:687-712`. `DecodeFlowState` base64 decodes, unmarshals, checks `Expiry`, and returns. The struct doc at `:649` calls it "the encrypted-at-rest blob", which it is not.

Everything the callback trusts comes out of that cookie. The CSRF check at `internal/api/auth_oidc.go:250` is `r.URL.Query().Get("state") != fs.State`, and both halves of that comparison are attacker suppliable if the attacker can set a cookie on the origin. A sibling subdomain or one plain HTTP hop is enough; cookies are not origin scoped the way storage is. The callback is in `AllowUnauthPath` and the cookie is `SameSite=Lax`, which is sent on top level navigation, so the whole thing is reachable by sending the victim a link.

Attack: attacker starts a login against the IdP as themselves, keeps their own `state`, `nonce`, `code_verifier` and authorization `code`, writes a flow cookie carrying those values into the victim's browser, then navigates the victim to the callback URL with the matching `?state` and `?code`. The victim's browser completes the exchange and gets a Bindery session for the attacker's IdP identity. From there anything the victim saves goes into an account the attacker also controls.

Two smaller problems in the same payload:

* `Expiry` is forgeable, so the ten minute replay window is not a window.
* `providerID` is not in the payload at all. The callback reads it from the URL (`internal/api/auth_oidc.go:216`) and never checks it against the flow, so a flow started for provider A completes against provider B.

The primitive already exists next door: `auth.MakeCSRFToken` and `auth.SignSessionWithEpoch` in `internal/auth`, both HMAC-SHA256 over the session secret, with `sessionSecrets()` already returning the ordered `{current, previous}` set for rotation.

Fix shape: append an HMAC over the payload keyed by the current session secret, verify against every configured secret on decode, and bind `providerID` into the signed payload so the callback can reject a mismatch. Cookie name and path stay as they are. In flight logins at deploy time will fail once and work on retry, which is the same thing a restart does today.

### Sequencing

Patch release material: this is a pre auth login CSRF, and the only thing standing between it and a working exploit is the attacker's ability to write one cookie on the origin.

Biggest change of the security batch, so it should land after the S1 plus S2 plus S5 PR (three one liners, all reachable by the lowest privilege account) rather than holding that one up behind this review.

vavallee:
The `bindery_oidc_flow` cookie was bare base64 JSON. Everything the callback trusts comes out of it: the `state` it compares `?state` against, the PKCE verifier, the expiry. Anyone able to write a cookie on the origin (a sibling subdomain, one plain HTTP hop) could start a login against the IdP as themselves, write the resulting flow into the victim's browser, and walk them onto the callback with the matching `?state` and `?code`. The victim ends up signed in as the attacker's IdP identity, and anything they save afterwards lands in an account the attacker also controls. The callback is in `AllowUnauthPath` and the cookie is `SameSite=Lax`, so a link is enough.

The value is now `<payload>.<hmac>`, HMAC-SHA256 over the encoded payload keyed by the install's session secret under its own domain-separation label (`bindery-oidc-flow.v1.`), so a flow blob can never be replayed as a session cookie or a CSRF token and vice versa. Two smaller holes closed with it: the signature is verified before the payload is parsed, so a forged `Expiry` no longer removes the ten-minute replay window and no attacker bytes reach the JSON decoder; and the provider id is baked into the signed payload and compared at the callback, so a flow started for one IdP cannot be completed against another.

### The shape, since the issue asked

Rotation and fail-closed behaviour follow what the rest of the auth stack already does. `DecodeFlowState` takes the ordered `{current, previous}` candidate set that `sessionSecrets()` already returns, so a login started just before a rotation still completes, and it accepts only a value some candidate reproduces in constant time. A missing or short secret is refused at both ends rather than treated as "sign with nothing"; the 32-byte minimum matches the session package's own.

Plumbing turned out not to be invasive. `OIDCHandler` already holds the `*AuthHandler`, and `sessionSecret` / `sessionSecrets` are unexported methods in the same package, so the secrets are passed in as arguments (`EncodeFlowState(secret, providerID, ...)`, `DecodeFlowState(secrets, providerID, ...)`) and `internal/auth/oidc` gains no dependency on the settings repo. Two thin nil-safe accessors on `OIDCHandler` are the whole of it.

Cookie name and path are unchanged. The callback now returns one flat "invalid flow state" for every server-side refusal and logs which check failed, so an operator can tell an upgrade-in-flight cookie (`ErrFlowStateUnsigned`) from a tampered one without the caller learning anything.

### Upgrade note

An unsigned cookie from before this change fails verification, so a login already in flight at deploy time fails once and works on retry. That is what a restart costs today. `docs/auth-oidc.md` says so.

Closes #2362

### Testing

* `TestCallback_ForgedFlowCookieRejected` in `internal/api` is the exploit at the handler: an attacker-built unsigned cookie plus a matching `?state` and `?code`. With `DecodeFlowState` reverted to pre-fix behaviour it returns **302 with a session issued**; with the fix it is 400 and no session cookie.
* Unit coverage in `internal/auth/oidc`: tampered payload under a real signature, a flow signed with a key the install does not hold and an expiry a century out, provider mismatch, valid round trip, previous-secret-still-verifies (and stops once dropped), and fail-closed on a nil or short secret.
* `go test -count=1 ./internal/api/ ./internal/auth/...` green. `go vet` and `golangci-lint run` clean on both.

### Sequencing

Patch release material. It is a pre-auth login CSRF, and the only thing between it and a working exploit is the attacker's ability to write one cookie on the origin.

It is also the largest change in the security batch, so **land it after the S1 + S2 + S5 PR** rather than holding three one-liners behind this review. It does not conflict with that PR, or with #2380 (migrate SSRF) or #2389 (guarded dials); #2389 touches `internal/auth/oidc/discover.go` while this touches `provider.go`, so merge order between them does not matter.


vavallee:
Pulling this out of v1.33.4. The crypto here is right, but the thing it says it fixes is not fixed, and I would rather ship nothing than announce this one.

Signing stops an attacker forging a flow cookie. Against the attacker #2362 describes, the one who can write a cookie on the origin, forging was never necessary. They can just ask for a real one.

`/api/v1/auth/oidc/{provider}/login` is public. `AllowUnauthPath` lets through anything under `/api/v1/auth/oidc/` ending in `/login` or `/callback`, which it has to, because the IdP round trip happens before anyone holds a session. That endpoint mints the flow cookie and hands it back in a `Set-Cookie`, with the state and nonce in the `Location`. So:

1. `curl -i https://host/api/v1/auth/oidc/google/login` with no auth at all, keep the cookie and the state
2. authenticate at the IdP as yourself, grab the `?code`, do not complete the callback
3. plant that cookie in the victim's browser with the same primitive the issue already grants
4. send them the callback link

The cookie is genuine, so the signature verifies, the provider binding matches, and the expiry is fresh. The PKCE verifier is inside the cookie, so the exchange goes through. Victim ends up signed in as your IdP identity, which is the whole of #2362. Reviewed on the branch and it still returns 302 with a session issued.

What the PR does genuinely fix, and these are worth keeping:

- forged expiry, since the MAC is checked before the payload is parsed, so the ten minute window is a real window now
- a flow started at one provider completing at another
- unauthenticated bytes reaching `json.Unmarshal`
- the old `"invalid flow state: "+err.Error()` echoing a parse error over attacker controlled bytes back to the caller

So the work is not wasted, the framing is just wrong. Three ways forward, cheapest first:

- **`__Host-` prefix on the flow cookie.** This is the one that actually matches the threat model, because it kills both write primitives in the issue: a sibling subdomain cannot set a `Domain` scoped cookie and a plaintext hop cannot set it at all. It forces `Path=/` and `Secure` though, so it cannot apply on plain HTTP installs and would need to key off `cookieSecure`.
- **A confirmation step at the callback**, "continue signing in as attacker@example.com?". Standard login CSRF mitigation and the only one that does not lean on cookie integrity.
- **Neither, and restate the severity honestly.** Keep the hardening, drop the CSRF claim from here, from #2362 and from `docs/auth-oidc.md`.

Leaving this open rather than merging it. Reopening the severity question on #2362 too, since it currently reads as closed by this.

vavallee:
Unblocked. Added the `__Host-` prefix in `ddcfa881`, which is option 1 from my earlier comment, so this PR now closes what it says it closes.

**What changed.** On HTTPS the flow cookie is `__Host-bindery_oidc_flow`, `Path=/`, `Secure`, no `Domain`. The callback reads **only** that name on a secure install. Accepting the unprefixed name as a fallback would have left the hole open while looking fixed, because the unprefixed name is exactly what a sibling subdomain or a plaintext hop can still write.

**Fail-before evidence**, since that was the whole problem with the original framing. Adding a legacy-name fallback to the callback and re-running:

```
--- FAIL: TestCallback_SecureInstallRefusesLegacyCookieName
    a planted bindery_oidc_flow cookie completed the flow on an HTTPS install; #2362 is still open
```

That test presents a genuine, correctly signed cookie under the old name over HTTPS, which is precisely how the attacker plants one. It has to be refused, and now is.

**Three costs, stated rather than buried:**

- `__Host-` forces `Path=/`, so on HTTPS the cookie rides every same-origin request instead of only `/api/v1/auth/oidc`. Still HttpOnly, signed, ten minutes. The spec does not allow a narrower path.
- One failed login for anyone mid-flow across the upgrade. Their cookie has the old name, they get "missing flow cookie" and press the button again. Ten-minute window, one retry.
- **Plain HTTP installs cannot use the prefix at all** and stay exposed. `docs/auth-oidc.md` now says that outright and points at TLS or `BINDERY_COOKIE_SECURE=always`, rather than implying #2362 is closed everywhere.

Both cookie names are expired at the callback so a pre-upgrade cookie is not left sitting unread in a browser.

The hardening from the first commit stands on its own and is unchanged: MAC checked before the payload is parsed (so a forged expiry no longer removes the replay window), provider binding, no attacker bytes reaching `json.Unmarshal`, and no parse error echoed back to the caller.

Five new tests in `auth_oidc_host_cookie_test.go` cover the attribute set a browser silently drops on, both accept paths, the refusal, and the double clear. `go vet`, `golangci-lint run ./internal/api/` at 0 issues, `internal/api` OIDC suite green.

PR #2406, opened 2026-09-05
fix(web): offer the relink metadata action for every author

### LINKED ISSUE #2329: "Find better metadata" is hidden for any linked author with a bio or cover, so relinking a provider is API-only
Line numbers are `main` at `c6895aed`.

## What is wrong

`AuthorDetailPage.tsx:466` decides whether the metadata action appears at all:

```ts
const showMetadataLinkAction = canLinkAuthorMetadata(author) || hasSparseMetadata(author)
```

`hasSparseMetadata` (`web/src/util/authorMetadata.ts:19`) returns false as soon as the author has any one of description, image URL, disambiguation, ratings count or average rating. So an author who is linked to a provider and has a bio *or* a cover photo gets no menu item, and there is no way to reach the relink flow from the UI.

That is backwards for the case the action exists to serve. An author record with a bio and a photo but a wrong or thin catalogue is exactly the one a user wants to move to another provider. An author record with nothing at all is arguably the one where it matters least, since a refresh may fill it. The current condition offers the action only in the second case.

The backend has no such restriction. `AuthorHandler.RelinkUpstream` (`internal/api/authors.go:1206`) accepts any author, and its only guard is `canRelinkAuthorToUpstream` on the no-body form (`:1249`), which is bypassed entirely when an explicit `foreignAuthorId` is supplied. The candidate search behind the modal (`Aggregator.SearchAuthorCandidates`, `internal/metadata/aggregator_author_candidates.go:15`) queries every configured provider and deliberately does not merge same-person records, precisely so a specific provider record can be chosen. All of that machinery is reachable only by curl for the affected authors:

```
GET  /api/v1/author/{id}/relink-upstream/candidates?term=<name>
POST /api/v1/author/{id}/relink-upstream  {"foreignAuthorId":"hc:1234","authorName":"..."}
```

## Repro

1. Add an author whose OpenLibrary record has a bio or a photo.
2. Open the author page, open the actions menu.
3. Neither "Link metadata" nor "Find better metadata" is present.
4. `POST /api/v1/author/{id}/relink-upstream` with a Hardcover `foreignAuthorId` succeeds against that same author.

## Suggested fix

Show the action unconditionally and keep `hasSparseMetadata` for the label only, which is what it reads naturally as:

```ts
const showMetadataLinkAction = true
const metadataLinkLabel = canLinkAuthorMetadata(author)
  ? t('authorMetadataLink.actionLink', 'Link metadata')
  : t('authorMetadataLink.actionFindBetter', 'Find better metadata')
```

`canLinkAuthorMetadata` still picks the wording, so an `abs:`/`calibre:` author keeps "Link metadata" and everyone else gets "Find better metadata". `AuthorDetailPage.test.tsx:297` covers the sparse case; a linked non-sparse author needs a case of its own.

If the action should stay hidden somewhere, the honest condition is "the author has no provider identity to change", not "the author record looks thin".

## Origin

Reported by omegaliterosmalakas on Discord (#support, "How to rebind an author to hardcover to get better metadata (especially series)", 2026-09-04), asking how to move an OpenLibrary-imported library to Hardcover. Not yet confirmed by them that the button is missing on their instance; the code path is clear enough to file either way. The larger half of what they were asking for is #2328.

alorentiar:
## What

The author page's actions menu only offered "Link metadata" / "Find better
metadata" when `canLinkAuthorMetadata(author) || hasSparseMetadata(author)`.
`hasSparseMetadata` returns false as soon as the author has any one of a
description, image, disambiguation or ratings, so a provider-linked author
with a bio or cover photo got no menu item at all. The relink flow
(`POST /api/v1/author/{id}/relink-upstream`) stayed reachable only via the
API for exactly the records users most want to move to a better provider.

## Why

The backend has no such restriction: `AuthorHandler.RelinkUpstream` accepts
any author, and the candidate search queries every configured provider
without merging same-person records, so a specific provider record can always
be chosen. An author record with a bio and photo but a wrong or thin
catalogue is precisely the case the action exists to serve; the old condition
offered it only for records with nothing at all.

## Change

- `web/src/pages/AuthorDetailPage.tsx`: always include the metadata action in
  the actions menu. The wording is still picked by `canLinkAuthorMetadata`
  ("Link metadata" for `abs:`/`calibre:` authors, "Find better metadata"
  otherwise), and `hasSparseMetadata` is no longer imported by the page.
- `web/src/pages/AuthorDetailPage.test.tsx`: new case asserting that a
  provider-linked author with a full record (description, image,
  disambiguation, ratings) sees the "Find better metadata" menu item.

## Verification

- `npx vitest run src/pages/AuthorDetailPage.test.tsx` — 36 passed
  (35 prior + 1 new). The new test fails on the old body:
  `TestingLibraryElementError: Unable to find role="menuitem" and name "Find better metadata"`.
- `npx tsc --noEmit` — clean.
- `npx eslint src/pages/AuthorDetailPage.tsx src/pages/AuthorDetailPage.test.tsx` — no issues.
- Full frontend suite: `make test-web` — 659 passed (67 files); `make lint-web` — 0 errors (9 pre-existing warnings).

Closes #2329

### DIFF
--- a/web/src/components/AddAuthorModal.test.tsx
+++ b/web/src/components/AddAuthorModal.test.tsx
@@ -294,6 +294,50 @@ describe('AddAuthorModal — search error handling', () => {
     alertSpy.mockRestore()
   })
 
+  it('offers find metadata when the conflicting canonical author is already linked with a full record', async () => {
+    vi.mocked(api.searchAuthors).mockResolvedValue([
+      author({
+        id: 0,
+        foreignAuthorId: 'OL13200512A',
+        authorName: 'Emilia Jae',
+      }),
+    ])
+    const err = Object.assign(new Error('author already exists'), {
+      status: 409,
+      body: {
+        error: 'author already exists',
+        canonicalAuthorId: 60,
+        canonicalAuthor: author({
+          id: 60,
+          foreignAuthorId: 'OL13200512A',
+          authorName: 'Emilia Jae',
+          metadataProvider: 'openlibrary',
+          description: 'A bio',
+          imageUrl: 'https://example.com/emilia.jpg',
+          ratingsCount: 12,
+          averageRating: 4.1,
+        }),
+      },
+    })
+    vi.mocked(api.addAuthor).mockRejectedValue(err)
+    const alertSpy = vi.spyOn(window, 'alert').mockImplementation(() => {})
+
+    render(<AddAuthorModal onClose={onClose} onAdded={onAdded} />)
+
+    fireEvent.change(screen.getByPlaceholderText('Search by author name...'), {
+      target: { value: 'emilia jae' },
+    })
+    fireEvent.click(screen.getByRole('button', { name: /^search$/i }))
+    await waitFor(() => expect(screen.getByText('Emilia Jae')).toBeInTheDocument())
+
+    fireEvent.click(screen.getByRole('button', { name: 'Select' }))
+    fireEvent.click(screen.getByRole('button', { name: 'Add author' }))
+
+    await waitFor(() => expect(screen.getByText('author already exists')).toBeInTheDocument())
+    expect(screen.getByRole('link', { name: 'Find metadata' })).toHaveAttribute('href', '/author/60?linkMetadata=1')
+    alertSpy.mockRestore()
+  })
+
   it('loads global monitor defaults and lets the backend apply them when unchanged', async () => {
     vi.mocked(api.getSetting).mockImplementation(async (key: string) => {
       if (key === 'default.media_type') return { key, value: 'ebook' }
--- a/web/src/components/AddAuthorModal.tsx
+++ b/web/src/components/AddAuthorModal.tsx
@@ -3,7 +3,7 @@ import { useTranslation } from 'react-i18next'
 import { api, AddAuthorRequest, Author, AuthorConflictBody, AuthorMonitorMode, MediaType, MetadataProfile, RootFolder } from '../api/client'
 import { splitAuthorSearchResults } from './addAuthorTitleGuard'
 import { useNeedsSetup } from './useNeedsSetup'
-import { authorProviderKey, canLinkAuthorMetadata, hasSparseMetadata } from '../util/authorMetadata'
+import { authorProviderKey } from '../util/authorMetadata'
 import { providerDisplayName } from '../util/metadataSource'
 
 interface Props {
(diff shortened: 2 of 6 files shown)


vavallee:
This is right, and the reasoning in the description is the same reasoning I filed the issue with. Two things before it can go in.

**The Sign-off check is failing.** The repo runs a DCO gate and your commit has no `Signed-off-by` trailer. I cannot add it for you, since it is your attestation rather than a formatting detail. On your branch:

```
git commit --amend -s --no-edit
git push --force-with-lease
```

That is the only blocker.

**The same gate exists in one more place**, and I would rather you got the credit for finishing it than leave it half done. `AddAuthorModal.tsx:203`:

```ts
const showConflictFindMetadata = addConflict?.canonicalAuthorId && (canLinkAuthorMetadata(addConflict.canonicalAuthor) || hasSparseMetadata(addConflict.canonicalAuthor))
```

That is the add author conflict path, where you have tried to add an author that already exists and the modal offers to go find better metadata for the existing record. Identical condition, identical consequence: the offer disappears exactly when the existing record has a bio or a cover, which is the case you most want it in.

My issue only named `AuthorDetailPage.tsx`, so you were right to scope to it. But if you drop the condition there too, `hasSparseMetadata` and its tests come out with it and the helper stops existing rather than surviving as the one caller nobody noticed. Happy either way: extend this PR, or say no and I will file the follow up.

Nice test, by the way. Building it so it fails on the old condition is the part people skip.

alorentiar:
Thanks for the thorough review, and for pointing at the conflict path. I took both:

- Amended the original commit with the Signed-off-by trailer (force-pushed with lease).
- Extended the fix to AddAuthorModal: the conflict offer now shows for any canonical author, and hasSparseMetadata is gone entirely along with its tests, as you suggested.

The new modal test builds a canonical author that is already linked to OpenLibrary with a full record, which is the case the old condition hid; it fails on the previous code. Local runs are green: 58 tests across the modal, util and page suites, eslint and tsc clean.

The Container Scan failure looks unrelated to this change (frontend only), but happy to take a look if it is not a known issue.

alorentiar:
Hi, quick check-in on this. Both review items are in (the DCO sign-off on the amended commit and the AddAuthorModal conflict path fix), and the full check suite is green now, including the container scan. Anything else you would like changed, or is this waiting on something on my side?

vavallee:
Merged, thanks for sticking with it. Sorry it sat a few days.

alorentiar:
Thank you

PR #2407, opened 2026-09-05
perf(indexer): stop compiling regexes on every cache hit

### LINKED ISSUE #2341: phraseRegex and inOrderRegex compile a regex on every call, even on a cache hit

v01dst:
## Summary

`phraseRegex` and `containsInOrder` both cached with:

```go
re, _ := regexCache.LoadOrStore(pattern, regexp.MustCompile(pattern))
```

Go evaluates arguments before the call, so `MustCompile` ran on **every** invocation and the freshly compiled regex was discarded on a cache hit — the cache saved a map entry and nothing else (#2341).

## Fix

Switch both call sites to the Load-then-Store pattern `WordBoundaryRegex` already uses (`internal/indexer/release.go:111`):

- check the cache first, return on hit
- compile only on a miss, then `Store`

`ContainsPhrase` runs once per release inside the search filter loop, so a 500-result search previously compiled the same regex 500 times and threw away 499; fan-out sweeps filtering thousands of releases across indexers felt this the most.

`go build ./internal/indexer/` and `go test ./internal/indexer/` pass.

Fixes #2341

### DIFF
--- a/internal/indexer/inorder.go
+++ b/internal/indexer/inorder.go
@@ -32,6 +32,10 @@ func inOrderRegex(seq []string) *regexp.Regexp {
 	// separator for the next word to assert on, which a bare `.*` would break.
 	gap := wordSep + `(?:.*` + wordSep + `)?`
 	pattern := `(?i)(?:^|` + wordSep + `)` + strings.Join(parts, gap) + `(?:` + wordSep + `|$)`
-	re, _ := regexCache.LoadOrStore(pattern, regexp.MustCompile(pattern))
-	return re.(*regexp.Regexp)
+	if v, ok := regexCache.Load(pattern); ok {
+		return v.(*regexp.Regexp)
+	}
+	re := regexp.MustCompile(pattern)
+	regexCache.Store(pattern, re)
+	return re
 }
--- a/internal/indexer/release.go
+++ b/internal/indexer/release.go
@@ -130,8 +130,12 @@ func phraseRegex(phrase []string) *regexp.Regexp {
 	// wordSep rather than \b/\W so non-ASCII words match (#1642). The inner
 	// separator stays a + run, matching the previous \W+ behaviour.
 	pattern := `(?i)(?:^|` + wordSep + `)` + strings.Join(parts, wordSep+`+`) + `(?:` + wordSep + `|$)`
-	re, _ := regexCache.LoadOrStore(pattern, regexp.MustCompile(pattern))
-	return re.(*regexp.Regexp)
+	if v, ok := regexCache.Load(pattern); ok {
+		return v.(*regexp.Regexp)
+	}
+	re := regexp.MustCompile(pattern)
+	regexCache.Store(pattern, re)
+	return re
 }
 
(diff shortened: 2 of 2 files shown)


v01dst:
The two failed checks from the first push are now resolved / explained:

- **DCO** — my commit was missing `Signed-off-by`. Amended and force-pushed (`8690dfd`); the Sign-off check now passes.
- **Security → Container Scan (Grype)** — pre-existing on `main`, not introduced by this PR. The Security workflow failed on `main` 6 hours before this PR was opened (run 33932221598, 2026-09-05T00:12:50Z). My diff is a 6-line source change in `internal/indexer/*.go`; the failing step is the base-image CVE gate (`fail-build: true` with `severity-cutoff: high`), which is sensitive to base-image freshness, not application code.

Current state on `8690dfd`: DCO ✅, CodeQL ✅, gitleaks ✅, Semgrep ✅, Checkov ✅, Hadolint ✅, ESLint ✅, dependency-review ✅, lint ✅, licenses ✅, validate (Go) ✅, validate (frontend) ✅, smoke ✅, grype (in the Security run's job list) — remaining Go race shards were still running at time of writing.

Happy to rebase if the base-image CVEs need a Dockerfile bump to clear the Container Scan gate, though that's likely a separate change.

vavallee:
Nice catch, and thanks for the clean fix. Merging this.

One thing that is my fault rather than yours: I filed #2341 and opened #2378 eight minutes later, and that PR bundles three perf issues so the link back to #2341 is easy to miss. You had no realistic way to see the overlap. I will stop filing issues I am about to fix myself without saying so in the issue.

Your version is the better one for #2341 anyway. It is the whole fix and nothing else, which is what that issue asked for. #2378 also replaces the cache with a bounded one for #2344, so I will rebase it on top of yours and it keeps only the eviction work.

PR #2444, opened 2026-09-06
fix(ci): the ping pin PR cannot merge itself, so stamp its contexts

vavallee:
The pin step added in #2441 opened its PR and then failed:

```
GraphQL: 3 of 3 required status checks are expected. (mergePullRequest)
```

## Why it could never have worked

A pull request opened with `GITHUB_TOKEN` does not trigger workflows, by design. So main's three required contexts (`lint`, `validate (Go)`, `Security Summary`) never report on the pin branch and the PR sits BLOCKED forever. `--admin` does not rescue it either: the default token is not a repo admin.

`deploy-prod` in `ci.yml` hit this first and solved it by stamping the three contexts through the statuses API, waiting for the required set to settle, then merging **without** `--admin`. I copied that job's branch-and-PR shape and missed the part that makes it land.

## The fix

Takes the rest of the deploy-prod pattern: same context list, same wait loop, same no-admin merge, plus `statuses: write` on the job.

Stamping is defensible for the same reason it is there: the diff is one image tag in a deploy manifest, no code, and the image it names was built and had its build sha verified by the stamp job in the same run. A required check failing or the loop timing out still aborts.

## Already handled

I merged the stranded PR (#2442) by hand, so `main` already pins `sha-f1d124a2`, the first build carrying the sha stamping, the Debian 13 base and `x/crypto` v0.56.0. This change is about the next one.

### DIFF
--- a/.github/workflows/ping-server.yml
+++ b/.github/workflows/ping-server.yml
@@ -94,6 +94,7 @@ jobs:
       contents: write # move the image pin in deploy/telemetry-server.yaml
       packages: write # push the image to GHCR
       pull-requests: write # the pin lands through an auto-merged PR
+      statuses: write # stamp the required contexts on the pin PR, see below
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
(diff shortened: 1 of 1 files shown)


vavallee:
Superseded by the branch that removes the pin step entirely. deploy/telemetry-server.yaml was never applied to anything: the live Deployment is managed by an ArgoCD app whose source is the private homelab repo, synced with selfHeal, so nothing this repo writes into that file could ever reach the cluster. There is no pin left to merge.

Issue #1467, opened 2026-07-07
Frontend: AuthorDetailPage and HistoryPage silently truncate at 100 rows


vavallee:
Found during a frontend sweep. Two silent data-truncation bugs of the same class as #1010 (fixed for BooksPage), resurfacing on other pages. Both have an in-repo template (`useServerPagination` + BooksPage) that makes the fix mechanical.

## AuthorDetailPage caps an author's catalogue at 100 books

`web/src/pages/AuthorDetailPage.tsx:153` (also 184, 271) calls `api.listBooks({ authorId, ... })` with no `limit`, and the server applies the default 100 to the by-author path (`internal/api/books.go:250`). The page destructures `{ items }` and discards `total`, with no pagination. Any author with more than 100 works shows a truncated catalogue with no indication, and the status counts, filters, series grouping, and "select all visible" all operate on the wrong set.

## HistoryPage can never show more than the latest 100 events

`web/src/pages/HistoryPage.tsx:74` calls `api.listHistory(...)` with no limit/offset, server defaults to 100 (`internal/api/history.go:44`), `total` is discarded, and the client then runs `usePagination(events, 100)` over the 100 rows: one page, forever. History older than the last 100 events is unreachable, and the event-type filter only offers types present in those 100 rows.

## Suggested fix

Thread server pagination through both (`useServerPagination`, as BooksPage does), or at minimum surface `total` and paginate. CalendarPage has a milder version of the same at `limit: 500` (`web/src/pages/CalendarPage.tsx:41`), which equals the server hard cap.

PR #2458, opened 2026-09-06
feat(web): monitor toggle on the book detail page (#2417)

### LINKED ISSUE #2417: No way to monitor or unmonitor a book from the book page
Reported by Daize in Discord: there is no way to monitor or unmonitor a single book from the book page.

Confirmed, and it is an omission rather than a decision. The book page shows the monitored state and gives you no way to change it.

## Where you can change it today

- **Books list**, tick a book's checkbox and use the bulk **Monitor** / **Unmonitor** action (`web/src/pages/BooksPage.tsx:416-417`). Works for one book, but you have to go back to the list and select it.
- **Wanted page**, "stop monitoring" on a row (`web/src/pages/WantedPage.tsx:107`, `api.updateBook(book.id, { monitored: false })`). One direction only.

So the capability exists and is reachable from two other screens, just not from the page dedicated to the book.

## Why this is small

`PUT /api/v1/book/{id}` already accepts it (`internal/api/books.go:452`, `:468`):

```go
Monitored    *bool     `json:"monitored"`
...
if req.Monitored != nil {
    book.Monitored = *req.Monitored
}
```

And `BookDetailPage` already has the generic save helper wired to that endpoint (`web/src/pages/BookDetailPage.tsx:365`):

```ts
const saveField = async (patch: Partial<Book>) => {
  ...
  const updated = await api.updateBook(book.id, patch)
```

So this is a control calling `saveField({ monitored: !book.monitored })`, not new plumbing.

## Shape

A toggle next to the status badge, which already renders the monitored state at `BookDetailPage.tsx:648` through `bookStatusBadge(book.status, book.monitored, t)`. Mirror the affordance the Books list bulk action uses so the wording matches, and use the `Toggle` component the settings screens already use rather than inventing a new control.

Daize also asked about setting a book to **wanted** from the same page. That is a different thing and mostly should not be a manual control: `books.status` is derived, `refreshBookStatus` computes wanted or imported from whether the book has files. What a user actually wants there is monitoring, plus the existing Search action. Worth confirming with them, but I would not add a manual status setter.

Adiiityaaaaaaaaa:
## Summary

The book page showed whether a book was monitored but gave no way to change it — you had to go
back to the Books list and use the bulk **Monitor** / **Unmonitor** action, or use the Wanted
page's one-way "stop monitoring". This adds a toggle next to the status badge on the book page.

As #2417 notes, the plumbing already existed: `PUT /api/v1/book/{id}` accepts `monitored`, and
`BookDetailPage` already has the generic `saveField` helper wired to it. So this is a control
calling `saveField({ monitored: !book.monitored })`, built from the existing `Switch` component.

- No new i18n keys — the accessible label reuses `common.monitor` / `common.unmonitor`, the same
  strings the Books-list bulk action uses, so the two surfaces can't drift.
- No visible state text on the switch: the status badge immediately to its left already renders
  "Not monitored".
- `Switch` gains an optional `disabled` prop (default `false`, so the four existing call sites are
  unaffected), used here to grey the control out while the save is in flight — matching the
  media-type select further down the page and stopping a double click from sending two PUTs.

I did not add a manual "set to wanted" status setter; the issue says derived status should stay
derived.

Placement follows the issue's wording ("next to the status badge"), so the toggle sits in the
header metadata row. If you would rather have it in a separate actions row like the author and
series pages, that is an 8-line move with no logic change — happy to do it.

Closes #2417

## Checklist

- [x] Tests added or updated — 4 vitest cases in `BookDetailPage.test.tsx` (state reflected in the
      switch, unmonitor updates the badge, monitor from the unmonitored state, one click = one PUT)
- [ ] Wiki pages updated if user-facing behaviour changed — the in-repo mirror
      `docs/User-Guide-Wiki.md` is updated here; tell me if you want the GitHub wiki updated too

## Test plan

- [x] `cd web && npm ci && npm run lint && npm run typecheck && npm run build && npm test`
      — lint 0 errors (6 pre-existing warnings in untouched files), typecheck clean, build clean,
      793/793 tests pass
- [x] `go build ./cmd/bindery` as a smoke check — this diff contains no Go code, so the Go half of
      `make check` was not run
- [ ] No visual check yet — the environment used has no browser, so placement in the metadata row
      at narrow widths is the one thing worth eyeballing in review

Note: `src/pages/HistoryPage.test.tsx` ("fetches later pages", #1467) flaked once under a full
parallel run and passed both in isolation and on a clean re-run of the whole suite. Unrelated to
this change, but you may see it go red intermittently.

This change was developed with AI assistance (Claude Code); the checks listed above were run on
this exact branch.

### DIFF
--- a/changelog.d/2417-book-page-monitor-toggle.md
+++ b/changelog.d/2417-book-page-monitor-toggle.md
@@ -1,0 +1,2 @@
+### Added
+- **Monitor toggle on the book page** (#2417) — a book's monitored state can now be flipped from its own page, next to the status badge. Previously the page showed the state but you had to go back to the Books list and use the bulk Monitor / Unmonitor action to change it.
--- a/docs/User-Guide-Wiki.md
+++ b/docs/User-Guide-Wiki.md
@@ -76,7 +76,9 @@ Separately, every book is **monitored** or not. Monitored means "Bindery
 actively pursues this." The **Wanted page** — and the automatic search sweep —
 only covers books that are *all three of*: status `wanted`, monitored, and not
 excluded. A book that is `wanted` but unmonitored shows as "Not monitored" and
-is left alone.
+is left alone. Flip a single book from the switch next to the status badge on
+its own page, or many at once from the Books list (select, then **Monitor** /
+**Unmonitor**).
 
 Two related labels:
 
--- a/web/src/components/Switch.tsx
+++ b/web/src/components/Switch.tsx
@@ -8,6 +8,8 @@ interface SwitchProps {
   /** Optional visible text rendered next to the switch. */
   children?: ReactNode
   title?: string
+  /** Blocks clicks while an in-flight save is settling. */
+  disabled?: boolean
   className?: string
 }
 
@@ -16,7 +18,7 @@ interface SwitchProps {
 // sliding affordance is the whole point: a badge and a toggle must not look
 // alike, or users can't tell which pills they can click. Track/knob sizing
 // matches the original inline switch on the Series page.
-export default function Switch({ checked, onChange, label, children, title, className = '' }: SwitchProps) {
+export default function Switch({ checked, onChange, label, children, title, disabled = false, className = '' }: SwitchProps) {
   return (
     <button
       type="button"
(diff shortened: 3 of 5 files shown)


vavallee:
Looks good, thanks. Switch is the right component here since the author and series pages already use it for monitoring, and the disabled prop change is clean.

It had drifted behind main on the buttons import in BookDetailPage.tsx (main added dangerLink), so I pushed a merge to your branch keeping both lines rather than bounce it back for a one liner. Typecheck, the page tests and the build all pass on the result.

One optional follow up: consider passing visible text as children like the author page does (Monitored / Unmonitored). The badge only says Not monitored for wanted books, so on an imported book the switch is an unlabeled knob unless you hover.

PR #2474, opened 2026-09-06
feat(metadata): add BINDERY_DISABLE_OPENLIBRARY switch

trevorswanson:
## Summary
- add `BINDERY_DISABLE_OPENLIBRARY` (default `false`)
- omit OpenLibrary from metadata enrichment when disabled
- avoid selecting OpenLibrary as primary when disabled, preferring Hardcover with a configured token
- add provider-selection and environment-loading regression tests

## Motivation
Hardcover-primary installations still receive OpenLibrary enrichment requests. When OpenLibrary is slow or unavailable, author catalogue refreshes can stall or fail to commit. Operators who want Hardcover-only metadata need an explicit deployment switch.

## Testing
- `docker run --rm -v /home/trevor/work/bindery:/src -w /src golang:1.26.6-bookworm sh -c 'gofmt -w internal/config/config.go internal/config/config_test.go cmd/bindery/main.go cmd/bindery/metadata_providers.go cmd/bindery/metadata_providers_test.go && go test ./internal/config ./cmd/bindery'`\n- Result: both packages pass\n\nRelated: #1888

### DIFF
--- a/cmd/bindery/main.go
+++ b/cmd/bindery/main.go
@@ -303,7 +303,8 @@ func main() {
 	if s, _ := settingsRepo.Get(ctxBoot, api.SettingMetadataPrimaryProvider); s != nil {
 		configuredPrimary = s.Value
 	}
-	primaryName := resolveMetadataPrimaryProvider(configuredPrimary, api.GetHardcoverAPIToken(ctxBoot, settingsRepo) != "")
+	hardcoverTokenConfigured := api.GetHardcoverAPIToken(ctxBoot, settingsRepo) != ""
+	primaryName := resolveMetadataPrimaryProvider(configuredPrimary, hardcoverTokenConfigured, !cfg.DisableOpenLibrary)
 	var primaryProvider metadata.Provider
 	switch primaryName {
 	case "dnb":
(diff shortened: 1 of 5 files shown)


vavallee:
Thanks for putting this together, but I am going to pass on it, and I want to explain why rather than just close.

The motivation points at #1888, and that one turned out not to be OpenLibrary at all. It was a full library walk per added book, fixed in #1930 and confirmed by the reporter. So the hour long refreshes this is meant to work around should already be gone on a current build. If you are still seeing stalls with OpenLibrary slow, please open an issue with the log lines and we will look at that directly.

On the shape: Bindery is moving everything behaviour related into settings with a descriptor, admin gating and a place in the UI (see #2445). This adds an env var the settings screen cannot see, so the Metadata tab keeps saying OpenLibrary is the primary while the process quietly runs Hardcover or DNB. Two things in particular: with the var set and no Hardcover token you get DNB as primary with no warning logged, which would surprise any English language install, and the recommender still calls OpenLibrary through WithOLClient in main.go, so the switch does not really switch it off.

If a Hardcover only, no enrichers mode is something you want, I would take that as a settings key with a registry entry and a toggle in the Metadata tab, plus a doc line in DEPLOYMENT.md. Happy to talk through the design in an issue first so you do not build the wrong thing.

PR #2486, opened 2026-09-06
feat(book): allow deleting an individual tracked file

### LINKED ISSUE #2485: Add an action to delete an individual book file
### Pre-flight

### Problem

The book detail page currently allows deleting files by format or deleting all files, but it does not provide a way to remove one specific file from disk.

This makes it difficult to clean up duplicate files, stale copies, or other unwanted files when multiple files exist for the same book. Deleting by format can remove more files than intended.

The existing **Forget this file** action only removes the file from Bindery's database tracking and does not delete it from disk. That behavior is useful and should remain separate.

### Proposed solution

Add a destructive **Delete this file** action to each tracked file's **More** menu, positioned below **Forget this file**.

When selected, the action should:

- Show the existing file-deletion confirmation dialog.
- List only the selected file path.
- Permanently delete that exact tracked file from disk.
- Remove only that file's `book_files` tracking row.
- Leave all other files on the book untouched.
- Refresh the book's file list and status.
- Record the deletion in history.

The action should be disabled for untracked files, and existing root-containment and ownership safety checks should remain enforced.

Audiobook directory-backed files should continue to be treated as one atomic audiobook bundle, since the tracked audiobook entry represents the directory-backed download.

Deleting one of multiple files for the same monitored format should not change the book's status while another file for that format remains. An imported book should only return to **wanted** when the last file for a monitored format is removed.

Existing **Forget this file**, format-level deletion, and delete-all behavior should remain unchanged.

### Alternatives considered

- **Use the existing format-level delete action:** This can remove sibling files that the user intended to keep.
- **Use Forget this file:** This leaves the unwanted file on disk and only removes its database tracking.
- **Manually delete the file outside Bindery:** This leaves stale tracking unless the user separately forgets the file.

A targeted delete action provides the expected cleanup behavior while preserving the distinction between deleting a file and forgetting its database record.

trevorswanson:
## Summary

Closes #2485.

Add a destructive **Delete this file** action to each tracked file's More menu on the book detail page. The action is placed below **Forget this file**, deletes only the selected tracked file, removes its `book_files` row, refreshes the book state, and records a history event.

The implementation preserves configured-root and cross-book ownership safety checks, keeps audiobook directory-backed entries atomic, and leaves format-level deletion and DB-only forgetting unchanged. Status remains `imported` when another file for the monitored format survives; an imported book returns to `wanted` only when the last required file for a monitored format is removed.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed (not applicable)
- [ ] Added a changelog fragment under `changelog.d/` (not added; maintainer/release process)
- [ ] Wiki pages updated if user-facing behaviour changed (not needed for this focused UI action)

## Test plan

- [x] `go test -count=1 ./internal/api ./internal/importer ./internal/db`
- [x] `npm test -- --run src/pages/BookDetailPage.test.tsx` (62 tests passed)
- [x] `npm run lint` (0 errors; existing warnings only)
- [x] `git diff --check`

### DIFF
--- a/internal/api/books.go
+++ b/internal/api/books.go
@@ -638,6 +638,10 @@ func (h *BookHandler) DeleteFile(w http.ResponseWriter, r *http.Request) {
 	id := existing.ID
 
 	if path := r.URL.Query().Get("path"); path != "" {
+		if r.URL.Query().Get("delete") == "true" {
+			h.deleteTrackedBookFile(w, r, id, path)
+			return
+		}
 		h.deregisterBookFile(w, r, id, path)
 		return
 	}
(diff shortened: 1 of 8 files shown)


vavallee (changes requested):
Tested locally against main and the fail before check holds, ownership and containment are all routed through the same guards as the scoped delete, good.

One thing to fix before merge: deleteTrackedBookFile throws away the skipped result from safeRemoveBookPathExact. When the root check or the other book ownership guard refuses the unlink, we still drop the row, answer 200 and write a history event saying the path was deleted. For a destructive action the response and the history need to say what actually happened. The format scoped delete only records paths it actually removed, so either do the same (put skipped into the event data and the response) or refuse the request when the file was not removed.

Also noting the exact delete never prunes an empty parent folder the way the scoped sweep does. Fine either way, just mention it in the PR if you leave it.

trevorswanson:
Addressed both pieces of feedback -- I wanted to fix the folder pruning than creating a regression / different behavior from other features.

1. **`deleteTrackedBookFile` now honors the skipped result** (`internal/api/books.go`) — previously it discarded `safeRemoveBookPathExact`'s `skipped` bool, so a containment/ownership refusal still dropped the `book_files` row, answered 200, and logged a "deleted" history event for a file that was never touched. Now a skipped unlink refuses the whole request with `409 Conflict`, leaving the row and the history untouched — no lie in the response.
2. **`safeRemoveBookPathExact` now prunes the empty parent folder** (`internal/api/path_safety.go`) after removing the target file, mirroring what `removeBookPathScoped` already does for the format-scoped sweep.

Added tests: `TestBookDeleteFile_IndividualPathRefusedWhenNotRemoved` (409, row/file survive, no history event) and `TestSafeRemoveBookPathExact_PrunesEmptyParent` / `_KeepsNonEmptyParent`.

Based on a comment in a different PR about DCO (https://github.com/vavallee/bindery/pull/2480#pullrequestreview-5188952845) I added sign-off from myself, with co-authored by Claude -- that's now triggering CI failure [here](https://github.com/vavallee/bindery/actions/runs/35528906114/job/106125873907?pr=2486#step:4:105). Let me know how you'd like me to address that.

Issue #2490, opened 2026-09-07
No way to deregister a book_files row whose file is gone


vavallee:
Split out of #1635, where it kept that issue open after its actual subject was fixed. This is general `book_files` hygiene, not Calibre-specific.

## What happens

`derivedFormatPath` (`internal/db/books.go:897`) skips `book_files` rows whose path no longer resolves on disk and renders the first live one, then logs a Warn per stale row. That is the right render behaviour and it is why a moved file no longer breaks the UI.

But the stale row is never removed, and `books.go:933` says so in as many words:

```go
// Warn, not Debug: this is a row the user has to deregister by hand,
// and the in-app log view is where they will find out it exists.
```

There is no by-hand way to do it. `RemoveBookFile` exists in the repo layer but nothing user-facing calls it for this purpose, `library/scan` re-derives rather than prunes, and `manual-import/match` needs a `downloadId`.

## Why it accumulates

Every path that moves a tracked file without repointing its row leaves one behind:

- a Calibre library reorganisation, then a re-import, which registers the new path beside the old (`internal/calibre/importer.go:446`)
- a file moved outside Bindery on any library
- a renamer run that failed partway

Each one costs a row and a Warn line per status recompute, forever. On the deployment that surfaced #1635 the library is on an SMB mount reorganised by another tool, so this is not hypothetical there.

## What would fix it

A per-row deregister on the book detail page, listing the tracked files with the unresolvable ones marked, is the smallest thing. `RemoveBookFile(ctx, path)` already does the work and refreshes the owning book's status.

A sweep is the alternative but is riskier: a path that fails to stat because a network mount is down is not a path that is gone, and `BookFilePathResolves` is deliberately asymmetric about exactly that (`books.go:868`). Anything automatic has to inherit that asymmetry, which is a good argument for making the first version a button a person presses.

vavallee:
Closing this. I was wrong, the thing it asks for already exists.

I wrote here that "RemoveBookFile exists in the repo layer but nothing user-facing calls it for this purpose". That is not true. `deregisterBookFile` in `internal/api/books.go:740` calls it, and it landed back in #1692. It is wired to a per file "Forget this file" action in the More menu on the book detail page, `BookDetailPage.tsx:884`, with a confirm dialog that says "Bindery stops tracking this path. Nothing is deleted from disk. Use this when the file has already been moved or removed elsewhere."

Which is exactly what I described as "what would fix it", down to it being a button a person presses rather than an automatic sweep.

[...]

PR #2491, opened 2026-09-07
fix(calibre): untrack the file rows a rolled back import created (#1635)

### LINKED ISSUE #1635: calibre/import: series_link violates the provenance entity_type CHECK (links escape rollback), and book_files paths are never written
### Pre-flight

*Duplicate search:* #905 covered series *not being imported* from Calibre and the scan log aggregating roots; neither of the two defects below — the provenance CHECK violation and the untouched `book_files` paths — is covered there or elsewhere.

### Current behaviour

Two separate defects in `POST /api/v1/calibre/import`, both reproducible on every run against my library. They're independent fixes but live in the same code path, so I've kept them together — say the word and I'll split.

**1. Every series link logs a CHECK-constraint failure.** The provenance table only permits three entity types (`internal/db/migrations/044_calibre_import_rollback.sql:24`):

```sql
entity_type TEXT NOT NULL CHECK(entity_type IN ('author', 'book', 'edition')),
```

but the importer upserts a fourth (`internal/calibre/importer.go:418`, with `entityTypeSeriesLink = "series_link"` from `import_types.go:14`):

```go
i.upsertProvenance(ctx, runID, entityTypeSeriesLink, …)
```

so each series link emits `CHECK constraint failed: entity_type IN ('author','book','edition')` for `entityType: series_link`. The series links themselves are created correctly and the error is non-fatal, but every provenance row for a series link is silently dropped — which means **those links aren't covered by the rollback machinery that migration 044 exists to provide**. A run rolled back afterwards will leave its series links behind.

Fix is presumably just adding `'series_link'` to the CHECK in a new migration.

**2. `calibre/import` never writes `book_files` paths.** The import reconciles title/author/series from `metadata.db` but leaves the tracked file path untouched. On my install it reported `booksUpdated: 17` while a book whose ebook had moved kept a `book_files.path` pointing at its old, now-nonexistent directory. The derived top-level `filePath` resolves correctly from Calibre, so the UI looks right and the discrepancy is invisible until something consumes the tracked path.

That leaves no supported way to re-point a stale `book_files` row for a Calibre-managed library: `POST /api/v1/library/scan` is the filesystem folder-parse (which, for a Calibre-shaped library, fights the DB import and spawns duplicate author/book records — and left the file `unmatched` in my case anyway), `manual-import/match` requires a `downloadId` that a Calibre-sourced book doesn't have, and the container is distroless with the DB at 0600, so there's no surgical option either.

### Expected behaviour

1. `series_link` accepted by the provenance CHECK (or the importer stops writing that entity type), so series links are rollback-covered like every other imported entity.
2. `calibre/import` writes/updates the `book_files` row for each book it reconciles, from the path `metadata.db` reports — so the tracked path and the derived path can't drift apart.

### Steps to reproduce

1. Point Settings → Calibre at an existing Calibre library (mine: 97 books, 12 authors, several with series).
2. `POST /api/v1/calibre/import`.
3. Watch the log: one CHECK-constraint error per series link.
4. Move a book's file within the Calibre library (or import a book whose `book_files.path` predates a library reorganisation), then re-run the import: `booksUpdated` counts the book, but its `book_files.path` still points at the old location.

### Trace logs

```text
calibre import: persist provenance … CHECK constraint failed: entity_type IN ('author','book','edition')  entityType: series_link
```

```text
POST /api/v1/calibre/import → {"booksUpdated":17, …}
# book_files.path for the affected book unchanged, still referencing the previous directory
# derived filePath (read from Calibre) resolves to the correct current file
```

### Deployment method

Docker (ghcr.io)

### Bindery version

v1.27.0 (`sha-2c705ea`)

### Environment details

- `ghcr.io/vavallee/bindery:latest`, distroless image run as `user: "1001:1001"`
- Calibre library managed by Calibre-Web-Automated on a Synology NAS; Bindery reads it read-only over an SMB mount and never writes `metadata.db`
- Settings → Calibre: library path set, write integration off, library import on

vavallee:
Closes #1635.

## The defect

`registerBookFiles` (`internal/calibre/importer.go`) inserts a `book_files` row for every format Calibre reports for a book. It recorded no provenance for those rows, so `ExecuteRollback` had nothing to undo them with.

For a book the run created, this is invisible: rolling back deletes the book and `ON DELETE CASCADE` takes the file rows with it. For a book that already existed in the library, the run's file rows survive the rollback, and with them the book's `imported` status and its effective file path. The library then claims to hold a copy the user never asked Bindery to track.

## The fix

The importer records a provenance row per book file it inserts, keyed by the path, and a create snapshot only when the `INSERT OR IGNORE` actually created the row.

Rollback gains a `book_file` case at rank 0, so paths are untracked before the book rows are considered. It untracks only rows that this run created and still owns, checked the same way the existing cases check it. A path a download placed, or one an earlier import already registered, is skipped with a stated reason. Nothing on disk is modified, and the existing files-left-on-disk warning still fires.

## Supporting changes

- `BookFileRepo.AddIfMissing` reports whether the insert created a row. That boolean is what separates "this run put it there" from "it was already tracked".
- `BookRepo.UntrackFilePath` deletes by path and refreshes the owning book's status through `r.exec`, so it runs inside the rollback transaction. `derivedFormatPath` moved from `r.db` to `r.exec` for the same reason: `WithTx` clones only `exec`, and `MaxOpenConns` is 1, so a bare pool read inside a transaction deadlocks. The first cut of this branch hung a test past 154s on exactly that.
- Migration `085` rebuilds `calibre_provenance` and `calibre_entity_snapshots` to admit `book_file` in the `entity_type` CHECK, mirroring 072's shape. The Down deletes `book_file` rows before restoring the old constraint.
- `FilesOnDiskWarning` counts each book once, via `filesCounted`. The untrack case clears a book's path fields before `delete_book` sees it, which otherwise silently loses the warning. An existing test caught this.

## Fail before

Both proofs recorded against the pre-fix tree:

```
no rollback case: preview has no untrack_file action ... Action:skip Reason:unsupported entity type
no provenance:    no calibre_provenance row for the book file (CHECK constraint rejected it?)
```

## Tests

New: `internal/calibre/import_book_file_rollback_test.go` (provenance recorded, untrack on a pre-existing book, a file the run did not insert is left alone, the on-disk warning survives the untrack) and `internal/db/migration_085_test.go` (up and down against a table restored to its pre-085 shape).

Green: `internal/calibre`, `internal/db`, `internal/api`, `internal/importer`. `gofmt`, `go vet` and `go build` clean.

## Not in this PR

The other two items raised on #1635. Deregistering a stale `book_files` row by hand is filed as #2490. The quality-profile ordering request was declined in the thread.

### DIFF
--- a/changelog.d/1635-calibre-book-file-rollback.md
+++ b/changelog.d/1635-calibre-book-file-rollback.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Rolling back a Calibre import no longer leaves its file rows behind** (#1635). The library import registers a tracked file for every format Calibre reports, but those rows were not covered by the run's rollback, so undoing a run against books that already existed left the files it had claimed still attached. Rollback now untracks them, and only the rows that run actually created, so a path a download placed is never touched. The files on disk are never modified.
--- a/internal/calibre/import_book_file_rollback_test.go
+++ b/internal/calibre/import_book_file_rollback_test.go
@@ -221,3 +221,70 @@ func TestRollback_UntrackWarnsAboutFilesLeftOnDisk(t *testing.T) {
 		t.Errorf("FilesAffected = %d, want 1 (the book counted once, not per action)", result.Stats.FilesAffected)
 	}
 }
+
+// The safety property this whole change turns on. Run one inserts the row, run
+// two re-reports the same path and takes over its provenance. Rolling back run
+// one must then leave the row alone: the claim on that path now belongs to a
+// run the user is not undoing. Without the ownedByRun guard the older run would
+// happily untrack a file the newer one still depends on.
+func TestRollback_SkipsAFileALaterRunNowOwns(t *testing.T) {
+	imp, fr, _, bookRepo, runsRepo, _, _ := newSeriesRollbackFixture(t)
+	ctx := context.Background()
+
+	fr.books = []CalibreBook{sampleCalibreBook(1, "Book One", "Alice Author")}
+	if _, err := imp.Run(ctx, "/lib"); err != nil {
+		t.Fatalf("first run: %v", err)
+	}
+	book, err := bookRepo.GetByCalibreID(ctx, 1)
+	if err != nil || book == nil {
+		t.Fatalf("book not found: %v", err)
+	}
+	path := filepath.Join("/lib", "Book One.epub")
+
+	if _, err := imp.Run(ctx, "/lib"); err != nil {
+		t.Fatalf("second run: %v", err)
+	}
+
+	runs, err := runsRepo.ListRecent(ctx, 2)
+	if err != nil {
+		t.Fatalf("ListRecent: %v", err)
+	}
+	if len(runs) != 2 {
+		t.Fatalf("expected 2 runs, got %d", len(runs))
+	}
+	firstRunID := runs[len(runs)-1].ID
+
+	preview, err := imp.PreviewRollback(ctx, firstRunID)
+	if err != nil {
+		t.Fatalf("PreviewRollback: %v", err)
+	}
+	if hasAction(preview.Actions, entityTypeBookFile, "untrack_file") {
+		t.Fatalf("the older run planned to untrack a file the newer run owns: %+v", preview.Actions)
+	}
+	var reason string
+	for _, a := range preview.Actions {
+		if a.EntityType == entityTypeBookFile {
+			reason = a.Reason
+		}
+	}
+	if reason != "run is no longer the current provenance owner for this book file" {
+		t.Errorf("want the ownership skip reason, got %q", reason)
+	}
+
+	if _, err := imp.Rollback(ctx, firstRunID); err != nil {
+		t.Fatalf("Rollback: %v", err)
+	}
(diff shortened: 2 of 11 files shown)


vavallee:
On the ownership point, that reasoning is about the wrong run. It is right that a run which re-reports an existing path gets no snapshot, so rolling back *that* run never enters the case. But the interesting rollback is the *earlier* one: run 1 did create the row and does have a create snapshot, and `upsertProvenance` has since moved the claim to run 2. `ownedByRun` is the only gate there.

Pushed a test for it in 4ff70210. Deleting the guard fails it, and the preview shows why:

```
{EntityType:book_file  Action:untrack_file Reason:}
{EntityType:edition    Action:skip Reason:run is no longer the current provenance owner for this edition}
{EntityType:book       Action:skip Reason:run is no longer the current provenance owner for this book}
{EntityType:author     Action:skip Reason:run is no longer the current provenance owner for this author}
```

Without it `book_file` is the only entity out of step with the other three.

Agreed on the rest, including the `refreshBookStatus` nit. Leaving that as is since it matches `AddBookFile`.

PR #2497, opened 2026-09-07
fix(hardcover): keep one book per position in a series (#2496)

### LINKED ISSUE #2496: Hardcover series catalog counts every translation and box set as its own volume, so a 7-book series arrives as 34 entries
### Pre-flight

Closest existing issues are #2379 (merged) and #2410, but both work one layer below this one: they fix how local books are *assigned* to catalog entries and assume the catalog itself is right. This is about what the catalog contains before any matching happens.

### Current behaviour

`GetSeriesCatalog` returns every book Hardcover files under a series, and Hardcover files far more than the volumes. Two habits of that data source collide with the assumption that one row is one volume:

1. **A translation is its own book record, and it keeps the original's position.** Position 1 of Harry Potter is not one book, it is the English novel plus the Dutch, French, Norwegian, Polish, Portuguese, Russian and Spanish books, each a separate `books` row with `position: 1`.
2. **An omnibus is filed at the position of the first book it contains.** `Harry Potter Collection #1-6`, `Harry Potter Boxed Set` and `The Harry Potter Trilogy` therefore also sit at position 1.

`dedupeSeriesCatalogBooks` cannot see any of this, because it matches on title and every one of these titles genuinely differs.

Measured against series 1185 (Harry Potter), which Hardcover itself reports as `primary_books_count: 7`:

```
130  rows returned by book_series
 34  entries left after dedupeSeriesCatalogBooks, i.e. what the series page shows
  7  actual volumes
```

The series page is then unusable in the way that matters: `Fill gaps` queues 27 books that do not exist as separate works, and every one of them has to be pruned by hand afterwards.

This is not specific to Harry Potter. The Way of Kings is nine records at position 1 of The Stormlight Archive — the English novel, Russian, Bulgarian split into two volumes, Polish, Portuguese, and two box sets.

### Expected behaviour

One catalog entry per position in the series, so the count matches the `primary_books_count` Hardcover already publishes.

### Steps to reproduce

1. Configure Hardcover as a metadata provider with an API token.
2. Create a series and link it to Hardcover `harry-potter` (series 1185), or use any series with published translations.
3. Open the series page.
4. The catalog lists 34 entries for a series of 7, with box sets and non-English editions among them.

### Trace logs

The clearest evidence is the API response rather than Bindery's log, since Bindery reports faithfully what it was given. Same query `GetBooksBySeries` issues, run against series 1185:

```
book_series rows, as queried today (canonical_id is null)      130
  ... adding  book: { compilation: { _eq: false } }            108
  ... then keeping one book per position                        10   (7 whole-numbered + 3 fractional)
  Hardcover's own primary_books_count                            7
```

Reader counts make the choice at a contested position unambiguous rather than marginal. At position 1:

```
Harry Potter and the Philosopher's Stone      users_count 17351
the other entries sharing position 1          users_count in the tens or zero
```

### Deployment method

Source build

### Bindery version

`v1.34.0-23-gf37672c4` (source build of current `main`)

### Environment details

Hardcover configured as the primary metadata provider, free-plan API token. Reproduced against the live Hardcover API, not a fixture.

### Notes

I have a fix and will open a PR against this issue: exclude compilations in the query, and where entries still share a position keep the one the most readers hold. Books with no position are left alone, since those are extras a series accumulated rather than duplicates of a volume.

totza2010:
## Summary

`GetSeriesCatalog` treated every row Hardcover files under a series as a volume of it. Hardcover files two other things there.

A **translation is its own book record that keeps the original's position**, so position 1 of Harry Potter arrives as eight books — the English novel plus the Dutch, French, Norwegian, Polish, Portuguese, Russian and Spanish ones. An **omnibus is filed at the position of the first book it contains**, so `Harry Potter Collection #1-6`, `Harry Potter Boxed Set` and `The Harry Potter Trilogy` sit at position 1 as well. Every one of those titles genuinely differs from the others, so `dedupeSeriesCatalogBooks` has nothing to match on and keeps them all: 130 rows in, 34 entries on the series page, for a series Hardcover itself reports as `primary_books_count: 7`. `Fill gaps` then queues 27 books that are not books, and they have to be pruned by hand.

Two changes, one for each habit.

**Compilations are excluded in the query rather than filtered afterwards**, because an omnibus at position 1 is indistinguishable from volume 1 once it has arrived — same shape, same position, a plausible title, and often more readers than a translation. `book.compilation` is the only thing that separates them and it is only available server-side.

**Where entries still share a position, the one the most readers hold wins.** That is `collapseSeriesPositions`. The margin is not a judgement call: the English *Philosopher's Stone* carries 17,351 readers where the other entries at position 1 are in the tens or at zero. The query's `order_by` gains `{book: {users_count: desc_nulls_last}}` so the result is stable rather than dependent on what Hasura returns first.

Three cases are deliberately left alone:

- **A book with no position** survives untouched. Those are extras a series accumulated — companions, artbooks — not competing editions of a volume, so several of them can and should coexist.
- **A fractional position is its own slot.** `Edgedancer` at 2.5 is not a rival edition of `Words of Radiance` at 2.
- **A book listed at two positions** still resolves to the lower one, as it did before.

The collapse runs *after* `dedupeSeriesCatalogBooks`, not before. Reversing them breaks `TestGetSeriesCatalog_ParsesAndDedupesBooks`: title dedup expects to still see the duplicates it exists to remove.

Closes #2496

### Tests

Four cases for the collapse and one for the query, in `internal/metadata/hardcover/series_positions_test.go`. The fixture for the first is the real shape and the real reader counts Hardcover returns for position 1 of series 1185.

Neutering the fix — `collapseSeriesPositions` returning its input, the `compilation` filter and the `users_count` ordering removed — fails them:

```
    series_positions_test.go:25: collapseSeriesPositions() kept 4 books, want one per position
--- FAIL: TestCollapseSeriesPositionsKeepsTheMostHeldEntry (0.00s)
--- PASS: TestCollapseSeriesPositionsLeavesUnpositionedBooksAlone (0.00s)
--- PASS: TestCollapseSeriesPositionsTreatsFractionalPositionsSeparately (0.00s)
    series_positions_test.go:79: collapseSeriesPositions() kept 4 books, want 3
--- FAIL: TestCollapseSeriesPositionsPreservesOrder (0.00s)
    series_positions_test.go:93: GetBooksBySeries no longer excludes compilations
    series_positions_test.go:96: GetBooksBySeries no longer orders competing entries by reader count
--- FAIL: TestSeriesCatalogQueryExcludesCompilations (0.00s)
FAIL	github.com/vavallee/bindery/internal/metadata/hardcover	0.760s
```

The two that pass under the neutered build pass for the right reason — they assert that books are *kept*, which is what the unfixed code does anyway. They are there to pin the exemptions against a future collapse that is too eager.

`Ran:` `go build ./...`, `go vet ./internal/metadata/hardcover/...`, `go test ./internal/metadata/hardcover/...` (ok), and the change was exercised against the live Hardcover API in a running instance — the Harry Potter series page goes from 34 entries to 10, and a Hercule Poirot catalog lines up with the entry count hardcover.app shows for that series.

I could not run `make check` in full here: this is a Windows host, where 19 tests in `internal/api`, `internal/db` and `internal/downloader` fail on hardlink and SQLite path behaviour before any of my changes. The same 19 fail on a clean `upstream/main` checkout, and none of them are mine.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed — no env or config change
- [ ] Wiki pages updated if user-facing behaviour changed — happy to, if you want the series page documented as showing volumes rather than records

## Test plan

- [x] `go test ./internal/metadata/hardcover/...`
- [x] Live series fetch against Hardcover for series 1185 (Harry Potter), and for Hercule Poirot as a series carrying many fractional positions


totza2010:
Pushed `c2d33787`, which closes a real gap rather than the Codecov number.

Looking at what the four original cases actually prove, they all call `collapseSeriesPositions` directly. Nothing asserted it was reached from `toSeriesCatalog`, which is the path the bug lives on — I removed the call and the whole package stayed green. That is the one mistake this PR could have shipped without anyone noticing.

`TestGetSeriesCatalogKeepsOneBookPerPosition` goes through `GetSeriesCatalog` with a position 1 holding the English novel and the Russian one, so title dedup cannot reach it either. Without the wiring:

```
catalog holds 3 books, want one per position:
hc:harry-potter-and-the-philosophers-stone, hc:garri-potter-i-filosofskii-kamen,
hc:harry-potter-and-the-chamber-of-secrets
```

On the Codecov report itself: patch coverage was 95% with one partial, and every block of the new function shows covered in a local `-coverprofile` run, so I think the partial is a branch-level artefact rather than an untested path. Happy to chase it if you would rather it were clean.

vavallee (approved):
Read the diff and ran it. This is a really well put together PR, and the writeup did most of my job for me. Thanks for the live API numbers especially, 34 down to 10 on Harry Potter and Poirot lining up with what hardcover.app shows is the kind of evidence that is hard to argue with.

Checked a few things myself:

`collapseSeriesPositions` keeps insertion order by replacing in place rather than appending, so the `order_by` position asc survives the collapse. Fractional positions stay separate because the key is the position string, so 2.5 never collides with 2. Both good.

I tried to break the wiring rather than the function, swapping the call site back to bare `dedupeSeriesCatalogBooks`, expecting the unit tests to pass anyway since they call `collapseSeriesPositions` directly. The package fails, so the path through `toSeriesCatalog` is pinned too. Nice.

One question, and it is a question rather than a blocker.

`compilation: {_eq: false}` in Hasura will not match a row where `compilation` is null, it only matches an explicit false. If any Hardcover book record has that column unset, those books would drop out of every series catalog silently. The codebase already spells this out elsewhere when a column can be null, `canonical_id: {_is_null: true}` right above it and `authorContributionFilter` in `client.go`, so the idiom is there if it turns out to be needed:

```
_or: [{compilation: {_eq: false}}, {compilation: {_is_null: true}}]
```

Your live check probably rules it out already, since a series losing books to nulls would show up as a count that came in under what hardcover.app reports rather than matching it. If you happen to know the column is non-null then ignore this entirely. If you are not sure, the `_or` costs nothing and removes the question.

Approving either way, since that is a small follow up if it is anything at all.

totza2010:
> Reference in n

Checked rather than guessed. books(where: {compilation: {_is_null: true}}, limit: 5) returns [] across the whole table, and on series 1185 the partition is complete: 35 rows after canonical_id: {_is_null: true}, 27 with compilation: {_eq: false}, 8 with {_eq: true}, 0 with {_is_null: true}. 27 + 8 = 35, so nothing is falling through today — which is also why the count matched hardcover.app rather than coming in under it.

That says the column is unset nowhere right now, not that it cannot be. A row added later without a backfill would vanish exactly as you describe, silently. Sending the _or as a follow-up — it costs nothing and it matches the idiom already on the line above it.

Issue #2324, opened 2026-09-04
Series view still shows excluded books as missing: neither series query selects books.excluded


vavallee:
Follow-up to #2302. That issue had two halves: the action paths (series fill and genre apply acting on excluded books) and the display paths (the series view showing them with no indication). #2310 fixed the first half only, in `ListBooksInSeries`. The second half is still live on `main` at `c6895aed`.

## What is wrong

`SeriesRepo.ListWithBooksForUser` (`internal/db/series.go:76`) and `SeriesRepo.GetByIDForUser` (`internal/db/series.go:320`) are the two queries behind `GET /api/v1/series` (`internal/api/series.go:172`) and `GET /api/v1/series/{id}` (`internal/api/series.go:190`). Neither selects `books.excluded`, and neither filters on it.

The scan targets are fixed lists, so `models.Book.Excluded` (`internal/models/book.go:72`) is left at its zero value and the API serialises `"excluded": false` for every book on every series, including books the user has excluded.

## What the user sees

`web/src/pages/SeriesPage.tsx` has no way to tell the difference, so an excluded book:

1. **Counts as a gap.** `gapCount` at `:271` is `books.filter(b => b.book && b.book.status !== 'imported').length`. An excluded book still sits at `wanted`, so it feeds the amber `{displayMissingCount} missing` pill at `:298`.
2. **Renders as a plain wanted book** in the expanded row at `:417`, with the amber `wanted` status badge and no excluded marker. `AuthorDetailPage.tsx` shows one at `:532`, `:554` and `:647`; the series page cannot, because it never receives the flag.
3. **Offers a Fill that cannot do anything.** `fillNeeded` at `:276` is `gapCount > 0`, so the Fill button is enabled. Post-#2310 `ListBooksInSeries` correctly skips the excluded book, so `fillSeriesAll` returns `queued: 0` and the UI prints `Nothing to fill` (`:177`) directly beside a badge that still says `1 missing`.

So a series whose only outstanding book is one you deliberately excluded reads as incomplete forever, and the one action offered to fix it reports that there is nothing to do.

## Repro

1. Add an author with a series, leave one book at `wanted`.
2. Exclude that book from the author page. The author page drops it, correctly.
3. Open Series. The series shows `1 missing`.
4. Expand it. The excluded book is listed with an amber `wanted` badge, no excluded marker.
5. Press Fill. `Nothing to fill`, while the badge still says `1 missing`.

## Suggested fix

[...]

PR #2503, opened 2026-09-08
fix(series): show excluded books correctly on the series view (#2324)

thejdubb02:
## Summary

Follow-up to #2302 / #2310. Those fixed the action path: after #2310, series fill and genre apply correctly skip excluded books. The display path was still wrong, because the two queries behind `GET /series` and `GET /series/{id}` never selected `books.excluded`.

`SeriesRepo.listWithBooksForUser` and `SeriesRepo.GetByIDForUser` scan into fixed column lists, so `models.Book.Excluded` stayed at its zero value and the API serialised `"excluded": false` for every book on every series. The Series page could not tell an excluded book from a wanted one, so an excluded book:

- counted toward the amber "N missing" pill,
- rendered as a plain wanted book with no excluded marker,
- and enabled a Fill that returned `queued: 0` ("Nothing to fill") right beside a badge that still said it was missing.

Changes:

- Backend: add `b.excluded` to both series SELECTs and scan it into `models.Book.Excluded`. `GetByIDForUser` joins books directly, so it scans a plain int like `monitored` does; `listWithBooksForUser` uses a LEFT JOIN, so it goes through the `sql.NullInt64` path the other nullable book columns already use. `ListByAuthor` and the exact-match `EXISTS_IN` query are intentionally left alone (out of scope; `EXISTS_IN` stays an exact match).
- Frontend (`SeriesPage.tsx`): drop excluded books from `gapCount` so the missing pill and the Fill button agree with the post-#2310 backend, and show a grey "excluded" badge on the row instead of the amber "wanted" one, reusing the badge styling already on that page.

## Checklist

- [x] Tests added or updated
- [ ] `docs/DEPLOYMENT.md` updated if env vars, config, or upgrade path changed (no config change)
- [ ] Wiki pages updated if user-facing behaviour changed (a bugfix restoring expected behaviour; no new user-facing concept)

## Test plan

- Backend: `go test -race -count=1 ./internal/db/` passes. Added `TestSeriesDisplayQueriesReturnExcludedFlag`, which links a kept book and an excluded book to a series and asserts both `ListWithBooksForUser` and `GetByIDForUser` return `Excluded` correctly. It fails on `main` before the change and passes after.
- Frontend: `cd web && npm run build` passes and `npm test` is green (all existing tests plus a new `SeriesPage.test.tsx` case asserting an excluded book is not counted as missing and is marked excluded; it fails before the change).
- `go vet`, `gofmt`, `npm run typecheck`, and `eslint` are clean on the touched files.

### DIFF
--- a/changelog.d/2324-series-excluded-flag.md
+++ b/changelog.d/2324-series-excluded-flag.md
@@ -1,0 +1,2 @@
+### Fixed
+- **Series view no longer shows excluded books as missing** (#2324). A book you exclude on the author page now shows as excluded on the Series view instead of being counted as a missing book.
--- a/internal/db/series.go
+++ b/internal/db/series.go
@@ -158,7 +158,7 @@ func (r *SeriesRepo) listWithBooksForUser(ctx context.Context, userID int64, ser
 		SELECT s.id, s.foreign_id, s.title, s.description, s.monitored, s.genre_override, s.created_at,
 		       sb.series_id, sb.book_id, sb.position_in_series, sb.primary_series,
 		       b.id, b.foreign_id, b.author_id, b.title, b.sort_title, b.status,
-		       b.monitored, b.image_url, b.release_date, b.created_at, b.updated_at
+		       b.monitored, b.image_url, b.release_date, b.created_at, b.updated_at, b.excluded
 		FROM series s
 		LEFT JOIN series_books sb ON sb.series_id = s.id
 		`+bookJoin+where+`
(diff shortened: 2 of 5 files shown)


vavallee (approved):
Looks good, merging. Pulled main in locally (clean, none of the later series work overlaps), ran the db and api suites plus typecheck, lint, vitest and the web build, all green. Also reverted the two non test files and confirmed both new tests fail without the fix, which is exactly what I want to see.

One tiny thing for a follow up if you feel like it: the Excluded badge is the same amber as the wanted pill next to it, so the two read as one. A grey badge like the author page uses would make the difference pop. Not worth holding this up over.

Thanks for the thorough writeup and for leaving ListByAuthor alone, that was the right call.

PR #2506, opened 2026-09-08
chore(release): correct the v1.35.0 entry and pull request counts

vavallee:
The v1.35.0 preamble opens with a count of what is in the section. It was written before #2489 and #2491 were folded in, so it reads "Forty one entries from twenty one merged pull requests" against an actual 43 entries.

Counted from the section on `main` at 80ee1d28: 1 Added, 11 Changed, 29 Fixed, 1 Removed, 1 Security.

The pull request figure goes 21 to 23 on the same arithmetic. #2488 also merged into this range but shipped no changelog fragment, so it adds no entry and is not counted.

Text only, no code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ All notable changes to Bindery are documented here. Format loosely follows
 
 **Search and author identity, largely rebuilt. Read the two warnings below before upgrading.**
 
-Forty one entries from twenty one merged pull requests, and most of them are one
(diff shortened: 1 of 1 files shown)


vavallee:
Superseded. The four fixes going into this release each add a changelog fragment, so the v1.35.0 section needs assembling again anyway, and the counts are better set once at the end than corrected twice. Folding this into that PR.

Issue #2525, opened 2026-09-10
A book in two series renames under whichever one SQLite reaches first, and nothing lets you choose


vavallee:
## Pre-flight

- [x] I searched existing issues and this isn't a duplicate (#2009 is the SkipSecondarySeries catalog filter, a different axis; #2328 is the inverse, series never linked at all)
- [x] Running v1.35.0

## Current behaviour

When a book belongs to more than one series, the renamer picks one of them at random and there is no way to influence the choice.

`series_books` (`internal/db/migrations/001_initial.sql:58-64`) is many-to-many with `primary_series INTEGER NOT NULL DEFAULT 1` and `PRIMARY KEY (series_id, book_id)`. Nothing enforces at most one primary row per book, and the column defaults to 1, so a book routinely ends up with several.

The renamer resolves `{Series}` and `{SeriesNumber}` through `Scanner.primarySeriesFor` (`internal/importer/scanner.go:219-229`), which calls `SeriesRepo.GetPrimarySeriesForBook` (`internal/db/series.go:915-926`):

```sql
SELECT s.title, sb.position_in_series
FROM series_books sb JOIN series s ON s.id = sb.series_id
WHERE sb.book_id = ? AND sb.primary_series = 1
LIMIT 1
```

`LIMIT 1` with no `ORDER BY`. With two matching rows SQLite returns whichever row the planner reaches first. In practice that follows `idx_series_books_book` (migration `048_list_endpoint_indexes.sql:36`), so it looks stable until an index, schema or ANALYZE change flips it.

Reported on Discord by Kerope: three books sit in both "King's Dark Tidings" and "King's Dark Tidings Universe", and those three rename under the Universe title while the rest of the series renames correctly.

Every naming and preview caller inherits the same pick: `scanner.go:1599` (audiobook import), `scanner.go:1985` (ebook import), `reorganize.go:114` and `:259` (library reorganize), `destination_preview.go:62` (the destination preview shown in the UI), `scanner_handoff.go:363`, and the Calibre push at `scanner_handoff.go:588-589` which writes `Metadata.Series` and `.SeriesIndex`.

### How the second primary row is created

Six link sites pass `primary = true` unconditionally, regardless of whether the book already has a primary series:

- `internal/api/series.go:1520`, `:1633`, `:1665` (series Fill and catalogue reconcile)
- `internal/abs/import_upserts.go:1165`
- `internal/abs/import_hardcover_series.go:435`
- `internal/calibre/importer.go:521-523`

[...]

PR #2528, opened 2026-09-10
fix(series): a secondary-only membership still names files (#2527)

### LINKED ISSUE #2527: A book whose only series membership is secondary renames with no series at all
## Pre-flight

- [x] Searched existing issues and this isn't a duplicate. #2525 is the sibling defect (a book with *two* primary rows renames under an arbitrary one); this is the opposite end, a book with *zero* primary rows renaming under none. #2009 is the SkipSecondarySeries catalog filter, unrelated.
- [x] Running v1.35.0

## Problem

`GetPrimarySeriesForBook` (`internal/db/series.go`) is the renamer's only source for `{Series}` and `{SeriesNumber}`, and it filters on `primary_series = 1`. A book whose every `series_books` row is secondary matches nothing and the function returns `("", "", nil)` — the same answer it gives for a book that is in no series at all.

`internal/importer/renamer.go:248-250` then drops any path segment that renders empty, so the file is written with no series folder and no series in its name. The book still shows its series everywhere in the UI, because `BookDetailPage` and the series pages list memberships without reference to the flag. The two disagree and nothing says so.

## How a book ends up with only secondary memberships

`internal/abs/import_hardcover_series.go:518-530` walks the Hardcover catalog for the series of the book being imported and links every *sibling* book it already has locally:

```go
linkCreated, err := i.series.LinkBookIfMissing(ctx, seriesRow.ID, localBook.ID, strings.TrimSpace(catalogBook.Position), false)
```

The hardcoded `false` is correct in intent — this run is importing a different book, so it should not claim the sibling's primary series. But when that sibling had no series membership at all, which is the normal state for a book imported from Audiobookshelf without series metadata, the `false` row is the only one it will ever have.

`POST /series/{id}/books` with `{"primarySeries": false}` reaches the same state directly.

Confirmed against the current tree with a throwaway probe: one membership, `GetPrimarySeriesForBook` returns `"" / ""`.

```
GetPrimarySeriesForBook = "" / ""
memberships = [1]
```

## Expected behaviour

A book that is in exactly one series should rename under that series, whether or not the row happens to be flagged primary. `primary_series` exists to break a tie between several memberships; with one membership there is no tie to break, and the flag should not be able to suppress the series entirely.

Two candidate shapes, either is fine:

1. Fall back in `GetPrimarySeriesForBook`: when no primary row matches, take the book's memberships under the same ordering #2525 introduced. This fixes every existing library on upgrade with no migration.
2. Promote at the link site: have `internal/abs/import_hardcover_series.go:529` use `LinkBookPreservingPrimary` like the other ABS sites, so a sibling with no primary gains one. This leaves already-imported libraries broken until something re-links them.

Option 1 is the one I would take, with option 2 as well for tidiness.

## Steps to reproduce

1. Have a book in the library with no series membership.
2. Import a different book from the same Hardcover series via Audiobookshelf, so the sibling walk links the first book with `primary_series = 0`.
3. `sqlite3 bindery.db "SELECT series_id, primary_series FROM series_books WHERE book_id = N"` shows one row, `primary_series = 0`.
4. The book page shows the series. Preview the destination or rename the file: the series folder and the series part of the name are both absent.

## Trace logs

Not applicable. No error occurs and nothing is logged; `GetPrimarySeriesForBook` returns a nil error with empty strings, and the renamer treats that as "this book has no series".

## Deployment method

Docker (ghcr.io)

## Version

v1.35.0

vavallee:
Closes #2527. **Stacked on #2526** — base it there, or merge #2526 first; this branch contains that commit.

## What was wrong

`GetPrimarySeriesForBook` (`internal/db/series.go`) filtered on `primary_series = 1`. A book whose every `series_books` row is secondary matched nothing and got `("", "", nil)` — the same answer as a book in no series. `internal/importer/renamer.go:248-250` drops a path segment that renders empty, so the file was written with no series folder and no series in its name, while `BookDetailPage` and the series pages went on listing the membership. Nothing logged, because no error occurred.

Reachable through `linkHardcoverCatalogSiblings` (`internal/abs/import_hardcover_series.go:518-530`), which links books it is not importing with `primary = false`. That `false` is right in intent: the run is importing a different book and must not claim the sibling's primary series. But when the sibling had no membership at all, which is normal for a book imported from Audiobookshelf without series metadata, that row is the only one it will ever have.

## What changed

**`GetPrimarySeriesForBook` drops the `WHERE` and sorts on `primary_series DESC` ahead of the #2525 tie break.** A primary membership still outranks a secondary one; with nothing but secondaries the book gets its series instead of nothing. This repairs existing libraries on upgrade with no migration, which is why it is the fix rather than only touching the link site.

**The sibling walk uses `LinkBookPreservingPrimary`.** A sibling with no primary series gains one, a sibling that already has one keeps it.

## Verification

Fail-before evidence: with the `WHERE sb.book_id = ? AND sb.primary_series = 1` filter restored, `TestSecondaryOnlyMembershipStillNamesFiles` fails with

```
series_test.go:1225: secondary-only membership = ""/"", want King's Dark Tidings/1
```

That test also pins the two boundaries the change must not cross: a primary membership still outranks a secondary one carrying a position, and a book in no series is still the empty answer.

Green: `go build ./...`, `go vet ./...`, `go test ./...`, `golangci-lint run ./internal/db/... ./internal/abs/...`.

No frontend change; the UI was already showing these memberships, which is half of why the mismatch was invisible.


vavallee:
Auto-closed when its base branch `fix/2525-primary-series-choice` was deleted on merging #2526, and GitHub will not reopen a PR whose base ref is gone. Same branch, same commit, reopened against `main` as the PR below.

Issue #2011, opened 2026-08-14
Writing metadata breaks seeding: tagging a hardlinked import mutates the file the torrent client is still serving


vavallee:
## Problem

Writing metadata into an imported file changes its bytes, which changes its hash, which means the file can no longer be seeded as the torrent it came from. Reported by Daize on Discord (2026-08-13):

> It does seem like updating the metadata does not keep the file "original enough" to be shared back through MAM. So that means a copy of the file is required... is that just an expected limitation, or is there a nice way to "enrich but don't change the checksum of the original file"?

They also asked the general version in #general the same morning, and it went unanswered:

> How does everyone manage keeping files the same for seeding, ideally using hardlinks I guess to avoid creating lots of copies, but also allowing for improving metadata?

That is the real tension. Hardlink import exists precisely so one copy on disk serves both the library and the torrent client. Writing tags into that file mutates the same inode the client is seeding, so the choice today is silently between seeding and enriched metadata, and nothing in the UI says so.

## Why this is not just "use copy mode"

Copy mode solves it at the cost of the thing hardlink mode exists for. On a large audiobook library that is a real amount of disk, and the operator picked hardlink deliberately.

## Prior art

[lidarr/Lidarr#1016](https://github.com/Lidarr/Lidarr/issues/1016) is the same problem in the music domain, cited by the reporter.

## Possible directions

None of these are obviously right and the tradeoffs differ:

- **Break the hardlink before writing** (copy-on-write for tagged files only), so the seeded file keeps its original bytes and the library gets an enriched copy. Costs disk for tagged files only rather than for everything.
- **Do not write tags to hardlinked files at all**, and surface that as a visible state rather than a silent one. Cheapest, and arguably the honest default.
- **Make it a setting** with the tradeoff spelled out where the import mode is chosen, next to the hardlink warning added in #1720.

## Worth checking first

Which paths actually write into the file after import, and whether the audiobook and ebook sides differ. Calibre `calibredb add` copies into a Calibre library rather than mutating in place, so the exposure may be narrower than it looks.

flaevers:
Not sure if relevant here, but I just read that Lidarr e.g. supports reflinks (of course it's still required that someone's setup supports it) - that way, if I understand correctly, metadata can be modified and it won't break seeding.

PR #2549, opened 2026-09-11
feat(importer): write an opt-in metadata.opf sidecar per book

gchahcg:
## Summary

Ebooks and audiobooks pulled from different sources can carry differently-spelled author names in their embedded metadata — punctuation, initials, diacritics ("J.R.R. Tolkien" vs "JRR Tolkien" vs an ASCII-transliterated "Joerg Mueller"). `textutil.MatchAuthorName` already reconciles this correctly for Bindery's own catalogue and folder naming, but that reconciled truth never goes anywhere else: `epubmeta.go` and `audiotag.go` are read-only (confirmed — no ID3/OPF/M4B writer exists anywhere in the codebase), and Bindery only ever renames files, never writes into or beside one. A downstream library app that trusts embedded (or sidecar) metadata over folder structure — this came up specifically with BookOrbit and Grimmory — still sees the inconsistency even though Bindery's own view looks perfectly clean.

This is genuinely open territory, not a declined request: `docs/ROADMAP.md`'s "Won't do" / "Explicitly out of scope" sections don't mention it either way.

This PR adds **Phase 1** of a two-phase plan and lays out **Phase 2** below for you to weigh in on — I'd rather you see the whole shape than have it trickle in as surprise follow-ups.

## Phase 1 (this PR): a `metadata.opf` sidecar

New opt-in setting `import.write_opf_sidecar` (off by default). When on:

- Every imported ebook/audiobook folder gets a Calibre-style `metadata.opf` — the same OPF package-document `<metadata>` schema and Dublin Core vocabulary an EPUB carries inside its own zip (`internal/importer/epubmeta.go`'s `ReadEpubMetadata` is the read-side counterpart of the same format), written standalone instead of unzipped from inside a book file.
- Carries Bindery's own canonical DB metadata — title, sort-title, author (name + file-as sort form), narrator, series + series-index, identifiers (ISBN/ASIN/provider IDs), language, publisher, publish date, description, genres — regardless of what the downloaded file's own tags said.
- Refreshed by Reorganize too (`internal/importer/reorganize.go`'s `applyOne`), so a later author rename or alias merge doesn't leave the sidecar stale.
- **Zero risk to the book file's own bytes** — this is a new file only, never a modification to the file it sits beside. That's deliberately the whole reason this is Phase 1 and not Phase 2: it sidesteps hardlink import mode entirely, where the "imported" file and the original download share an inode and any embedded-tag write would mutate the seed file too.

### Field scope

Title, sort-title, author, narrator, series, identifiers, language, publisher, date, description, and genres are written. **`Book.AverageRating` and `Book.RatingsCount` are deliberately excluded**, and so is any cover reference (no cover file is ever placed in the library folder to reference — covers are proxied/cached separately at `/api/v1/images`).

This isn't a new policy call — it's the same line `docs/third-party-data.md` already draws for Hardcover-sourced data: *"Title, author, series, edition, publisher, narrator, description, and cover are facts and are fine"* vs. *"`books.average_rating` and `books.ratings_count`... are aggregates of other users' ratings, not facts about the book. Strip them..."* A `metadata.opf` sidecar is exactly the kind of new surface that line exists to keep closed, so this PR keeps to it rather than opening a new gap. `TestBuildOPFXML_ExcludesRatings` pins this as a regression test, citing the doc directly.

### Implementation notes

- `internal/importer/opfsidecar.go` is new, sibling to `epubmeta.go`. `BuildOPFXML` is hand-built XML (not `encoding/xml` struct tags) — namespaced attributes (`opf:role`, `opf:file-as`, `opf:scheme`) are awkward to get byte-stable through Go's marshaller, and `parseOPFMetadata` already established this codebase's precedent of walking/writing OPF tokens directly rather than binding a fixed struct.
- Reuses `calibre.IdentifiersForBook`, `calibre.NormalizeLanguageForCalibre`, and `calibre.FormatPublishedDate` from the existing Calibre-push integration (`internal/calibre`) rather than reimplementing identifier/language/date resolution. A book's `metadata.opf` and its `calibredb`-pushed metadata are guaranteed to agree by construction, not by two implementations happening to match.
- Hook points mirror the existing best-effort push pattern exactly (`pushToCalibre`/`pushToCWA`/`pushToGrimmory` in `scanner.go`, `pushToCalibre`/`pushToABS` on the audiobook side) — logged and swallowed on failure, never rolling back an otherwise-good import.
- Sidecar writes are atomic (temp file + rename, matching `calibre.MaterializeCover`'s existing pattern), and for the ebook path happen once per book after the whole file loop completes — not once per file — so a multi-format download (epub+mobi+pdf sharing one folder) doesn't regenerate the identical sidecar three times, and so the write happens *after* the `#1160` embedded-language backfill rather than before it (a first import with no catalogue language previously risked a sidecar with a blank `dc:language` for the seconds before the backfill landed).
- Reorganize removes an orphaned `metadata.opf` when it's the last thing left in a folder being pruned, so a rename never strands an empty folder that `os.Remove` can't clear.

### Where it's mounted

- Settings → General, next to the drop-folder/flatten-multi-disc import toggles (`import.write_opf_sidecar`).
- `internal/importer/scanner.go`: ebook hook after the per-file loop (once per book), audiobook hook after `pushToABS`.
- `internal/importer/reorganize.go`: `applyOne`, after the history event, gated the same way.

## Phase 2 (proposed, not built here): embed metadata into the file itself

Phase 1 doesn't close the loop for every downstream app — I checked both concretely:

- **BookOrbit** supports OPF-sidecar-as-a-metadata-source with a configurable precedence order, but there's a [currently open upstream bug](https://github.com/bookorbit/bookorbit/issues/200) where the sidecar is ignored even when ranked first.
- **Grimmory** uses its own `.sidecar.json` convention and doesn't consume `.opf` sidecars at all (though it does have its own manual "write metadata into the EPUB" feature, and Bindery already has a direct `pushToGrimmory`/BookDrop integration that's a separate mechanism from this).

The real fix for both is embedding metadata directly into the file:

- **EPUB is a near-trivial follow-on.** It's the exact same OPF `<metadata>` schema/vocabulary this PR already generates — Phase 2 would splice `BuildOPFXML`'s `<metadata>` block into the EPUB's *internal* package document (found via `META-INF/container.xml`, same as `ReadEpubMetadata` does today) instead of writing it standalone, and rewrite the zip. No new format to learn, same generation function, different write target.
- **Audio formats need real work.** ID3 (mp3) and MP4 atoms (m4b/m4a) are different binary container formats with no shared vocabulary — this needs an actual tag-writing dependency (checked: `dhowden/tag`, the only tagging dep currently in `go.mod`, is read-only) and a license check per the project's MIT-vs-copyleft conventions.
- **The one real risk, stated plainly:** under `hardlink` import mode, the "imported" file and the original download share an inode. Writing tags into one mutates the other's bytes too. Phase 2 would need to detect hardlink mode and either skip the write or force a copy first — this is exactly the risk Phase 1 was scoped to avoid entirely, and it doesn't go away for Phase 2, it just becomes a solvable, scoped problem instead of an unscoped one.

I'm not asking for a decision under time pressure — Phase 1 stands on its own regardless of which way you land on Phase 2. But I'd rather you see the full shape now than have "embed into EPUB" show up as a surprise follow-up PR with no warning it was coming.

## Suggested review order

1. `internal/importer/opfsidecar.go` + `opfsidecar_test.go` — the core logic and its test contract (especially `TestBuildOPFXML_ExcludesRatings`).
2. `internal/importer/scanner.go` + `reorganize.go` — the hook wiring, including the orphaned-sidecar cleanup and the post-loop write ordering.
3. `internal/api/settings_handler.go` + `settings_registry.go` (+ `settings_registry_test.go`) — the setting plumbing.
4. `web/src/pages/settings/GeneralTab.tsx` + `en.json` — the UI.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — `README.md` (Import & organize feature list), godoc on every new exported symbol, `changelog.d/opf-sidecar-metadata.md`. Not touched: `docs/DEPLOYMENT.md` — happy to add a line there too if you'd like one (its closest analogue, `cwa.ingest_path`, gets a topology mention; the closer sibling toggles `drop_pair_gating`/`flatten_multi_disc` don't).
- [ ] Wiki pages updated — n/a, no existing wiki page covers this surface

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -l .`
- [x] `golangci-lint run` (`internal/importer/...`, `internal/api/...`)
- [x] `govulncheck ./...`
- [x] `go test ./internal/importer/...`, `./internal/api/...`, `./internal/calibre/...`
- [x] `go test -race ./internal/importer/...`, `go test -race -run TestSetting ./internal/api/...`
- [x] `cd web && npm run typecheck && npm run lint && npm run build && npm test` (861/861)

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ Clean-room Go rewrite, modern React UI, MIT-licensed, actively developed.
 - Calibre integration in three modes: `calibredb` CLI hook on import, [Bindery Bridge plugin](https://github.com/vavallee/bindery-plugins) (cross-container), or direct read of an existing Calibre library's `metadata.db` as Bindery's catalogue.
 - **Audiobookshelf import** — pull an existing ABS server's book libraries in as Bindery's catalogue (metadata-first, dry-run, review queue for ambiguous matches, rollback), with an ABS library-scan trigger after every audiobook import. See [docs/ABS-Import-Wiki.md](docs/ABS-Import-Wiki.md).
 - **Grimmory push** (preview) — imported ebooks are sent to a self-hosted [Grimmory](https://grimmory.org) library via its BookDrop inbox, with a bulk **Push all** for existing files.
+- **`metadata.opf` sidecar** (opt-in) — write a Calibre-style `metadata.opf` next to each imported book, carrying Bindery's own canonical title/author/series/identifiers/etc. so a library app that reads sidecar metadata sees consistent data regardless of which source the file came from. Refreshed on Reorganize.
 
 **Metadata sources** — all stable, documented, public APIs. No Goodreads scraping.
 
--- a/changelog.d/opf-sidecar-metadata.md
+++ b/changelog.d/opf-sidecar-metadata.md
@@ -1,0 +1,2 @@
+### Added
+- **Optional `metadata.opf` sidecar per book** — when `import.write_opf_sidecar` is enabled (off by default), Bindery writes a Calibre-style `metadata.opf` next to each imported ebook/audiobook, carrying Bindery's own canonical title, author, series, identifiers, language, publisher, narrator, description, and genres — regardless of what the downloaded file's own embedded tags say. Refreshed by Reorganize too, so a later author rename or metadata edit doesn't leave it stale. Never touches the book file itself, only adds this extra file alongside it.
--- a/internal/api/books.go
+++ b/internal/api/books.go
@@ -886,6 +886,12 @@ func removeBookPathScoped(p, format string, ownedByOther func(string) bool) erro
 		}
 	}
 
+	// A book file's metadata.opf sidecar (import.write_opf_sidecar) is not a
+	// book file itself, so it survives the sweep above and would otherwise
+	// strand this folder the same way it used to strand a Reorganize move —
+	// reclaim it first so the empty check below actually sees empty.
+	importer.RemoveOrphanedSidecar(parent)
+
 	// Clean up parent directory if it is now empty.
 	remaining, err := os.ReadDir(parent)
 	if err == nil && len(remaining) == 0 {
--- a/internal/api/path_safety_test.go
+++ b/internal/api/path_safety_test.go
@@ -345,6 +345,34 @@ func TestSafeRemoveBookPath_OwnershipErrorFailsSafe(t *testing.T) {
 	}
 }
 
+// TestSafeRemoveBookPath_ReclaimsFolderWithOrphanedSidecar regression-tests a
+// PR review finding: import.write_opf_sidecar's metadata.opf is not a book
+// file, so it survived the old empty-folder check in removeBookPathScoped
+// and stranded every book's folder on delete, holding nothing but a stale
+// sidecar. removeBookPathScoped now reclaims an orphaned sidecar the same
+// way Reorganize already did (RemoveOrphanedSidecar), before checking empty.
+func TestSafeRemoveBookPath_ReclaimsFolderWithOrphanedSidecar(t *testing.T) {
+	dir := filepath.Join(t.TempDir(), "Book")
+	ebook := mustWrite(t, filepath.Join(dir, "Book.epub"))
+	if err := os.WriteFile(filepath.Join(dir, "metadata.opf"), []byte("<package/>"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	skipped, err := safeRemoveBookPath(context.Background(), nil, stubOwner{}, 1, ebook, "ebook")
+	if err != nil {
+		t.Fatalf("safeRemoveBookPath: %v", err)
+	}
+	if skipped {
+		t.Fatal("expected the delete to proceed, not be skipped")
+	}
+	if exists(ebook) {
+		t.Error("ebook file should have been removed")
+	}
+	if _, statErr := os.Stat(dir); !os.IsNotExist(statErr) {
+		t.Errorf("book folder should be gone once only the sidecar was left behind, stat err = %v", statErr)
+	}
+}
+
 // TestSafeRemoveBookPath_SiblingOwnedByAnotherBook is the file branch of the
 // same guard: the same-stem sweep must not take a sibling another book tracks.
 func TestSafeRemoveBookPath_SiblingOwnedByAnotherBook(t *testing.T) {
--- a/internal/api/settings_handler.go
+++ b/internal/api/settings_handler.go
@@ -131,6 +131,15 @@ const (
 	// never-arriving second format would wedge the first forever. The importer
 	// reads this key as a string literal; keep the literal in sync.
 	SettingImportDropPairGatingTimeoutHours = "import.drop_pair_gating_timeout_hours"
+	// SettingImportWriteOPFSidecar opts imports (and Reorganize moves) into
+	// writing a Calibre-style metadata.opf sidecar next to each book, using
+	// Bindery's own canonical DB metadata rather than whatever the source
+	// file happened to carry. "true" enables it; unset/"false" (default)
+	// keeps the prior behaviour — Bindery only ever renames files, it never
+	// writes a new one into the library. The importer reads this key as a
+	// string literal to avoid an import cycle; keep the literal in sync with
+	// this constant.
+	SettingImportWriteOPFSidecar = "import.write_opf_sidecar"
 )
 
 // SettingLibraryLastScan is the KV key holding the JSON summary of the most
@@ -673,6 +682,13 @@ func validateSettingValue(key, value string) error {
 			return nil
 		}
 		return fmt.Errorf("import.drop_pair_gating %q is not one of: true, false", value)
+	case SettingImportWriteOPFSidecar:
+		// Boolean flag; empty or "false" = off (default). Only accept the two
+		// canonical values so a typo can't be silently misread as truthy.
+		if value == "" || value == "true" || value == "false" {
+			return nil
+		}
+		return fmt.Errorf("import.write_opf_sidecar %q is not one of: true, false", value)
 	case SettingImportDropPairGatingTimeoutHours:
 		// Empty falls back to the 72h default; a non-empty value must be a
 		// positive integer number of hours so a typo fails loudly here rather
--- a/internal/api/settings_registry.go
+++ b/internal/api/settings_registry.go
@@ -236,6 +236,11 @@ var settingDescriptors = []SettingDescriptor{
 		State:       SettingStateActive,
 	},
 	{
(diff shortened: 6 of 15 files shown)


vavallee (changes requested):
Nice work, and thanks for laying out Phase 2 up front instead of trickling it in. Off by default, atomic write, ratings kept out, external mode skipped (it returns before the hooks), all good.

One thing to fix before merge:

**Deleting a book's files strands the folder.** `removeBookFileAndSiblings` in `internal/api/books.go` only removes the parent when `os.ReadDir` comes back empty. With the sidecar on, `metadata.opf` is still there, so every deleted ebook leaves a folder holding just the sidecar. You handled this for Reorganize with `removeOrphanedSidecar`; the delete paths need the same treatment (single file delete, delete all files, and the format-scoped one). A test that deletes a book with the sidecar on and asserts the folder is gone would pin it.

On CodeQL: the three `go/path-injection` alerts are the same class as the ~15 already open on the importer's destination writes on main, same taint source. Not blocking.

On Phase 2: I'd keep it out. Writing into the file is exactly the #2011 problem, in hardlink mode the imported file and the seed share an inode, so any tag write mutates what the torrent client is serving. If it ever happens it'd need to be copy-mode only and clearly opt-in, but I'd rather see whether Phase 1 covers enough downstream apps first.

vavallee:
Small correction to my review: the function is `removeBookPathScoped` (`internal/api/books.go:839`), not `removeBookFileAndSiblings`. Same point, the empty-folder check at the end of it.

gchahcg:
CodeQL flagged 3 "high severity" path-injection alerts in `internal/importer/opfsidecar.go` (the `MkdirAll`/`CreateTemp`/`Rename` calls), tracing the `dir` argument back to author/title strings sourced from remote provider metadata (Hardcover/OpenLibrary).

I looked into it: `dir` is never raw at that point. Every caller derives it from `Renamer.DestPath`/`AudiobookDestDir`, both of which already sanitize each templated path component and call `ensureContained(dest, rootFolder)` before returning — a containment check whose own doc comment says it exists specifically "to prevent book-derived path components (e.g. author or title strings seeded from remote metadata) from escaping the configured library." CodeQL's interprocedural taint tracking just doesn't credit sanitization that happened several calls upstream, so it kept flagging the argument as tainted regardless.

Rather than leave it as an explained-away false positive, I added the same containment check directly at the sink: `WriteOPFSidecarFile` now takes the caller's configured library root(s) and refuses to write if `dir` doesn't resolve inside at least one of them (reusing `ensureContained`, same as the renamer), independent of whatever the caller already did upstream. All three call sites now pass their already-available root(s) through, and `TestWriteOPFSidecarFile_RefusesOutsideRoots` exercises both an out-of-root path and a `..`-traversal attempt directly.

Not strictly required to pass CI here (`lint`/`validate (Go)`/`Security Summary` are the required checks and all green), but wanted it resolved rather than just noted.

gchahcg:
Follow-up: pushed the containment check (`a8e2e47`), but CodeQL's `go/path-injection` query still flags the same 3 sinks even with `dirWithinRoots`'s explicit check directly guarding them. Looked into it — that query doesn't appear to recognize a custom validation function as a sanitizer barrier at all (a known limitation of CodeQL's Go security queries versus their more mature Python/JS equivalents), so no amount of in-code validation will turn this particular check green on its own.

The code is correct regardless (tested directly in `TestWriteOPFSidecarFile_RefusesOutsideRoots`, including a `..`-traversal case), and this repo's CodeQL workflow is the unmodified default with no query-filter config, so there's no repo-level suppression mechanism for this from a contributor's side — that's a Security tab dismissal on your end if you agree it's a false positive, not something I can or should do from here. Flagging so the red check doesn't look unexplained.

gchahcg:
Fixed in `5bcfd7a` — thanks for catching this and for the correction on the function name.

`removeBookDirScoped` (the audiobook/directory path) was already fine: its second pass removes any non-book-file leftover before `RemoveAll`, `metadata.opf` included. The bug was specifically in `removeBookPathScoped`'s single-file branch — its "clean up parent if empty" check didn't know the sidecar wasn't real content. Pulled the orphan-reclaim logic out of `removeOrphanedSidecar` into an exported `RemoveOrphanedSidecar(dir)` so `removeBookPathScoped` can call the same "sidecar is the sole remaining entry" check Reorganize already used, rather than duplicating it — call it once before the existing empty check, and the empty check does the rest.

Added `TestSafeRemoveBookPath_ReclaimsFolderWithOrphanedSidecar`, which does exactly what you asked: deletes a book's only file with `metadata.opf` present and asserts the folder is gone.

On Phase 2 — agreed, holding it out. Appreciate you engaging with the shape now rather than me finding out later it wasn't wanted.

vavallee:
Re-checked the follow ups on a local merge with main: build, vet, importer and api tests, golangci-lint all clean. The delete fix is right, and you were right that removeBookDirScoped already handled it via pass 2. I also threw some hostile author and title strings at dirWithinRoots and the renamer and could not get anything outside the root.

On CodeQL you have the diagnosis right. The Go path injection query only honours a handful of guard shapes and only when they sit in the same function as the sink on the same variable. It does recognise filepath.IsLocal, so one last change: swap dirWithinRoots for a filepath.Rel(root, dir) followed by filepath.IsLocal(rel), and feed filepath.Join(root, rel) into MkdirAll, CreateTemp and Rename. Roughly:

```go
var target string
for _, root := range roots {
    if root == "" { continue }
    rel, err := filepath.Rel(filepath.Clean(root), filepath.Clean(dir))
    if err != nil || !filepath.IsLocal(rel) {
        continue
    }
    target = filepath.Join(filepath.Clean(root), rel)
    break
}
if target == "" {
    return fmt.Errorf("opfsidecar: refusing to write outside configured library roots: %q", dir)
}
```

That shape is also a better containment check than the prefix one: it handles a root of / and does not depend on string prefixes. IsLocal(".") is true so dir == root still passes. With that in and the three alerts gone I will merge.

gchahcg:
Pushed in `66076a6` — swapped the containment check for exactly the `filepath.Rel` + `filepath.IsLocal` shape you described, `target` built from `filepath.Join(root, rel)` and fed into `MkdirAll`/`CreateTemp`/`Rename` in the same function. Removed the old `dirWithinRoots` helper since the check now lives inline where CodeQL needs to see it. Existing containment tests (out-of-root path, `..` traversal, valid-path sanity check) all still pass against the new implementation, build/vet/lint/full importer+api suite all clean.

Thanks for actually tracking down which guard shape the query credits rather than leaving it as an accepted false positive — appreciate the specificity.

vavallee:
That's the one, thanks. CodeQL is green on it now and the tests still cover the out of root and traversal cases. Merging.

Issue #2548, opened 2026-09-11
Prev/next navigation between authors (and books) without returning to the list


lucaslra:
## Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I checked the roadmap and this isn't already planned

## Problem

You can't move between authors without going back to the Authors list first.
Opening one author, then the next, means: back to the list, find your place
again, click in. Same problem on the Book detail page.

## Proposed solution

Add "‹ Previous" / "Next ›" controls to the Author and Book detail pages that
step to the adjacent entry in the list you came from (same sort, same
filters) — no need to leave the detail view. Arrow-key shortcuts would be a
nice add-on.

## Alternatives considered

- **Modal/drawer over the list** instead of a full page — keeps the list
  visible, but that's a much bigger change.
- **Browser back button** — this is the current workaround, and it's the
  thing this issue is about fixing (still means re-finding your place each
  time).

vavallee:
Makes sense, the back-then-find-your-place loop is annoying. Stepping through in the same sort and filters you came from is the right shape, and it's a frontend-only change since the list state is already kept client-side. Leaving this open as an enhancement; no promise on when, but it's a small one.

lucaslra:
Tried to give it a go myself :)
My first attempt had a bit more functionality, but it grew bigger than felt right for this issue, so I went with a minimal, frontend-only version instead — #2552

PR #2552, opened 2026-09-11
Add Previous/Next navigation across authors

lucaslra:
## Summary

Adds Previous/Next navigation on the author detail page, stepping through the authors on the list page you came from — no round trip back to the list to find your place. Frontend-only, per your comment on the issue. Part of #2548 (book side not included — see scope note below).

## Implementation notes

- `AuthorsPage` already loads its current page (in whatever search/sort/monitored order) into React state before a row is ever clicked, so no backend endpoint is needed: each author `Link` now carries `{ ids, index }` as router `state`, and `AuthorDetailPage` reads it back to render Previous/Next and step the chain (`index - 1` / `index + 1`) client-side.
- Not shared as an exported type between the two pages — each side builds/reads the same `{ ids, index }` shape independently, the same idiom already used for the series picker's `{ seriesId }` router state between `AuthorsPage` and `SeriesPage`.
- Kept deliberately minimal: no position counter, no disabled-button state at either end (a direction with nowhere to go simply isn't rendered), no client-side whitelist on the sort value — none of those were asked for, and the backend's `authorSortOrder()` already falls back safely on an unrecognized sort value.
- "← Back" goes to the Authors list only when there's Previous/Next nav state (i.e. you arrived via a list row or a chain hop) — a few hops through Next/Previous land one page short of where the list actually was, so history alone isn't enough there. With no nav state (Wanted, the Books list, a book page, a direct link) it falls back to `navigate(-1)`, so those flows land back where they came from instead of always on the Authors list.

## Review fixes

- The page stays mounted across Next/Previous (only the `:id` param changes), which leaked two bits of state between authors:
  - **Group-by-series data**: the lazy-load effect skipped refetching once `authorSeries` held *any* data, so hopping to a new author kept grouping its books against the previous author's series. Now tracked by a ref recording which author's series are loaded, so a new `authorId` always triggers a refetch.
  - **Error banner**: an error from the previous author (e.g. a failed load) stayed visible after hopping to the next one. The load effect now clears it before each fetch.

- **Trade-off, accepted deliberately:** this only reaches as far as the currently loaded list page (its own page size), and does not survive a hard refresh or a direct link — router state is gone in both cases, and the controls simply don't render then, same as when opened from somewhere that never set the state.
- **Filter preservation on Back is a separate PR** (not included here): carrying the list's search/sort/monitored filter across Back and the Previous/Next chain is a related-but-distinct fix with its own code path, kept out to match "one branch, one change."
- **Book side of #2548 is out of scope for this PR** (per review) — not marking this as closing the issue since only the author half is done here.

## Checklist

- [x] Tests added or updated
- [x] `docs/DEPLOYMENT.md` — n/a, no env vars/config/upgrade-path changes
- [x] Wiki pages — n/a, no wiki page covers per-page list-navigation UX

## Test plan

- [x] `cd web && npx tsc --noEmit`
- [x] `cd web && npm test` (871 tests)

No Go files changed in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Cover images are fetched and cached server-side under `<dataDir>/image-cache/` (
 - Modern React 19 + TypeScript + Tailwind CSS SPA with search-first author acquisition and deep-linkable routed `/book/:id` and `/author/:id` pages.
 - Light / dark themes (respecting `prefers-color-scheme` first paint), grid / table view toggles, mobile-friendly responsive layout, hamburger nav, agenda-style mobile Calendar.
 - Full pagination, search, filter, and sort on every list page; preferences persist to `localStorage`.
+- Previous/Next navigation between authors on the author detail page, stepping through the list page you came from.
 - 8 languages — English, French, German, Dutch, Spanish, Filipino (Tagalog), Indonesian, Korean — auto-detected from the browser, override in Settings.
 - **OPDS 1.2 catalogue** at `/opds/` for KOReader, Moon+ Reader, and other reading apps. HTTP Basic auth with the API key as the password.
 
--- a/changelog.d/2548-author-nav-frontend.md
+++ b/changelog.d/2548-author-nav-frontend.md
@@ -1,0 +1,2 @@
+### Added
+- **Previous/Next navigation on the author detail page** (#2548) — step through the authors on the list page you came from, with no server round trip. Frontend-only: works within the currently loaded list page and doesn't survive a hard refresh or a direct link.
--- a/web/src/i18n/locales/en.json
+++ b/web/src/i18n/locales/en.json
@@ -228,6 +228,12 @@
     "refreshMetadata": "Refresh metadata"
   },
   "authorDetail": {
+    "nav": {
+      "previous": "‹ Previous",
+      "next": "Next ›",
+      "previousAriaLabel": "Previous author",
+      "nextAriaLabel": "Next author"
+    },
     "description": {
       "showMore": "Show more",
       "showLess": "Show less"
--- a/web/src/pages/AuthorDetailPage.test.tsx
+++ b/web/src/pages/AuthorDetailPage.test.tsx
@@ -94,7 +94,13 @@ function LocationProbe({ onLocation }: { onLocation?: (location: string) => void
   return null
 }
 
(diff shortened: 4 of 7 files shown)


vavallee (changes requested):
Nice minimal shape, and the state validation against authorId is the right call. Two things before merge:

1. The page stays mounted across Next, so some state leaks between authors. With Group by series on, step from one author to the next and listAuthorSeries is never fetched for the new author (the effect at line 129 skips because authorSeries is already populated), so the new author's books are grouped against the old author's series. The error state survives the hop too. Easiest fix is resetting those in the authorId load effect, or giving the page a key={id} so it remounts.

2. Making Back always go to / breaks the other ways into this page. From Wanted, the Books list or a book page, Back now lands on the Authors list instead of where you were. Keep navigate(-1) when there is no nav state and only use the list link inside a Prev/Next chain.

Also, the issue covers books as well, so either drop Closes #2548 or add the book side in this PR. Happy with the rest.

lucaslra:
Thanks for the thorough look — fixed all three:

1. **State leak across Next/Prev** — `authorSeries` refetch was gated on `authorSeries.length > 0`, which stayed true forever after the first author, so a hop to a new author kept grouping its books against the previous author's series. Now tracked by a ref recording which author's series are loaded, so a new `authorId` always refetches. The error banner is also cleared at the start of each author's load, so a stale error no longer survives the hop.

2. **Back** — now only goes to the Authors list when there's Previous/Next nav state (arrived via a list row or a chain hop). With no nav state — Wanted, Books, a book page, a direct link — it falls back to `navigate(-1)` instead of always landing on Authors.

3. **`Closes #2548`** — dropped, changed to "Part of #2548". Book side stays out of scope for this PR — `BookDetailPage` carries a lot more state than `AuthorDetailPage` and is linked from five different list pages instead of one, so the same leak-audit work would be substantially bigger. Happy to open a follow-up for it.

Added tests for all three. `tsc`/lint/871 tests/build all green. Pushed as 406035c0.

vavallee (approved):
All three land where I wanted them, and the tests for each are the right shape. One small follow through pushed to your branch: the ref stops the stale skip, but the previous author's series still sat in state until the new fetch resolved, and stayed there if it failed, so I clear the list before the fetch. Page tests still pass. Merging once CI is green. Follow up for the book side is welcome, agreed it is a bigger audit.

PR #2576, opened 2026-09-14
Add Previous/Next navigation across books

lucaslra:
## Summary

Adds Previous/Next navigation on the book detail page, stepping through the
list you came from (Books, an author's own book list, or Wanted) — no round
trip back to find your place. Frontend-only, same shape as #2552. Completes
#2548.

## How it works

- Each list page hands its currently-loaded `{ ids, index, hopDepth }` to
  `BookDetailPage` as router state — no backend change needed.
- `hopDepth` lets Back skip the whole Previous/Next chain in one jump
  (`navigate(-hopDepth)`) instead of landing one book short. There's no
  single "the list" to hard-code Back to, since a book can be reached from
  several places.
- The detail page remounts on `:id` change (`key={id}`) to reset its ~10
  pieces of book-scoped state in one shot, instead of auditing each one.

## Notes

- Chain state survives a same-tab hard refresh (browser history persists it),
  but not a fresh navigation (direct link, bookmark, new tab).
- Scope matches the merged author-side PR: only Books, an author's book list,
  and Wanted carry nav state. Series/History/Queue links don't.
- Caught on self-review: Back landing one book short after several hops
  (fixed via `hopDepth`).

## Checklist

- [x] Commits signed off
- [x] Tests added/updated
- [x] Changelog fragment added
- [x] No deployment/wiki changes needed

## Test plan

- [x] `npx tsc --noEmit`, `npm run lint`, `npm test` (911 tests), `npm run build`
- [x] Manually verified in browser: multi-hop nav from all three sources,
      Back skip, hard refresh, first/last-position edge cases

No Go files changed.


vavallee (changes requested):
Thanks, nice follow up, and key={id} is the right call. I tried to break it with a throwaway test that hops between two books with an error banner, a series row and a late series response on the first one, and nothing leaks. Back with hopDepth holds up through browser back mid chain and a same tab reload too.

A few things before merge:

1. With Group by series on, Next follows filteredBooks order instead of what's on screen. If Elantris loads first but shows last under Standalone, Next from Elantris jumps to Mistborn 1, and Mistborn 2 has no Next at all. #2548 asks for the order you see, so I'd build the ids from seriesGroups when grouping is on (first occurrence wins for a book in two series). The test at AuthorDetailPage.test.tsx:1169 would flip.
2. The ASIN draft test still passes with key={id} removed, because the load effect already resets asinDraft on every id change. Something the load effect doesn't touch, like the error banner from a failed save, would actually guard the remount.
3. Small one: the nav row sits below the loading and not found early returns. So every hop blanks the header, keyboard focus drops to the body, and a book deleted since the list loaded leaves you on Book not found with no Back or Previous. Rendering that row above those returns fixes all three.

Heads up on ordering: #2563 (the unified Add modal) is about to land and also adds a const navigate = useNavigate() to BooksPage. Git merges the two cleanly but tsc then fails on the duplicate declaration, so after a rebase just drop your line and reuse the existing one.

lucaslra:
Thanks for the review — all four addressed, rebased onto `main`.

1. **Group by series order** — `bookIds` now follows `seriesGroups` (each series section, then Standalone) instead of load order when grouping is on. A book in two series counts once, at its first-occurrence position. Flipped the test at `AuthorDetailPage.test.tsx:1169` and added a two-series case.
2. **ASIN draft test** — confirmed it wasn't a real guard. Added a test using a failed-save error banner instead, since the load effect never touches `error` on success. Verified it actually fails without `key={id}`.
3. **Nav row placement** — hoisted above the loading/not-found returns; added tests for both. One correction: focus still lands on `<body>` on every hop regardless, since `key={id}` remounts the whole subtree — that part isn't something this fixes, so I didn't claim it.
4. **`useNavigate` collision** — dropped my duplicate, kept yours from #2563.

Also drove the three UI fixes manually against the real dev backend (client-side `fetch` intercepts faking a series and a slow/404 response, no writes sent) before trusting the tests alone.

vavallee (approved):
All four points addressed, thanks, and good call being precise about the focus behaviour. Merged onto current main locally: typecheck, lint, 1002 vitest tests and the build all pass.

vavallee:
Merged, thanks @lucaslra. With #2552 that's Previous/Next on both sides now.

Issue #2361, opened 2026-09-05
GET /setting hands server filesystem paths to every authenticated user while /system/storage is admin gated


vavallee:
`GET /setting` and `GET /setting/{key}` are open to every authenticated role (`cmd/bindery/main.go:996-997`; the admin group starts at `:1000` and covers only `PUT` and `DELETE`). `SettingsHandler.List` filters on `isSecretSetting` (`internal/api/settings_handler.go:196-224`) and nothing else, so everything that is not a credential comes back, including:

* `calibre.library_path` (`internal/api/calibre.go:19`)
* `calibre.binary_path` (`internal/api/calibre.go:20`)
* `import.drop_folder` (`internal/api/settings_handler.go:107`)
* `cwa.ingest_path` (`internal/api/calibre.go:43`)

Meanwhile `GET /system/storage` is admin gated, with this rationale sitting right above it at `cmd/bindery/main.go:1094-1099`:

```go
// Storage paths (read-only view of the env/config-driven dirs plus
// exists/writable/hardlink-able health, #1183). Admin-only: it reveals
// server filesystem layout and writability probes.
```

Two endpoints, the same class of information, opposite gates. An OPDS only account can read the library path today.

Filing this as a discussion rather than a fix because the answer is not one line. Options, roughly:

1. Add an admin only classifier alongside `isSecretSetting` and decide which keys go in it. Filesystem paths are the obvious first entries; the harder question is what else (indexer and provider hostnames? ABS and Calibre URLs?).
2. Invert the default: `List` returns only an allowlist of keys the non-admin UI actually reads, and everything else needs admin. Safer, but it needs somebody to walk `web/src` and enumerate what the non-admin screens read, and it will break any third party client that reads a setting we did not think of.

Option 2 is the better end state and the more disruptive change. Either way it wants a decision on the key set before code, which is why there is no PR attached.

Worth checking against the S2 work on `isSecretSetting` in the same sweep so the two classifiers do not end up disagreeing about what is sensitive.

### Sequencing

Blocked on a decision, not on code. Should land after the `isSecretSetting` change from the S1/S2/S5 PR so both classifiers are edited once, in agreement. Not patch material.

vavallee:
Option 1 is up in #2418. Leaving this open on purpose, the PR does not close it.

[...]

PR #2604, opened 2026-09-14
fix(api): admin gate the library scan status, the second door to scan paths (#2361)

vavallee:
## Summary

`GET /library/scan/status` served the `library.lastScan` blob verbatim to every authenticated role. #2418 already reserved that key on `GET /setting` because it carries `library_dir`, `audiobook_dir`, `scanned_paths` and the absolute path of every unmatched file, so this route was the second door to the same value. It is now admin only, the same way `/system/storage` is.

Refs #2361. Not closing it: server paths still reach non admins through the reads listed under "Still open", and the allowlist and hostname questions on the issue are untouched.

## Consumers of the route

| Consumer | Where | Who reached it |
|----------|-------|----------------|
| Settings > General mount fetch | `web/src/pages/settings/GeneralTab.tsx:70` (before this PR) | every role: the fetch ran unconditionally, the panel rendered only for admins |
| Poll after "Scan library" | `web/src/pages/settings/GeneralTab.tsx:128` | admin only, the button sits in the `isAdmin` block |
| API client wrapper | `web/src/api/library.ts:39` | n/a |
| Vitest mocks | `SettingsPage.test.tsx`, `SettingsPage.absReview.test.tsx`, `GeneralTab.*.test.tsx` | tests |

Nothing else reads it: no scheduler job, no OPDS path, no smoke script, no doc. The importer writes the key and its own tests read it back through the settings repo, not the route.

## Why gate rather than redact

The only screen that shows the scan result is the Library block of GeneralTab, which lives inside the `isAdmin && (<>...</>)` fragment (lines 197 to 709). No non admin page shows scan progress or counts, so redaction would keep a response shape alive for no reader and add a third shape to the API. Gating matches `/system/storage` and the #2418 read gate: both use the role `auth.RequireAdmin` reads, so API key and trusted local requests (stamped admin by `auth.Middleware`) are unaffected.

## Implementation notes

| File | Change |
|------|--------|
| `cmd/bindery/sensitive_routes.go` | `registerLibraryScanStatusRoute` mounts the route inside a `RequireAdmin` group, the same testable registrar shape the log, indexer and OIDC routes use |
| `cmd/bindery/main.go` | calls the registrar in place of the inline `r.Get` |
| `web/src/pages/settings/GeneralTab.tsx` | the fetch moves into an effect keyed on `isAdmin`, so a non admin never asks and an admin whose auth status resolves after mount still loads it |
| `internal/api/settings_handler.go` | the comment that said this route was still open now says it is gated |
| `docs/API.md` | documents both library scan routes and the admin gate |
| `docs/multi-user.md` | capability row for server paths, and which General sections a non admin sees |
| `changelog.d/2361-scan-status-paths.md` | Security fragment |

`POST /library/scan` is left as it was (see below).

## Behaviour note

With `BINDERY_AUTH_MODE=disabled` and no session cookie the request carries no role, so this route now answers 403, the same as `/system/storage` and the #2418 setting reads already do in that configuration.

## Still open on #2361 (reported, not fixed here)

None of these is the same bug in the same shape: each has a non admin reader or needs a response shape decision.

| Read | Where | What leaks | Why not here |
|------|-------|------------|--------------|
| `GET /rootfolder` | `cmd/bindery/integration_routes.go:28`, handler `internal/api/root_folders.go:21`, field `internal/models/settings.go:43` | every configured library root `path` plus `freeSpace` | non admin add author and edit author flows read it (`web/src/components/authorAddDefaults.ts:31`, `web/src/components/EditAuthorModal.tsx:61`), so it needs a redacted shape and a picker that labels roots without the path |
| `GET /queue` | `cmd/bindery/main.go:971`, `internal/models/download.go:71` (`errorMessage`) | download paths formatted into import failures: `internal/importer/scanner.go:1435`, `internal/importer/scanner_handoff.go:319` | the message is the user facing failure reason; wants a rewrite of the reasons, not a gate |
| `GET /history` | `cmd/bindery/main.go:1000`, `data` blob | `path` keys at `internal/importer/scanner.go:2093`, `internal/importer/scanner_handoff.go:453`, `internal/importer/scanner_pair_gating.go:124`, and failure `message` via `internal/importer/scanner.go:815` | same as queue, plus history is per user content |
| Book responses | `cmd/bindery/main.go:941`, `:943`, `:954`; fields `internal/models/book.go:36` `filePath`, `:77` `ebookFilePath`, `:78` `audiobookFilePath`, `:83` `bookFiles` (`internal/models/book_file.go:14` `path`) | absolute file paths; with `BINDERY_ENFORCE_TENANCY` off, every user's | the book page shows files; needs a decision on what a non admin sees |

Also noticed: `POST /library/scan` (`cmd/bindery/main.go:1148`) is open to every role, while the capability matrix in `docs/multi-user.md` lists scanning as admin only. It is not a disclosure, so it is left for its own decision rather than folded in here.

Checked and clean for non admins: the arr compatible `GET /api/queue` record (`internal/api/queue.go:505`, no path or error fields), `GET /grimmory/config` (`internal/api/grimmory.go:106`, base URL and flags only, hostnames being the separate question on the issue), `GET /authors/refresh-all/status`, and OPDS, which uses the stored path only to pick the link and MIME type (`internal/opds/builder.go:438`). Already admin gated: `/system/storage`, `/abs/config`, `/calibre/*`, `/downloadclient`, manual import, reorganize, and the path keys on `/setting` (#2418).

## Fail before

Go: with the route extracted into `registerLibraryScanStatusRoute` but still mounted open (behaviour unchanged):

```
--- FAIL: TestLibraryScanStatusRouteRequiresAdmin
    sensitive_routes_test.go:424: status = 200; want 403 (RequireAdmin should reject role="user")
    sensitive_routes_test.go:424: status = 200; want 403 (RequireAdmin should reject role="")
--- PASS: TestLibraryScanStatusRouteAllowsAdmin
```

Web: before the `GeneralTab` change:

```
FAIL GeneralTab.scanStatusRole.test.tsx > does not request the scan summary for a non admin and shows no error
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 1 times
```

The admin case passed before and after. The "no error shown" assertions also passed before, because the old fetch swallowed every rejection; they are there to pin that a non admin sees nothing, not to prove the fix.

## Checklist

- [x] Tests added or updated
- [x] Doc update gate cleared (`docs/API.md`, `docs/multi-user.md`, godoc on the registrar)
- [ ] Wiki pages updated if user facing behaviour changed (no wiki page covers this route)

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/api ./cmd/... -count=1` (api ok 194.8s, cmd/bindery ok)
- [x] `golangci-lint run ./cmd/bindery/... ./internal/api/...` (v2.11.4, 0 issues)
- [x] `cd web && npm run lint` (0 errors, 6 existing warnings in untouched files)
- [x] `cd web && npm test` (80 files, 925 tests)

https://claude.ai/code/session_016fJcCVbNnKsmj2MAAMwWj9


vavallee:
Holding this until the auth disabled fix lands. With auth turned off, requests carry no role while the UI is told it is admin, so every RequireAdmin route already answers 403 on those installs (storage included), and this gate would make the scan panel go blank the same way. That fix goes first in its own PR, then this merges. Added a test for the admin role resolving late and filled in the admin section list in multi-user.md.

PR #2624, opened 2026-09-15
fix(importer): record a missing content path as an import failure (#2616)

### LINKED ISSUE #2616: A qBittorrent download whose files are gone stays in progress forever instead of failing
### Pre-flight

- [x] Searched existing issues. #969 and #1884 are close relatives (both closed); this is a different way into the same kind of stuck row
- [x] Found in source while working on #2289 (PR #2605)

### Current behaviour

When a torrent's content path no longer exists, the qBittorrent poll branch for grabbed and completed rows (`internal/importer/scanner_poll.go`, around :560 to :590) logs "content path not found, will retry next cycle" and leaves the row's status alone. The 120 cycle skip limit only counts `importFailed` rows, so a row in this state never gets there: it can sit in grabbed or completed forever, retrying every cycle, and the Queue page shows it as in progress.

The easiest way to hit it: import mode set to `move`, the book is imported (the files move out of the client's folder), the book is deleted, and the same release is grabbed again. qBittorrent adopts the torrent it still has, and there are no files left to import. The "already in library" guard doesn't catch it because the re-added book has no files yet.

### Expected behaviour

A row whose content path stays missing should end up somewhere the user can see and act on, the same way an import failure does: count the misses and move it to importFailed and then importBlocked with a message saying the files are gone from the client folder.

### Steps to reproduce

1. Import mode `move`, qBittorrent as the client.
2. Grab and import a book.
3. Delete the book, keep the torrent in qBittorrent.
4. Search and grab the same release again.
5. The Queue entry stays in progress and never imports; the log repeats "content path not found".

### Trace logs

Source review, no live trace.

### Also noticed

Two places say qBittorrent 5.x answers an add with JSON, but upstream only does that from 5.2 (5.0 and 5.1 still answer `Ok.`): the comment in `internal/downloader/qbittorrent/client.go` around :436 and `docs/Troubleshooting-Wiki.md` around :61.

### Deployment method

Any

### Bindery version

main after v1.36.0

tunglambk:
## Summary

A complete qBittorrent download whose content path is not on this host no longer sits in **grabbed** forever: the row records the miss, says why, and a content path that never appears reaches `importBlocked` instead of retrying silently every cycle.

Closes #2616.

## Implementation notes

Both qBittorrent poll branches had the same hole. The grabbed/downloading/completed branch logged "will retry next cycle" and left the status alone; the `importFailed` retry branch logged the same thing. Since the retry guard only counts `importFailed` rows, neither branch could ever move the row, so it stayed in progress for as long as the condition lasted.

The first branch now walks the row through `completed` — the state machine has no grabbed/downloading → importFailed edge, and the client does report the torrent complete on this path — and writes the reason the Queue shows. From `importFailed`, the existing `skipImportRetry` streak counts the misses and blocks the row once they run out, which is the re-grabbable, retryable end state the issue asks for. The retry branch counts the miss the same way; it deliberately does not rewrite the error message, because a changed message restarts the skip streak.

Two existing tests pinned the old "status untouched, retried next cycle" behavior. They now assert the recorded failure, while keeping what they were really guarding: no fall back to the shared download root (#3 — plus a library-dir walk, since an empty import of a non-book file would also end in `importFailed`), and no close-out of one format because its sibling is on disk (#1885).

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared — `docs/Troubleshooting-Wiki.md` "A grabbed book never imports" already describes this case as an `importFailed` with nothing at the path on this host, which is what the code now does; no env, config, upgrade, README or Helm surface changed
- [ ] Wiki pages updated if user-facing behaviour changed

## Test plan

- [x] `go test ./internal/importer/ -run TestCheckQbittorrentDownloads` — pass
- [x] `go test ./internal/importer/` — the same 9 pre-existing hardlink failures as `main` on this filesystem, everything else passes
- [x] `gofmt -l internal/importer/*.go` and `go vet ./internal/importer/` — clean

Regression evidence: with the source change reverted and the new test kept, `go test ./internal/importer/ -run TestCheckQbittorrentDownloads_ContentPathGoneMarksImportFailed` fails with `status = "grabbed", want "importFailed"`.

`internal/downloader/qbittorrent/client.go` around :436 and `docs/Troubleshooting-Wiki.md:61` still say qBittorrent 5.x answers an add with JSON; upstream only does that from 5.2 (5.0 and 5.1 still answer `Ok.`).


vavallee:
Thanks, this reads right and the state walk through completed is the correct edge. I ran the importer suite on the branch and it passes here too. Two small things before I merge it into v1.36.2:

1. The message in `importContentMissingReason` says "set PathRemap on the download client". In the UI that setting is called "path remap", so "set a path remap on the download client" matches what the user will look for.
2. Could you add a `changelog.d/2616-content-path-gone.md` fragment? A `### Fixed` section with one bold lead phrase and a user facing sentence, referencing (#2616). See `changelog.d/README.md`.

One note on scope: `resolveQbitContentPath` returns ok whenever `content_path` is set, so this branch only fires when qBittorrent sends an empty `content_path`. On a current qBittorrent the repro in #2616 already ends in importFailed through the "no book files found" path. Worth saying in the PR body so nobody expects it to change that case.

vavallee:
Correction on my earlier note: "PathRemap" is already the wording in the two neighbouring messages in scanner.go, so leave it as is. I pushed the changelog fragment to your branch so this can go out in v1.36.2. Thanks!

PR #2626, opened 2026-09-15
feat(filterengine): scored-signal catalogue filtering engine (#2235)

### LINKED ISSUE #2235: Additive scored signals instead of first-drop-wins boolean filtering
### Problem

`shouldFilterOLNoise` and the language / part-book checks around it in
`internal/metadata/openlibrary` are sequential boolean gates: the first one that returns
`true` decides the record, and nothing downstream is visible. That's fine when one rule is
clearly right or clearly wrong. It's a problem when several weak signals should have added
up to a decision, or when you want to know *what a record failed on* beyond the first check
that caught it.

This has a measured cost, not just a theoretical one. I dumped a real sync (Nora Roberts, a
profile with `unknownLanguageFail: true`, 2026-08-15) and found **60.9% of all raw provider
results across 14 sampled authors never reached `books[]`**. The single largest cause was an
*empty* `language` field being treated as *foreign*, not *unknown* — this is with
`FillMissingWorkLanguages`'s OpenLibrary edition-sampling (**#1058**) already active, so
it's a residual gap after the existing remediation, not a "hasn't been tried" one. The
dump's own `skippedLanguageSample` caught a case directly: `"Summer Stars: Second Nature /
One Summer"`, dropped for `language: ""` — an obviously real English-language bind-up that
edition-sampling didn't resolve.

This is exactly the failure mode **#1160** is already about. `unknown_language_behavior` is
configurable with a safe `pass` default, so this isn't "Bindery has a bug" — it's that the
`fail` option, when selected, silently eats 40-60% of a real catalogue, and there's no way
today to make that decision *gradually* instead of as a hard boolean.

### Proposed solution

Replace (or run alongside) the boolean chain with:

```
raw record → cluster (same normalized title) → signals emit Observations → sum → band
```

Each signal emits zero or more `Observation`s instead of a bool:

```go
type Observation struct {
    Signal     string  // "language.taggedNonEnglish"
    Direction  string  // "keep" | "exclude" — sign of Weight, for audit readability
    Weight     float64 // magnitude, from config (a metadata_profiles column)
    Confidence float64 // 0..1, computed in code, not config
    Reason     string  // human-readable, goes straight into an audit log
}
```

`Contribution = Weight * Confidence`, `Score = Prior + sum(Contribution)`, then banded:
`EXCLUDE` below a threshold, `KEEP` above another, `REVIEW` in between. **`keep == exclude`
in a profile reproduces `shouldFilterOLNoise`'s current binary behavior exactly** — this is
a superset of what exists today, not a replacement. A profile that wants strict pass/fail
still gets it unchanged; a profile that wants a curator queue gets one without a second code
path.

Two measured findings behind this, picked because one is a correction and one is negative —
not just numbers that support the idea:

- **Cluster-level `edition_count` discriminates real catalogue from noise; per-record
  doesn't.** Per-record, confirmed-real / misattributed / compilation works are
  statistically indistinguishable — all three sit around 61-65% with `edition_count ≤ 1`.
  Aggregated as `max(edition_count)` over records sharing a normalized title: **92% of real
  clusters are retained at a `≥3` gate, vs. 90% of noise clusters dropped.** It only works
  as a clustering-stage signal, which is why clustering has to be an explicit pipeline
  stage, not a per-record check.
- **Negative finding**: ISBN-presence-based omnibus tiering (Chaptarr's `SkipOmnibus` /
  `SkipMissingIdentifierOmnibus` split) does **not** transfer to OpenLibrary data — 96.7% /
  95.7% / 91.9% ISBN presence across confirmed-real / excluded / unknown records, a 5-point
  spread, not a useful discriminator here. Flagging this so the numbers below read as
  measured, not cherry-picked.

Worth noting Bindery's own code already grew toward this shape independently: `#2032` added
five more `Skipped*`/`Skipped*Sample` categories (part books, missing date, min popularity,
min pages, missing ISBN) alongside the existing language/junk/media-type ones — six parallel
counters now, each hand-wired into the sync loop, summed only at the very end via
`SkippedTotal()`. That's not a criticism — it's evidence the codebase is already drifting
this direction, just not unified into one code path yet.

Two more signals worth naming, both scoped specifically to avoid adding request churn — I
know rate limiting and bulk-import concurrency (#2075/#2101) have already been a real cost
here, so neither of these is "call another provider more":

- **A second `workLanguageFiller` for the #1058 residual.** `workLanguageFiller` is already
  an interface, already gated to run only when `allowed_languages != "any"` (#891), and
  OpenLibrary already implements it. Google Books' client already parses a per-volume
  `Language` field (`types.go`) but doesn't implement the interface yet. Chaining it in as a
  second filler — invoked only on whatever OpenLibrary's own edition-sampling still leaves
  at `Language==""`, not on every candidate — reuses the exact gate #1058 already has rather
  than adding a new call site. This is what would have caught "Summer Stars..." above.
- **Contributor-role, but only where the data is already in hand.** Hardcover's
  `contributions { contribution }` field (already fetched for author-identity resolution
  whenever Hardcover is a profile's primary provider, per #2040) can tell a real work apart
  from a foreword/translation/introduction credit — a case no title heuristic reaches. This
  should only read that field on the Hardcover-primary path where it's already in the
  response; it's a no-op for the more common OpenLibrary-primary profiles rather than a new
  Hardcover call for everyone.

### Alternatives considered

**Keep adding one-off boolean gates**, the `#2032` pattern — works, and is what's happening
today. The cost compounds: every new filter concept means a new counter, a new capped
sample, a new struct field, and another line in `SkippedTotal()`, forever, with no way to
see two weak signals stack.

**A learned/fitted model** — considered and rejected for this dataset. ~59% of records that
need a judgment call are unlabeled (raw pulls include a lot that's neither confirmed-real
nor confirmed-noise), and the biggest discrimination axis (catalogued-real vs.
catalog-noise) has no negative labels at all. Weights here are hand-set with documented
rationale instead, and any Bindery weights should start the same way — fit against real
data later if it turns out to help, never shipped as a serialized model.

### Additional context

Built and measured against a hand-verified ground-truth dataset, not designed in the
abstract:

- **[fiction-author-dataset](https://github.com/gchahcg/fiction-author-dataset)** —
  1,422 hand-verified rows across 20 authors (804 confirmed-real, 600 across 7 exclusion
  categories), CC-BY-4.0. `ARCHITECTURE.md` has the full pipeline-level findings, including
  the 60.9%/language-empty number above.
- **[fiction-author-dataset-engine](https://github.com/gchahcg/fiction-author-dataset-engine)**
  — `spec/bindery.md` is the full write-up this issue is condensed from (signal catalog,
  maintainer-review-constraint mapping, migration path). `pull/lib/engine.js` is the
  reference implementation (cluster → score → band, in JS, not proposing this as the Go
  implementation). `scripts/eval-filter.js` is the evaluation harness — it reports
  precision as a bounded range rather than a single number, since most of a raw provider
  pull is unlabeled rather than confirmed-bad, and conflating the two would overstate
  precision.

If this direction looks reasonable, I'd rather do it as one PR than stage it across several
smaller ones — a change that touches the scoring engine, the profile schema, and the
existing `shouldFilterOLNoise` call sites drifts and needs rebasing every time an unrelated
PR lands in between if it's split up, and the pieces don't mean much reviewed in isolation
anyway. Still fully parity-gated and non-breaking on landing, just delivered as one diff:
`metadata_profiles` gets `keep_threshold` / `exclude_threshold` columns defaulting to values
that reproduce `shouldFilterOLNoise`'s current behavior exactly (a parity gate, checked
against the dataset above), the existing checks get ported to signals emitting
`Observation`s at veto-equivalent weight, and clustering lands as an explicit stage — all
with `keep == exclude` as the only reachable profile shape at merge, so no existing profile
changes behavior. Enabling `keep != exclude` (the first graded signal, edition-count-cluster)
would be a deliberate follow-up once that's proven in, not part of the same PR. Full design
is in `spec/bindery.md` if useful background, §9 in particular walks the same steps this PR
would just do in one pass instead of five.

gchahcg:
## Summary

Replaces the author catalogue sync's sequential boolean noise-filtering (strict media type, junk title, provider-flagged noise, language, part-book, missing date, missing ISBN, min pages) with `internal/metadata/filterengine`: signals emit weighted observations, summed into a score and banded into KEEP/REVIEW/EXCLUDE. At v1 every signal ships at veto weight (-1000) and `metadata_profiles.keep_threshold`/`exclude_threshold` default to 0/0 for every profile — the exact value that reproduces the prior boolean-chain's keep/exclude decision, proven with a golden parity test (`TestRegistryIsVetoOnlyAtV1`, `TestAuthorSyncParity_EqualThresholdsReproduceBooleanChain`). This is a pure refactor: no existing profile's sync behavior changes.

This is the foundation for #2235's actual ask (additive scored signals instead of first-drop-wins boolean filtering). The first real graded signal (a cluster-level edition-count corroboration signal, validated offline against a 1,422-row hand-verified ground-truth dataset) is a focused follow-up PR on top of this one, opened separately so the risk-free refactor and the real behavior change get reviewed independently.

Closes #2235.

## Implementation notes

- `internal/metadata/filterengine` is a new package: `Signal`/`Candidate`/`Context`/`Decide`/`BandFor` core, plus one file per v1 signal (media type, junk title, provider noise, language, part book, missing date, missing ISBN, min pages). Registry order matches the pre-#2235 boolean chain's check order exactly, so `strongestObservation`'s magnitude tie-break reproduces first-drop-wins by construction, not by accident.
- `internal/api/authors.go`'s `fetchAuthorBooks` now does exactly two `Decide()` calls per candidate (free-signals pass, then free+structural+edition combined pass) instead of seven fragmented per-call-site checks — the actual #2235 gap: a candidate tripping more than one filter used to only ever record the first one a given call site happened to check.
- `filterengine.BuildClusters`/`Cluster` ship in this PR (no v1 signal reads a `Cluster` yet) as the explicit prerequisite the first graded signal needs — see `cluster.go`'s doc.
- OpenLibrary's client now *flags* companion material (study guides, film tie-ins) via `models.SignalProviderOpenLibraryNoise` instead of silently dropping it before the sync ever saw it. An independent review pass caught that `internal/abs/import_upserts.go`'s upstream title-match lookup doesn't route through `filterengine` and would otherwise start matching against flagged noise — fixed by skipping flagged works there directly (see the `fix(abs)` commit).
- `migration 086` adds `keep_threshold`/`exclude_threshold` to `metadata_profiles`; `validateScoreThresholds` currently rejects any non-zero value — relaxing that is exactly what the follow-up PR does, once a real graded signal exists to justify it.

- No UI changes: `keep_threshold`/`exclude_threshold` are DB/API-only in this PR, not surfaced in Settings. The follow-up PR adds a preset-based UI control (never raw threshold numbers).
- `AuthorSyncSummary.Total` is measurably higher for some OpenLibrary-primary authors now that companion material survives as a flagged, counted candidate instead of vanishing before accounting — see the changelog fragment.

- The first real graded signal (cluster-level edition-count corroboration), gated behind a closed set of server-tuned presets — separate PR, on top of this one.
- `catalogue_reconciliation.go` and a couple of `authors.go` helpers hand-check specific `filterengine` signal IDs instead of routing through `Decide` — harmless while every signal is veto-only, worth revisiting once graded signals are real.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (`changelog.d/2235-scored-filter-engine.md`, `docs/ARCHITECTURE.md`, `docs/User-Guide-Wiki.md`)
- [ ] Wiki pages updated if user-facing behaviour changed — no user-facing UI in this PR

## Test plan

- [x] `go vet ./...`
- [x] `golangci-lint run --timeout=5m` (0 issues)
- [x] `govulncheck ./...` (0 reachable vulnerabilities)
- [x] Independent multi-angle code review pass on the full diff (line-by-line scan, removed-behavior scan, cross-file consistency, conventions) — findings applied: an ABS import regression (this PR's own companion-material change leaking into `internal/abs`'s title matching, unrelated to `filterengine`) and a currently-unreachable `BuildClusters` edge case.


vavallee:
Closing along with #2628; I've explained the reasoning there, along with the smaller version I'd be glad to take. Thanks @gchahcg.

PR #2628, opened 2026-09-15
feat(filterengine): opt-in cluster-based noise filtering preset (#2235)

gchahcg:
## Summary

Stacked on #2626 — this PR is currently based on `main` but its first three commits are #2626's; once that merges, GitHub will automatically shrink this diff down to just the commits below. Review order: skip straight to the commits titled `feat(filterengine): ship ClusterEditionCountSignal...`, `docs: changelog and user guide...`, and `fix(filterengine): aggressive cluster preset defanged...`.

Adds `ClusterEditionCountSignal`, #2626's first real graded (non-veto) signal: it cross-checks each candidate work against how well-attested it is across every provider record sharing its title (clustered via #2626's already-shipped `filterengine.BuildClusters`), and can independently push a candidate toward KEEP or EXCLUDE based on that evidence. Gated behind a new metadata-profile field, `ClusterFilterPreset` — a closed set of server-tuned tiers (`off`/`conservative`/`balanced`/`aggressive`), never raw threshold numbers a client could hand-pick. `off` (the default for every existing profile) is provably byte-identical to current behavior.

Closes the rest of #2235.

## Implementation notes

- `internal/api/authors.go`'s `fetchAuthorBooks` now builds clusters once per sync (gated behind a non-`off` preset, so the default path pays nothing extra) and attaches each candidate's `Cluster` alongside the existing structural/edition signal pass — so a well-attested cluster's keep-direction observation can offset a `structure.partBookTitle` veto in the *same* `Decide` call.
- A non-`off` preset supplies its own shared threshold (applied to *both* `KeepThreshold` and `ExcludeThreshold`, never just one — see `ThresholdForPreset`'s doc for why keeping the two equal matters: it's what keeps the REVIEW band permanently unreachable, exactly as it is under the v1 veto-only default). This is a separate, additive knob from `KeepThreshold`/`ExcludeThreshold` themselves, which stay locked at exactly 0/0 by #2626's unchanged `validateScoreThresholds`.
- New metadata profile settings-form control (`web/src/pages/settings/MetadataTab.tsx`) — a preset selector, never raw scoring numbers.

## Validation

Tuning validated offline against `fiction-author-dataset`'s 1,422 hand-verified rows across 20 authors (CC-BY-4.0). Isolating the one known, filtering-unsolvable pen-name misattribution in that dataset (Nora Roberts/J.D. Robb — OpenLibrary attributes all of J.D. Robb's "In Death" books to the Nora Roberts identity, so no per-candidate signal can separate them), the `balanced` preset scores recall=90.9%/precision=83.7% on the other 19 authors, versus a lower baseline with the signal off.

An independent review pass caught a real bug before this shipped: `aggressive`'s original threshold (-1000) coincided exactly with the engine's veto weight, which — because `BandFor` uses a strict `<` — silently defanged *every* v1 signal whenever exactly one of them fired, not just this signal's own exclude branch. Fixed (now -950, with a genuinely looser keep gate to still earn the name), re-measured honestly, and pinned with a regression test (`TestThresholdForPreset_NeverDefangsALoneVeto`) that I verified fails against the old value before confirming it passes against the fix. Full writeup in that commit's message.

- Three other spec-derived signals (non-Latin-script title, thin-catalog-by-subject-count, lexical-foreign-language) were also built and grid-searched during tuning; they measured zero effect on this dataset and one has an identified misfire risk on romanized non-English titles. Left out of this PR — unvalidated, not worth shipping speculatively.
- No raw threshold UI. A profile can only pick one of the four named presets; the API rejects anything else.

## Checklist

- [x] Tests added or updated
- [x] Doc-update gate cleared (`changelog.d/2235-cluster-filter-preset.md`, `docs/User-Guide-Wiki.md`)
- [x] Wiki pages updated — new "Cluster-based noise filtering" section

## Test plan

- [x] `go vet ./...`
- [x] `golangci-lint run --timeout=5m` (0 issues)
- [x] `govulncheck ./...` (0 reachable vulnerabilities)
- [x] `cd web && npm run typecheck && npm run lint` (0 errors)
- [x] `cd web && npm test` (AuthorSyncNotice, MetadataTab)
- [x] Independent code-review pass on the Phase 2 diff specifically (separate from #2626's review) — the aggressive-preset defanging bug above was its finding.


vavallee:
Hi @gchahcg, thanks for the huge amount of work here, and for the measurement behind #2235. The finding that edition count only separates real works from noise once records are grouped by title is genuinely useful, and the writeup was a pleasure to read.

I'm going to close this and #2626 though, and that's partly on me: nobody replied on #2235 before you started, so you built a full scoring framework without anyone saying how much of it Bindery wanted. The honest answer is less than this. A general engine with weights, bands and presets is more machinery than I want to maintain for the catalogue filters, especially with most of it (the REVIEW band, the threshold columns) unused by design.

A review also turned up a few things that would have blocked it anyway, in case they're useful:

1. **Parity broke in the OpenLibrary client.** Noise works are now flagged instead of dropped in `openlibrary/client.go`, so `mergeAuthorWorks` can fold a real Hardcover book with the same title into the flagged row and the sync then excludes it. On main that Hardcover book gets added. The parity tests didn't catch it because they compare against hand written counts through a stub that skips the OpenLibrary client.
2. **Any preset excludes works with no edition count.** Only OpenLibrary search results fill `EditionCount`, so a Hardcover primary author loses their whole catalogue on conservative, balanced or aggressive (0 of 2 added in a test, and `skipped_thin_cluster` is missing from the sync log line).
3. **The keep bonus overrides filters the user turned on.** On balanced and aggressive it cancels part book, missing date, ISBN and min pages vetoes, which undoes the box set fix from #1780.
4. Housekeeping: three commits lack a sign off (the `gch ahcg` identity), migration 086 collides with #2607 and #2633, and the CC-BY sample doesn't credit the dataset author.

If you'd like to pursue the idea in a smaller form, I'd welcome a PR for just the useful part: one optional metadata profile filter that skips an OpenLibrary work when its title group's highest edition count is below a threshold, applied only to OpenLibrary works, with 0 treated as "no data" rather than noise, never overriding a filter the user set, and with a test that runs main's current filter functions on the same records. Worth a quick comment on #2235 first so we can agree the shape before you write it.

Thanks again, and sorry for the silence on the issue that led here.

Issue #2612, opened 2026-09-14
Add Book by ISBN takes a fallback provider's record when the primary doesn't answer


vavallee:
### Pre-flight

- [x] Searched existing issues; nothing covers this path
- [x] Found in source on main while reviewing #2610 (the importer half of #2332)

### Current behaviour

Add Book with an ISBN resolves the book through `ResolveBookByISBN`, which walks the providers in order. If the primary metadata provider times out or errors, the walk just moves on to the next provider and the book and its author get rewritten to that provider's ids.

Path: `internal/api/authors.go` Add Book, around `:3077`, then `resolveAuthorForBook` (around `:3480`), then `ResolveBookByISBN` (around `:3499`), then the ids are replaced with the fallback's (around `:3083`).

So during an OpenLibrary outage an ISBN add can bind to a DNB or Google Books record even when OpenLibrary has the right one. It's the same mistake #2332 fixes for the CSV, Readarr and Goodreads importers, just on the interactive path.

### Expected behaviour

When the primary didn't answer, don't silently take a fallback's record. Either say the primary is down and ask the user to try again, or show the fallback's match and let them confirm it. #2610 adds `ResolveBookByISBNWithOutcome`, which reports whether the primary failed, so the plumbing is there.

### Steps to reproduce

1. OpenLibrary as primary, DNB enabled.
2. Make OpenLibrary time out (block the host).
3. Add Book with an ISBN that both providers know.
4. The book and author come back with DNB ids.

### Trace logs

Source review, no live trace.

### Deployment method

Any

### Bindery version

main after v1.36.0

Issue #2332, opened 2026-09-05
CSV, Readarr and Goodreads importers skip the #2271 SafeToBind guard and hardcode metadata_provider=openlibrary, so a provider timeout binds authors to the wrong provider


vavallee:
Follow-up to #2117 and #2075. Line numbers are `main` at `c6895aed`.

#2271 established the rule: a caller that PERSISTS a provider link from an author search must use `SearchAuthorsWithOutcome` and consult `SearchOutcome.SafeToBind` before writing a foreign ID, because that link decides which provider the author syncs from forever. The doc comment naming the callers is at `internal/metadata/aggregator.go:68-73`: "the Audiobookshelf import, the Calibre relink, the add flow's canonical match".

The CSV, Readarr and Goodreads importers do exactly that kind of write and were never converted.

## What is wrong

Two independent defects on the same three paths.

**1. No `SafeToBind` guard.** The guarded callers are `internal/api/authors.go:1586` and `:2930`, and `internal/abs/import_author_matcher.go:539`. The migrate importers use the plain fan-out instead, so they bind whatever answered. `Aggregator.SearchAuthorsWithOutcome` (`internal/metadata/aggregator.go:74`) only returns an error when EVERY provider fails (`:90`), so with OpenLibrary timing out and DNB answering, the search reports success and hands back a DNB record, which the importer writes as the author's permanent identity. That is #2271's failure mode verbatim, on a path #2271 did not touch.

**2. The provider label is hardcoded.** Whatever record the search returned, these three lines stamp OpenLibrary on it:

- `internal/migrate/csv.go:175` — `full.MetadataProvider = "openlibrary"`
- `internal/migrate/readarr.go:133` — same
- `internal/migrate/goodreads.go:349` — same

So a row can be stored with a `dnb:` foreign ID while claiming `metadata_provider = openlibrary`. The two fields disagree, and `models.AuthorProviderFromForeignID` routes catalogue fetches off the prefix, so the label is not merely cosmetic, it is a lie the rest of the code steps around.

## Consequence

The duplicate authors in #2117. Once OpenLibrary is answering again, a lookup by the real OpenLibrary key misses the row bound to the DNB key and mints a second author for the same person. The user is left with two rows, neither obviously wrong, and no indication that a provider timeout during a one-off import is what split them.

## Repro

1. Make OpenLibrary author search time out (block egress or point it at a black hole) while leaving DNB reachable.
2. Import a CSV containing an author DNB also knows.
[...]

Issue #2271, opened 2026-08-27
A throttled Hardcover call silently and permanently downgrades an author to OpenLibrary


jmerhar:
### What happens

With `metadata.primary_provider = hardcover` and a valid token, a Hardcover call that returns HTTP 429 falls back to OpenLibrary for that lookup. The author is then created linked to an OpenLibrary id — and per the v1.33.0 release notes, *"the catalogue provider is chosen per author from the ID they are already linked to, so authors added under OpenLibrary keep syncing from OpenLibrary."*

So a **transient** rate limit produces a **permanent** provider choice, and nothing reports it.

### Why this is worse than a slow import

Re-running the import does not repair it — the author is already linked, so Hardcover is never consulted again for that author. The only remedy is a manual relink per author, or wiping the library, re-importing, and hoping to win the rate-limit race.

On a free-tier token, 429 is the *expected* steady state for a library of even a few dozen items, so "primary provider = Hardcover" silently becomes "mostly OpenLibrary" on first import, permanently.

### Evidence

Bindery `sha-ff9b62d`. One Audiobookshelf import, 72 items, produced 86 log lines of `HTTP 429: API rate limit exceeded for tier 'Free'. Try again in 1 seconds.` from Hardcover, across `hardcover search authors`, `hardcover search series`, `hardcover get series catalog` and `hardcover get book by isbn`.

Resulting library:

```
books   by provider:  hardcover 19 · openlibrary 35 · audiobookshelf 12
authors by id type:   OpenLibrary 38 · abs 3 · hardcover 1
```

35 of 66 books and 38 of 42 authors landed on OpenLibrary despite Hardcover being the configured primary and the token never being rejected. Startup logs `metadata primary provider provider=hardcover`, and there is no fallback or token-rejected warning at any point.

### Why this is not #2237

#2237 (fixed in v1.33.1) covers the case where *"Hardcover's author search does not return"* the author, and added a warning plus a `providerMismatch` object on the add response. That path is about a **miss**.

A **429** is not a miss — the record may well exist — and it produces **no** `providerMismatch` and no warning of any kind in my logs.

### Relationship to #2075

#2075 asks for `Retry-After`/backoff and a provider-level rate limit, which would reduce how often this is triggered. This issue is about the *durability* of the result when it is triggered: even with perfect backoff, any 429 that does slip through still permanently changes which provider an author syncs from, silently. The two are worth fixing independently.

### Suggested behaviour

[...]

Issue #2237, opened 2026-08-26
Adding an author silently falls back to OpenLibrary when Hardcover's author search misses, defeating `metadata.primary_provider = hardcover`


ianepreston:
### Summary

With `metadata.primary_provider` set to `hardcover`, adding an author whose canonical
Hardcover record isn't returned by `search(query_type: "Author")` silently creates an
**OpenLibrary**-linked author. Because `providerForForeignID` derives the catalogue
provider from the foreign-ID prefix, that author then syncs from OpenLibrary forever —
the exact outcome selecting Hardcover as primary is meant to avoid (#2040). Nothing in
the UI or the API response indicates this happened.

The record itself is reachable: `GetAuthor("hc:<slug>")` resolves it fine. Only the
search path misses it.

### Version

v1.33.0 (ghcr.io/vavallee/bindery:v1.33.0), Hardcover token configured,
`metadata.primary_provider = hardcover`, `hardcover.enhanced_series_enabled = true`.

### Reproduction

Author: Matt Dinniman — https://hardcover.app/authors/matt-dinniman (Hardcover author 241306).

`GET /api/v1/search/author?term=Matt%20Dinniman` returns two results:

```
OL3101279A       openlibrary   Matt Dinniman
hc:supersummary  hardcover     SuperSummary
```

No query form reaches the canonical record:

| term | Hardcover results |
| --- | --- |
| `Matt Dinniman` | `hc:supersummary` |
| `matt-dinniman` | `hc:supersummary` |
| `Dinniman` | `hc:dinniman` ("Dinniman"), `hc:supersummary` |
| `Dungeon Crawler Carl` | `hc:dungeon-crawler-carl_-a-litrpg_gamelit-adv`, `hc:dinniman`, `hc:matt-dinnamen`, … |

`hc:dinniman` and `hc:matt-dinnamen` are duplicate contributor-credit rows with no books,
no ratings and no bio. The real record never appears.

This is not the `dedupeAuthorsByName` merge collapsing it: `GET /api/v1/author/{id}/relink-upstream/candidates`,
which is explicitly broader and does **not** merge records, misses it identically. The
canonical record is absent from the Hardcover `search` response itself.

Adding from that list produces an OpenLibrary author, and `providerForForeignID`
(`internal/metadata/aggregator_providers.go:37`) routes every later catalogue fetch to
OpenLibrary because the id has no `hc:` prefix.

### Workaround

[...]

PR #2654, opened 2026-09-17
fix(abs): refuse a fallback ISBN relink while the primary is down (#2642)

### LINKED ISSUE #2642: Audiobookshelf import relinks books to a fallback provider's ISBN record while the primary is down
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate (searched `lookupUpstreamBook`, `GetBookByISBN`; #2271, #2332 and #2612 cover the same guard on other paths)
- [x] Found in source while fixing #2612 (PR #2641)

### Current behaviour

`internal/abs/import_upserts.go` `lookupUpstreamBook` (around :567) matches an Audiobookshelf item to an upstream book by ISBN through `metadata.Aggregator.GetBookByISBN`. That method does not report whether the primary provider answered, so when the primary times out and a fallback (DNB, Google Books, Hardcover) returns the ISBN, `enrichBook` hands the fallback's record to `mergeUpstreamBook`, which sets `book.ForeignID` and `book.MetadataProvider` to the fallback's (around :520) and counts it as relinked.

That is the binding #2610 refuses for authors in the CSV, Readarr and Goodreads importers, and #2612 (PR #2641) refuses for Add Book: once the primary is back, a later lookup by the primary's key misses the relinked row, which is how duplicates start (#2117, #2271, #2332). `GetBookByISBN` also caches the fallback's answer, so for up to the cache TTL the next import repeats it even with the primary answering.

### Expected behaviour

While the primary did not answer, an ISBN match that came from a fallback should not relink the book: skip the relink for that item with a message like the existing "book relink skipped" ones, and leave the row on its current identity so the next import retries. A primary that answered with no match should still allow the fallback (the #2237 rule).

### Steps to reproduce

Source review, no live trace:

1. OpenLibrary primary, DNB or Hardcover enabled as a fallback.
2. Make OpenLibrary time out (black hole it).
3. Run an Audiobookshelf import over items with ISBNs that already exist in Bindery.
4. The books are relinked to the fallback's ids and provider.

### Trace logs

Source review, no live trace.

### Deployment method

Any

### Bindery version

main after v1.36.1

tunglambk:
## Summary

`lookupUpstreamBook` matched an ABS item's ISBN with `GetBookByISBN`, which steps past a primary that timed out and returns the next provider's record without saying so. `enrichBook` handed that record to `mergeUpstreamBook`, which wrote the fallback's `foreign_id` and `metadata_provider` onto the row and counted it as relinked.

The ISBN branch now uses the `GetBookByISBNWithOutcome` #2612 added for the Add Book dialog, and refuses an outcome `SafeToBind` rejects with the same `PrimaryProviderUnavailableError` `lookupUpstreamAuthor` already returns. `enrichBook` catches it and reports `book relink skipped: …` naming the provider that did not answer. The row keeps the identity it came in with, so the next import retries. A primary that answers and simply has no record for the ISBN still lets the fallback's record through (#2237).

Changelog fragment and `docs/abs_import.md` / `docs/ABS-Import-Wiki.md` updated.

Closes #2642.

## Why

A primary that times out is a fact about a few seconds, not about the ISBN. Writing the fallback's id onto the row makes that provider permanent, and the primary's own key then misses the relinked row, which is how one book becomes two (#2117, #2271, #2332). #2610 fixed the CSV, Readarr and Goodreads importers, #2612 fixed Add Book, and #2641 left the Audiobookshelf matcher for a separate issue. This is that one.

## Test plan

- [x] `TestABSImportRefusesBookRelinkOnPrimaryFailure` fails before the fix (`book.ForeignID = "dnb:1305873874"`, `MetadataProvider = "dnb"`, `Relinked = 1`, `book relinked by isbn metadata match`) and passes after (`abs:book:…` kept, `Relinked = 0`, `book relink skipped: … did not answer …`)
- [x] `TestABSImportStillRelinksBookWhenPrimaryMerelyMisses` passes before and after — the #2237 counterweight
- [x] `go build ./...`, `go vet ./internal/abs/... ./internal/metadata/...`, `gofmt -l` — clean
- [x] `go test -count=1 ./internal/importer/... ./internal/abs/...` — `internal/abs` ok. `internal/importer` fails 9 hardlink/rename tests, all of which fail the same way on unmodified `main` in this sandbox (`/tmp` is a different device); verified against a worktree at `af0ecd3e`
- [ ] `golangci-lint run` — the binary available here is v2.12.2, not CI's pinned v2.11.4, and it panics on Go 1.27, so I couldn't run the lint gate locally


vavallee:
Merged, thanks @tunglambk. You beat me to it by about fifteen minutes: I had a near identical fix up as #2655 and closed it in favour of yours. This will go out in the next patch release.

Issue #2117, opened 2026-08-20
cross-provider author-identity/duplicate behaviour


TheOpsMgr:
### Pre-flight

- [x] I searched existing issues and this isn't a duplicate
- [x] I'm running the latest version (check Settings → About)

### Current behaviour

Duplicate authors are created when OpenLibrary author looked fails and author fetching fails over to DNB.

### Expected behaviour

Authors should be automatically de-duped when OpenLibrary queries come back online.

### Steps to reproduce

1. spoof OpenLibrary being offline via hosts end or firewall block
2. import author.csv file that contains some existing authors to trigger author fetch process
3. see that duplicate authors are created

<img width="1883" height="1661" alt="Image" src="https://github.com/user-attachments/assets/34cd7acc-f20c-445f-aefb-fb9bb1adb8ed" />

### Trace logs

```text
08/19, 10:11:38 | WARN |   | metadata search: provider failed | error="search authors: openlibrary request: Get \"https://openlibrary.org/search/authors.json?q=Adam+Roberts&limit=20\": context deadline exceeded" provider=openlibrary -- | -- | -- | -- | -- 
08/19, 10:11:35 | ERROR |   | failed to fetch books | author="Adrian Tchaikovsky" error="dnb get author works dnb:1394981503: Get \"https://services.dnb.de/sru/dnb?maximumRecords=50&operation=searchRetrieve&query=per%3D1394981503&recordSchema=MARC21-xml&version=1.1\": context canceled" 
08/19, 10:11:35 | WARN |   | author profile refresh: metadata lookup failed | author="Adrian Tchaikovsky" error="dnb does not support author lookup by ID" foreignId=dnb:1394981503 
08/19, 10:11:35 | INFO |   | fetching books for author | author="Adrian Tchaikovsky" foreignId=dnb:1394981503 
08/19, 10:11:35 | WARN |   | metadata search: provider failed | error="search authors: openlibrary request: Get \"https://openlibrary.org/search/authors.json?q=Adrian+Tchaikovsky&limit=20\": context deadline exceeded" provider=openlibrary
```

### Deployment method

Docker (ghcr.io)

### Bindery version

1.32.0

TheOpsMgr:
[Related Issue and analysis](https://github.com/vavallee/bindery/issues/2075#issuecomment-5358396137)

```
Your duplicate author theory is right, and I found the mechanism. You worked it out from log symptoms, which was a good call.

[...]

PR #2655, opened 2026-09-17
fix(abs): do not relink a book to a fallback ISBN record while the primary is down (#2642)

### LINKED ISSUE #2642: Audiobookshelf import relinks books to a fallback provider's ISBN record while the primary is down

vavallee:
## Summary

The Audiobookshelf import's ISBN match now checks whether the primary metadata provider answered. When it timed out and a fallback supplied the record, the book keeps its current identity and the item result says `book relink skipped: metadata provider openlibrary did not answer: ..., so it was not bound to a fallback provider`. The ABS field update and the rest of the item import are unchanged.

Closes #2642.

## Why

`lookupUpstreamBook` used `GetBookByISBN`, which drops the lookup outcome, so a fallback's record reached `mergeUpstreamBook` and overwrote `ForeignID` and `MetadataProvider` for good. Once the primary is back, a lookup by its key misses the row, which is how duplicates start (#2117, #2271, #2332). This follows #2610 (CSV, Readarr, Goodreads) and #2641 (Add Book): `GetBookByISBNWithOutcome` plus `SafeToBind`, returning the existing `PrimaryProviderUnavailableError` the author path already uses. A primary that answers with no match still lets the fallback through (#2237).

ASIN and title paths checked, no fix needed: `GetCanonicalBookByASIN` only returns records from the primary's own search (a failed search gives no match), and the title path reads works and the book from the provider that owns the author's existing ForeignID.

## Test plan

- [x] New `TestABSImportDoesNotRelinkBookToFallbackISBNRecordWhilePrimaryDown` failed before the fix (book moved to `dnb:1234567890`, metadataRelinked 1, no skip message) and passes after
- [x] New `TestABSImportRelinksBookToFallbackISBNRecordWhenPrimaryMisses` (#2237 rule) passes before and after
- [x] `gofmt -l internal/abs` clean
- [x] `go vet ./internal/abs/...`
- [x] `go test -count=1 ./internal/abs/... ./internal/metadata/...`
- [x] `golangci-lint run --allow-parallel-runners ./internal/abs/...` 0 issues

https://claude.ai/code/session_016fJcCVbNnKsmj2MAAMwWj9


vavallee:
Closing in favour of #2654 from @tunglambk, which landed the same fix first, with the same guard, message and tests, and also updated docs/abs_import.md.

PR #2673, opened 2026-09-17
feat(author): drop thin edition clusters from catalogue sync (#2235)

### LINKED ISSUE #2235: Additive scored signals instead of first-drop-wins boolean filtering

gchahcg:
## Summary

Metadata profiles gain an opt-in **minimum edition count** filter for author catalogue syncs: works whose title cluster reports fewer known editions than the profile's floor are skipped before they become book rows. Closes #2235.

This is the smaller, safer shape of the request from the issue thread — a per-profile number, not a global catalogue-wide policy — and it leaves the closed PR #2628's cluster-signalling approach behind.

## Implementation notes

| Aspect | Behaviour |
|---|---|
| Cluster key | Normalised (lowercased, trimmed) work title |
| Cluster count | Max of the known `EditionCount` across the works in the cluster — one well-editioned work saves the title's thinner siblings |
| Unknown count | Passes the filter ("unknown is not zero", same rule as MinPages). Critical for Hardcover-primary authors, whose records never carry an edition count; a sum-based rule would have emptied their catalogues |
| Already-tracked books | Maintained in place, never dropped, never counted as skipped |
| Single-work adds | Exempt, matching the #1612 rule that explicit user picks are not vetoed by catalogue heuristics |
| Default | `0` = filter off; the cluster map is only built when the profile sets a floor, so the default is zero-cost |

- Migration `086_metadata_profile_min_edition_count.sql` adds `min_edition_count INTEGER NOT NULL DEFAULT 0` to `metadata_profiles`.
- The skip is reported in the sync summary (`SkippedThinCluster` + sample, capped at 5) and surfaced as its own line in the author page's sync notice.

## Where it's mounted

- `fetchAuthorBooks` (`internal/api/authors.go`): cluster map built in the candidate loop, filter applied in the create loop after existing-book resolution.
- Profile CRUD: `internal/db/metadata_profiles.go`, `internal/models/metadata_profile.go`.
- Web: profile editor form + profile-list chip (`MetadataTab.tsx`), sync notice breakdown (`AuthorSyncNotice.tsx`), en i18n keys (other locales fall back to en).

- The issue's larger ask — a global, catalogue-wide thin-title policy and the two-stage cluster signal from #2628 — can land separately once this shape is proven in practice.

## Checklist

- [x] Tests added or updated — 4 new `fetchAuthorBooks` tests (drop, opt-in default, owned-book exemption, single-work exemption) plus the twin-title fixture fix in `EditAuthorModal.test.tsx`
- [x] Doc-update gate cleared — `docs/User-Guide-Wiki.md` profile-filter list

## Test plan

- [x] `make test` (full Go suite, CI gate)
- [x] `gofmt -l .`, `go vet ./...`, `golangci-lint v2.11.4`, `govulncheck`
- [x] `make smoke` (real binary, exercises the new migration)
- [x] `cd web && npm run typecheck && npm run build && npm test` (1002 tests) and `npm run lint` (0 errors; remaining warnings pre-existing)


vavallee:

This PR adds a migration numbered **086**, and so does #2673 or #2633, whichever this is not. Both are open and both claim it. `assertUniqueMigrationVersions` (`internal/db/db.go:685`) refuses to boot when two files share a numeric prefix, so whichever merges second breaks startup for anyone with both.

086 is a genuine gap on main, which goes 085 then 087, and it has never been filled by a merged commit. Three separate branches have each created their own 086 at some point, which is how the gap became attractive to all of them at once.

* 086 → #2673, #2633
* 090 → #2761, #2750, #2713
* 091 → #2714
* 092 → #2737

Suggest renaming to 093 or above rather than reusing the gap, and picking the number by checking open PRs rather than main. The reason this keeps happening is that each branch computes "next free" against a main that excludes everyone else's in flight work.

PR #2713, opened 2026-09-19
feat(downloader): remove the torrent after a successful import (#2046)

### LINKED ISSUE #2046: feat: remove torrent from download client after successful import
When Bindery successfully imports a book from a torrent client, the completed torrent is left in the download client indefinitely. There's no built-in way to clean it up after import.

This is already implemented for Usenet — SABnzbd calls `DeleteHistory` and NZBGet calls `RemoveHistory` via a cleanup callback passed to `tryImportInternal`. Torrent clients pass `nil` in that slot, and `RemoveTorrent` exists in the Transmission client but isn't wired up post-import.

**Proposed change**

Add a per-client `remove_on_import` toggle (default `false`). When enabled, pass a cleanup callback to `tryImportInternal` after successful import calling `RemoveTorrent` with `deleteFiles: false` — keeping Bindery's importer in control of the source files.

**Scope**
- DB migration: `remove_on_import bool` on `download_clients`
- `tryImportTransmission`, `tryImportQbittorrent`: pass callback when flag is set
- API: expose field on GET/PUT download client endpoints
- UI: toggle in download client settings

Happy to open a PR if this direction looks good.

ccarpinteri:
Closes #2046.

## What it does

Adds a per client "remove on import" toggle, off by default. When it is on, Bindery removes the torrent from the client once it has imported the download. `deleteFiles` is false, so the files stay on disk and a hardlinked import keeps working.

Off by default for the reason you raised on the issue: private tracker users need it off, because removing a torrent stops it seeding and costs them ratio. The help text under the toggle says so.

## Wiring it into every success route

You said the import path has more than one success route and the callback needs to be at each of them, so I went through them all.

`tryImportInternal` already ran `cleanupFunc` on three routes. The fourth, the "no book files at the path but the book is already in the library" return, did not, so it does now.

The two "book already in library" shortcuts in `checkQbittorrentDownloads` were the real gap. Both close a download out and mark it imported without ever calling `tryImportInternal`, so no cleanup could have run there at all. They now build the same callback through a shared helper, `qbittorrentRemoveOnImportCleanup`.

Transmission's poller has no equivalent shortcut. Everything goes through `tryImportInternal`, so it was already covered.

The manual import API path builds a synthetic Download with no `TorrentID` and no client, so there is nothing to remove and it is left alone.

## Usenet

Untouched. `tryImportNZBGet` already calls `RemoveHistory` on import unconditionally, so there is nothing for the toggle to control. The settings form hides it for usenet clients, and the edit form sends it as false if a client is switched from a torrent type to a usenet one, so a stale flag cannot be left on the row.

## UI

Checkbox in Settings, Download Clients, under the path remap field, for torrent clients only. Screenshot of it on my own install with Transmission selected: the toggle, the label "Remove torrent after import", and the help text explaining the ratio tradeoff.

## Tests

- `ClientsTab.test.tsx`: hidden for usenet, reflects the stored value, defaults off and sends the change on save, never sent enabled for a usenet client
- existing `SettingsPage` payload assertions updated, since torrent clients now carry the extra key
- Go side: `importer` and `db` suites pass, including the existing cleanup callback tests

Full web suite: 92 files, 1058 tests, passing. It needs Node 21 or newer to run, since jsdom's undici calls `webidl.util.markAsUncloneable`; on Node 20 the suite cannot start. I ran it on the same Node the Dockerfile uses.

## Note on ordering

Stacks on #2711 and #2712. #2711 matters for this one: it makes a Transmission download's stored identifier stable. Without it, a stored id can come back pointing at a different torrent after a daemon restart, and this feature would then remove a torrent the user never grabbed. I would rather this did not land first.

Migration is numbered 090 against current main.


Ararana:
Thanks for this, it covers exactly the gap I hit. One request: could the toggle offer a "remove torrent and data" option alongside "remove torrent only"?

vavallee:

This PR adds a migration numbered **090**, and so do two others that are open right now: #2761 and #2713 and #2750 all claim 090. `assertUniqueMigrationVersions` (`internal/db/db.go:685`) refuses to boot when two files share a numeric prefix, so whichever of the three merges second breaks startup for anyone who has both.

* 086 → #2673, #2633 (also a collision, and 086 is a gap on main that three different branches have each filled with something different)
* 090 → #2761, #2750, #2713
* 091 → #2714
* 092 → #2737

The usual gotcha here is that each of us computes "next free" against a main that does not yet contain anyone else's unpushed work, so this keeps happening. Worth renaming to something above 093 and picking by checking open PRs rather than main. No rush on my side, just flagging it before two of these land in the same window.

